PR #20, opened 2020-08-06
Add a template to the workshop custom post type.

StevenDufresne:
### Description
This PR adds a template to the `workshop` post type with the necessary components.

If we are good with this approach, I'll add in some more components and style the default components within the theme.

@coreymckrill If you want, you can merge this into your PR as well.


wp-content/plugins/wporg-learn/inc/post-type.php
```diff
@@ -58,8 +58,37 @@ function register_workshop() {
 		'exclude_from_search'   => false,
 		'publicly_queryable'    => true,
 		'capability_type'       => 'page',
-		'show_in_rest'          => true
-
+		'show_in_rest'          => true,
+		'template' => array(
+			//array( 'wporg-learn/workshop-video' ), <- Doesn't exist yet
```
StevenDufresne: @coreymckrill What do you think about this approach?

StevenDufresne: I haven't yet figured out how to set the column with programmatically yet though.

coreymckrill: I think a template is a good approach, but I was thinking it would only include the stuff that's within the content column in the mockups. So, video embed, description, learning outcomes, and comprehension questions. I think, at least for now, it would be simpler to build the details column and the speaker bio into the theme template itself, rather than have those as custom blocks.

StevenDufresne: Yeah, that's a good call. I'll move forward with that.

PR #24, opened 2020-08-08
Try consolidating the plugins

coreymckrill:
This moves the details list block functionality over into the wporg-learn plugin, since that block can only be used in the context of the wporg_workshop post type. It also replaces the link-button block with a custom block style applied to the core/button block.

This also introduces a build config/step to the wporg-learn plugin. Since the wporg-learn theme also has a build config, it might make sense to set up some yarn workspaces in the root of the repo so we could build everything in one step.

@StevenDufresne let me know what you think about this approach.


coreymckrill:
Merging this now so I can get it committed to svn, but if there are concerns, we can still discuss.

Issue #22, opened 2020-08-07
Workshop page Url should be `workshop/{id}`.


StevenDufresne:
It currently is `/wporg_workshop/{id}`.

PR #28, opened 2020-08-10
Update the url to be workshop/id.

StevenDufresne:
### Description
This PR adds a filter to plugin to change the urls from `wporg_workshop/${id}` to `workshop/${id}`.

Addresses: #22


dd32:
Can't this just be added directly to https://github.com/WordPress/learn/blob/trunk/wp-content/plugins/wporg-learn/inc/post-type.php#L99-L142 ? Or is there some re-use that i'm missing?

tellyworth:
Seems fine to me with very light testing, but I have the same question as Dion.

StevenDufresne:
Moved!

PR #30, opened 2020-08-10
Use information from the local user for the workshop.

StevenDufresne:
### Description

This PR uses local user information to populate the presenter info. I recognize this may fall a bit short of what we want long term but we don't currently have an easy way to populate that information.

This makes use of a WordPress users:
- Display Name
- Nickname (ie: @username)
- Url (.org profile url)
- Gravatar url.
- Description

Addresses: #42


wp-content/themes/pub/wporg-learn-2020/template-parts/component-author.php
```diff
@@ -1,9 +1,27 @@
...
+$nickname = get_the_author_meta( 'nickname' );
+$user_url = get_the_author_meta( 'url' );
+$url = get_avatar_url( get_the_author_meta( 'ID' ) );
+
+?>
+
 <div class="workshop-author">
 	<div>
-		<img class="workshop-author_profile" src="https://secure.gravatar.com/avatar/1064bd40c9ade0b3c9cf0e27d9e3432a?s=200&d=mm&r=g" />
+		<img class="workshop-author_profile" src="<?php echo $url; ?>" />
```
coreymckrill: I would recommend using `get_avatar` here, which will give you the whole `img` tag.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-author.php
```diff
@@ -1,9 +1,27 @@
...
+/**
+ * Template part for display author information
+ *
+ * @link https://codex.wordpress.org/Template_Hierarchy
+ *
+ * @package WPBBP
+ */
+
+$display_name = get_the_author_meta( 'display_name' );
+$nickname = get_the_author_meta( 'nickname' );
```
coreymckrill: I think you want `user_nicename` here instead of `nickname`. The nickname isn't necessarily the same as the "user name" that you see on their wporg profile.

coreymckrill (commented):
The main thing here, I think, is that this is going to output the info for the author of the workshop _post_, which most likely won't be the same as the person/people who is presenting in the workshop.

We've got a post meta value set up for storing the wporg user name for one or more presenters, so what I would recommend is getting that first,
`get_post_meta( $post_id, 'facilitator_wporg_username' )`
which will give you an array of usernames, and then iterate through those and output this markup for each one. Instead of `get_the_author_meta`, you'd use `get_user_by( 'login', $username )` which will give you a `WP_User` object, and then you can retrieve user meta values for that by doing, for example, `$user_object->display_name`.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-author.php
StevenDufresne: Here I'm using the nickname as a hack to handle the @{username} to match the designs. I'm not sure the display name here will match the .org profile name?

coreymckrill: Yeah, `display_name` and `nickname` are both fields where the user can enter something arbitrary. To match the `@{username}` in the design, you'll want to use `user_nicename` or `user_login`.

StevenDufresne:
@coreymckrill Can you take another look when you get the chance? Feel free to make any updates you see fit.

PR #34, opened 2020-08-10
Freeze jetpack at 8.7 for now.

StevenDufresne:
### Description
This sets jetpack to 8.7. It appears there is a problem with 8.8. 

We'll need to dig in later.


coreymckrill:
The issue was fixed in #67, so I think we can close this.

PR #32, opened 2020-08-10
Update the first category name and set the lesson plans to it.

StevenDufresne:
### Description
This PR changes the name & slug of the `uncategorized` category and sets all the imported lesson plans to that category.

This makes sure the archive page displays something.

It's a potential solution for #31.

Next iteration, we could ask the training team to update the manifest to have more categories.

Issue #31, opened 2020-08-10
Lesson Plans need categories


StevenDufresne:
The lesson plan archive is incomplete. The lesson plan archive will pull in lesson plan categories and make them available in the "navigation" bar. Lesson plans are then displayed if they match categories. Since we don't automatically create categories, nor have lesson plan assigned, no lesson plans are visible.

Maybe the quickest route is to add an "All" category and after import assign them all to that category?

PR #47, opened 2020-08-10
Lesson Plans: Overhaul the lesson plan archive

coreymckrill:
This simplifies the lesson plans archive view and related category
filters, removing custom rewrite rules and trimming template logic.

* Register a new taxonomy for Lesson Categories, instead of using the
  core category taxonomy. This way lesson plans won't be mixing with
  blog posts.
* Use the same rewrite slug for the lesson categories as for the lesson
  plan post type. This way we get the same desired URL structure of
  `lesson-plans/{category name}/page/2` but with way less complexity
  in the code.
* Set up archive template files specifically for lesson plans and the
  new taxonomy. Blog posts and other post types can now fall back on
  the default archive template without having the filtering UI at the
  top.
* This eliminates the need for all lesson plans to be assigned some kind
  of "All" category upon creation. The default view in the archive is
  now "All", which just means no particular category term has been
  specified.

Refs #31, #32


StevenDufresne:
I created a category and expected to be able to add it to a lesson plan but it didn't show up in the category list.

coreymckrill:
🤔 @StevenDufresne which screen were you on when you created a category? It's working for me. Maybe you created a category for posts instead of a category for lesson plans?

StevenDufresne (approved):
Looks good.

Issue #40, opened 2020-08-10
Hide "Captions" on frontend if none exist


hlashbrooke:
The workshop details on the frontend show an item labelled "Captions" - this should be hidden if no captions are added to the workshop post.

PR #55, opened 2020-08-11
Hide all empty captions.

StevenDufresne:
### Description
This PR updates the details blocks to not output row for empty fields. The initial issue reported in #40 suggested to not to it for captions, but I couldn't find a reason to not do it for all empty fields.


coreymckrill (approved):
👍 Looks good

Issue #57, opened 2020-08-11
Home Page buckets are not linked


StevenDufresne:
The buckets of content & links are not linked to anywhere. When addressing this, we should make sure the copy matches the links seeing that the copy reflects a post MVP state when there are advanced filters and numerous workshops. We only plan on launching with a couple and without advanced filtering.

### Screenshots
![](https://d.pr/i/v3Ek49.png)

hlashbrooke:
Right now, none of that content is relevant or useful - I'd be more in favour of removing it entirely for now to be honest. I'd rather have that space looking empty in terms of content rather than having unnecessary content and links in there.

StevenDufresne:
@hlashbrooke Sounds good. I'll pull it out.

PR #58, opened 2020-08-11
Link all the buckets to the lesson plan & workshop lander.

StevenDufresne:
This PR adds links to the homepage buckets. This PR will link any lesson plan content bucket to `lesson-plans` and any workshop bucket to `workshops`.

I tried a few different layouts, adding/removing copy, and couldn't find one that worked since we don't have categorization on the lesson-plan or workshop archives yet. I tried the homepage without the content but it makes it pretty lean and since the buckets still provide important information about what lesson plans/workshops are, and their 'kinds', I thought that even if they all link the same place, the content stays,...... it may be the best approach. 

Any other thoughts on what can be done here?
@hlashbrooke @coreymckrill

Addresses: #57


StevenDufresne:
Closing in favour of #70

Issue #17, opened 2020-08-06
Video Grid Item: Support clicking image for navigation.


StevenDufresne:
Right now the link under the image will lead you to the workshop. You should be able to click the image as well.  I didn't wrap the image within the <a> because I don't think that's great for accessibility. Will revisit.

PR #59, opened 2020-08-11
Add a javascript file to make the workshop images clickable.

StevenDufresne:
### Description
This PR will allow users the ability to click on the workshop thumbnails in the small list and the feature image and navigate to the appropriate workshop.

Addresses: #17


coreymckrill:
@StevenDufresne Is there a way we could do this without introducing another JS file / http request? Could the image be wrapped in a separate anchor tag instead?

StevenDufresne:
It's historically what I would have done but from working through the meta accessibility tickets, I have convinced myself that the most logical html is the best (for seo & accessibility). This may be a case where that logic is overstepping though and there aren't any negative accessibility concerns. 

@ryelle What are your thoughts on this?

ryelle:
It looks like the alt attributes for the images are empty. If that's intentional (these are "decorative" images), then I think it's fine to wrap the image in the link too, ex:

```html
<li class="col-4 video-grid_item">
	<a href="https://learn.wordpress.org/workshop/intro-to-gutenberg-block-development/">
		<img alt="" src… />
		<span class="video-grid_item_link">Intro to Gutenberg Block Development</span>
	</a>
</li>
```

This way a screen reader pulling out the link content would still only get "Intro to Gutenberg Block Development". What you want to avoid is duplicate info, so don't wrap it in a separate link, and don't add an `alt` that would be repetitive (bad ex: "thumbnail for intro to…").

StevenDufresne:
Closing in favour of wrapping the image in #76

PR #67, opened 2020-08-11
Build: Use a custom webpack config to provide a jsonpFunction

### LINKED ISSUE #27: Workshop Details block: editor script causes whitescreen when minified
Then the script for the `workshop-details` block is built with `wp-scripts build`, the code is minified, and when it loads in the editor, you get a white screen of death. Curiously, the errors in the console relate to Jetpack, rather than our plugin. Nevertheless, if you disable our block, everything works again.

If the block script is instead built with `wp-scripts start`, the code is not minified, and no white screen of death occurs. 🤷‍♂️

ryelle:
The jsonpFunction is a global function used to load application chunks. If multiple webpack-bundled scripts are on the same page, these functions will conflict. To avoid these conflicts, a custom jsonpFunction is provided. Since this must be done in a webpack.config.js, the custom entrypoints have also been moved off the command line & into the config.

Fixes #27

_Ironically, [I've run into this before.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/520)_

> As noted in [the webpack docs:](https://webpack.js.org/configuration/output/#outputjsonpfunction)
>
> > A JSONP function name used to asynchronously load chunks or join multiple initial chunks (SplitChunksPlugin, AggressiveSplittingPlugin).  **This needs to be changed if multiple webpack runtimes (from different compilation) are used on the same webpage.**

Ideally this would be fixed in wordpress/scripts, but until then we need to have our own webpack.config.js, extending from the defaults.


coreymckrill (approved):
Yay!

PR #69, opened 2020-08-11
Workshop Details: Make the human readable duration more succinct

### LINKED ISSUE #60: Duration is too verbose.
I don't think we need to show the duration with complete accuracy. Users can get that from the video. I see it more as a "2 min read" experience where you are just glancing at it to have a rough sense of how long it would take. 

### Solution
I think we can say:

1. Under 1 hour?
 -> "43 minutes"

2. Over 1 hour?
-> "1 Hour, 43 minutes."

3. Between 24 & 48 hours?
-> "1 day" (round up?)

_The last one may not apply._

I think we don't ever show seconds.

coreymckrill:
```
    Under 1 hour?
    -> "43 minutes"

    Over 1 hour?
    -> "1 Hour, 43 minutes."

    Between 24 & 48 hours?
    -> "1 day" (round up?)

I think we don't ever show seconds.
```

There might be a time when the duration is only in seconds, though, so in that case we'll show seconds.

Fixes #60

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -103,7 +103,7 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
 					);
 				}
 			} elseif ( $interval->i > 0 ) {
-				$return = $minutes = human_time_diff( 0, $interval->i * MINUTE_IN_SECONDS );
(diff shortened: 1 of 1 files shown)


StevenDufresne:
This is working as expected. I think we could tighten up the meta box labels to simplify the sidebar but it's not a blocker for this pr.

StevenDufresne (approved):

PR #73, opened 2020-08-12
Theme: Show video for featured workshop on lander instead of thumbnail

coreymckrill:
The mockup appears to call for the video of the featured workshop to be embeded at the top of the lander page, rather than a featured image.

https://www.figma.com/file/UtmG8KQv3Csy8hNtuAWkwZ/Learn-WordPress?node-id=106%3A0

Since the workshop post type doesn't have a meta field for the video URL, but instead just has a required video block, we can search for that block and render it individually on the lander page.

This would eliminate the need for the video URL meta field, as called for in #3


StevenDufresne:
Right, I wanted to cycle back here. Although the designs show a video, I find it weird because it appears to me like it needs to be watched within the context of the workshop to see the additional information (Outcomes & Comprehension). 

If you start watching here and then click over, progress will be lost. With that in mind, I didn't see a use case where watching the video here made sense so used the feature image.

coreymckrill:
Good points. I do think having a giant featured image there is also a bit awkward. Thoughts @hlashbrooke @melchoyce?

dd32:
Seems like as with #76 maybe the image should be a link, and/or have a fake play button that loads the workshop page?

dd32:
> Seems like as with #76 maybe the image should be a link, and/or have a fake play button that loads the workshop page?

I ended up putting together an experiment of that, #89

hlashbrooke:
Playing the video on the archive page isn't ideal - the solution of a play button overlay is good and I'd rather to go ahead with that than playing a video directly on the archive page.

PR #75, opened 2020-08-12
Markdown import: Use `$supported_post_types` as a string

### LINKED ISSUE #72: PHP Warnings due to Markdown importer
`E_WARNING: wp-content/plugins/wporg-learn/inc/class-markdown-import.php:168 - in_array() expects parameter 2 to be array, string given`

https://github.com/WordPress/learn/blob/trunk/wp-content/plugins/wporg-learn/inc/class-markdown-import.php#L15 is supposed to be an array (based on it's name, and how it's used), but is a singular string.
Either that should be set to an array, or all the `in_array()`'s should be swapped to direct matches.

However, it's also later also used as a string: https://github.com/WordPress/learn/blob/trunk/wp-content/plugins/wporg-learn/inc/class-markdown-import.php#L96

coreymckrill:
It appears that the class was originally written to allow for multiple
post types, and so some places in the class assume the property is an
array. However, other places in the class assume it's a string, and
it's actually defined as a string.

Since we have no plans to use this for other post types, this changes
the property name to `$supported_post_type` (singular) to be more
semantically correct, and ensures that everywhere in the class treats
it as a string instead of as an array.

Fixes #72

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/class-markdown-import.php
+++ b/wp-content/plugins/wporg-learn/inc/class-markdown-import.php
@@ -12,7 +12,7 @@ class Markdown_Import {
 	private static $meta_key = 'wporg_learn_markdown_source';
 	private static $nonce_name = 'wporg-learn-markdown-source-nonce';
 	private static $submit_name = 'wporg-learn-markdown-import';
-	private static $supported_post_types = 'lesson-plan';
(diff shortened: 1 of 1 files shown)


coreymckrill:
Tested in my local environment and the imports are still working fine.

PR #79, opened 2020-08-12
Make the site better for mobile.

StevenDufresne:
### Description
This PR improves the sites layout on mobile.

See https://d.pr/v/veeRMK for what it looks like.


wp-content/themes/pub/wporg-learn-2020/template-parts/component-workshop-presenter.php
```diff
@@ -9,10 +9,11 @@
 
 $args = wp_parse_args( $args );
```
ryelle: If you give [`wp_parse_args` a default value,](https://developer.wordpress.org/reference/functions/wp_parse_args/) you can always assume that class exists on L16.

```suggestion
$args = wp_parse_args( $args, array( 'class' => '' ) );
```

ryelle (commented):
The parallelograms seem to be acting up on medium & smaller screens. If it helps, this is what we used on [five for the future](https://wordpress.org/five-for-the-future/) - only do the parallelogram styles on larger screens: https://github.com/WordPress/five-for-the-future/blob/b1cbcdb01b541924b5a046f7b7c90b732774b82c/themes/wporg-5ftf/css/utilities/_columns.scss#L1

Screen Shot 2020-08-12 at 11 00 34 AM

I also noticed that there's a point between two media queries at 768px wide, where some styles have kicked in but not everything.

(this can be merged since it's a good improvement, but the rest should be addressed)

StevenDufresne:
> The parallelograms seem to be acting up on medium & smaller screens. If it helps, this is what we used on [five for the future](https://wordpress.org/five-for-the-future/) - only do the parallelogram styles on larger screens: https://github.com/WordPress/five-for-the-future/blob/b1cbcdb01b541924b5a046f7b7c90b732774b82c/themes/wporg-5ftf/css/utilities/_columns.scss#L1
> 
> Screen Shot 2020-08-12 at 11 00 34 AM
> 
> I also noticed that there's a point between two media queries at 768px wide, where some styles have kicked in but not everything.
> 
> (this can be merged since it's a good improvement, but the rest should be addressed)

I'll make the update. Thanks for testing!

StevenDufresne:
Okay, i've updated the code. I decided just to fix the current code. I moved it into its own `.scss` file for future replacement.

Issue #39, opened 2020-08-10
Featured image for workshops


hlashbrooke:
Right now, if a featured image is not set for a workshop then nothing is displayed on the frontend. Could this be changed so that, if a featured image is not set, a default generated thumbnail is displayed? Like what you see on the homepage of WPTV: https://wordpress.tv/

tellyworth:
I added a quick implementation for this in #80.

I think the longer-term solution might be for Jetpack to fully support `wordpress.tv` embeds, since I _think_ it's supposed to generate & use automatic featured images for videos.

coreymckrill:
There is an enhancement for this in the works in #89, but the basic functionality for this is now merged, so I'm going to call this fixed.

PR #80, opened 2020-08-12
Use default video poster if no featured image was set for a workshop

tellyworth:
This is a quick stab at fixing #39.

I considered using Photon but I don't think it makes a significant difference in this case - it's using the VideoPress poster image, retrieved via `videopress_get_video_details()`, which is hosted on videos.files.wordpress.com. It might be worth adding that if we want to properly resize and/or crop, but I think the video posters will naturally be small enough already for this to work as a MVP.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -270,7 +270,7 @@ function wporg_get_post_thumbnail( $post, $size = 'post-thumbnail' ) {
         foreach ( get_post_meta( $post->ID, '', true ) as $key => $value ) {
             if ( substr( $key, 0, 8 ) === '_oembed_' && preg_match( '#https://video.wordpress.com/embed/(\w+)#', $value[0], $match ) ) {
                 $video = videopress_get_video_details( $match[1] );
-                if ( is_object( $video ) && isset( $video->poster ) ) {
+                if ( !is_wp_error( $video ) && isset( $video->poster ) ) {
                     return '<img class="attachment-' . esc_attr( $size ) . ' wp-post-image" src=' . esc_url( $video->poster ) . ' loading="lazy" />';
                 }
             }
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -254,3 +254,27 @@ function wporg_get_workshop_presenters( $workshop = null ) {
 
 	return $wp_users;
 }
+
+/**             
+ * Display a featured image, falling back to the VideoPress thumbnail if no featured image was explicitly set.
+ *          
+ * @param $post The Workshop post for which we want the thumbnail.
+ * @param $size The image size: 'medium', 'full'.
+ */     
+function wporg_get_post_thumbnail( $post, $size = 'post-thumbnail' ) {
+    $thumbnail = get_the_post_thumbnail( $post, $size );
+    if ( $thumbnail ) {
+        return $thumbnail;
+    } else {
+        $post = get_post( $post );
+        foreach ( get_post_meta( $post->ID, '', true ) as $key => $value ) {
+            if ( substr( $key, 0, 8 ) === '_oembed_' && preg_match( '#https://video.wordpress.com/embed/(\w+)#', $value[0], $match ) ) {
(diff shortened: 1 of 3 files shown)


wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -244,3 +244,27 @@ function wporg_get_workshop_presenters( $workshop = null ) {
...
+function wporg_get_post_thumbnail( $post, $size = 'post-thumbnail' ) {
+    $thumbnail = get_the_post_thumbnail( $post, $size );
+    if ( $thumbnail ) {
+        return $thumbnail;
+    } else {
+        $post = get_post( $post );
+        foreach ( get_post_meta( $post->ID, '', true ) as $key => $value ) {
+            if ( substr( $key, 0, 8 ) === '_oembed_' && preg_match( '#https://video.wordpress.com/embed/(\w+)#', $value[0], $match ) ) {
+                $video = videopress_get_video_details( $match[1] );
+                if ( is_object( $video ) && isset( $video->poster ) ) {
```
dd32: I think checking for `! is_wp_error()` is probably better, but that works too.

wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -244,3 +244,27 @@ function wporg_get_workshop_presenters( $workshop = null ) {
...
+ */     
+function wporg_get_post_thumbnail( $post, $size = 'post-thumbnail' ) {
+    $thumbnail = get_the_post_thumbnail( $post, $size );
+    if ( $thumbnail ) {
+        return $thumbnail;
+    } else {
+        $post = get_post( $post );
+        foreach ( get_post_meta( $post->ID, '', true ) as $key => $value ) {
+            if ( substr( $key, 0, 8 ) === '_oembed_' && preg_match( '#https://video.wordpress.com/embed/(\w+)#', $value[0], $match ) ) {
+                $video = videopress_get_video_details( $match[1] );
```
dd32, line 272: Data is cached for 1 hour, seems short but it'll do to start with.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-video-grid-item.php
```diff
@@ -10,10 +10,7 @@
 ?>
 
 <li class="col-4 video-grid_item">
-	<?php if( has_post_thumbnail() ) : ?>
-		<?php echo the_post_thumbnail( 'medium' ); ?>
-	<?php else : ?>
-		<div class="video-grid_item--no-image"></div>
+		<?php echo wporg_get_post_thumbnail( $post, 'medium' );; ?>
 	<? endif; ?>
```
dd32: Looks like the `endif;` should've gone here too? and while i'm here, double `;;`

wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -244,3 +244,27 @@ function wporg_get_workshop_presenters( $workshop = null ) {
...
 	return $wp_users;
 }
+
+/**             
+ * Display a featured image, falling back to the VideoPress thumbnail if no featured image was explicitly set.
+ *          
+ * @param $post The Workshop post for which we want the thumbnail.
+ * @param $size The image size: 'medium', 'full'.
+ */     
+function wporg_get_post_thumbnail( $post, $size = 'post-thumbnail' ) {
```
dd32, line 264: I'd be tempted to just rely upon the global `$post` object here, skip the arg and do `$post = get_post()` in the function instead

dd32 (changes requested):
I love how hacky this approach is, but it looks like it'll work to me (except for the syntax error).

I'll actually do some testing on it in a bit, but leaving these review comments for now.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-video-grid-item.php
tellyworth: Ugh, yeah, thanks - copy-paste errors.

dd32 (approved):

dd32:
Works for me in testing.

StevenDufresne:
Yep, this is working as expected.

StevenDufresne (approved):

PR #89, opened 2020-08-13
Theme: Add a "play" button on featured workshop image

dd32:
as per #73 and as an alternative to embedding the video directly on the page without the context of the rest of the Workshop data, I attempted to (probably badly) add an overlay for a play button on top which at least gives the user a call to action.

Creating this PR as a proposal of this - https://cloudup.com/cLKje5hHWBX
Clicking the image/"button"/etc causes it to load the Workshop, video doesn't auto-play though as it seems VideoPress's autoplay functionality doesn't work in Chrome, Firefox, or Safari.

The gradient is an attempt to match close to what the VP gradient is, but its definitely not the same.. I didn't dig far enough into it to find out what's used there.


coreymckrill:
Related: #95

coreymckrill:
It looks like the Learn team is moving towards using featured images that aren't necessarily screenshots from the video anymore, so it may not always make sense to have a play button overlay. Closing this, but feel free to reopen if you think there's still a use case here.

PR #107, opened 2020-08-27
Update archive page to have filtering

### LINKED ISSUE #5: Workshops Archive
workshop-archive

Filter UI:
workshop-archive-filter

https://www.figma.com/file/UtmG8KQv3Csy8hNtuAWkwZ/Learn-WordPress?node-id=51%3A3565

#### Milestone 1
- Workshop list

#### Milestone 2
- Search
- Filters
- Advanced Features

StevenDufresne:
### Description
This PR modifies the archive page to match the designs in #5.

#### Features
- Users can toggle filter drawer
- Users can search for workshops within the archive
- Users can apply filters to workshops

**Still need to implement:**
- Retrieved all the filters (meta) information and render them in filter template-part
- The query that searches for workshops with filters & keyword applied

### Considerations
This PR takes a progressive enhancement approach where javascript would be layered on top. This will keep the code base simple and easier for the community to manage long term. It will also handle SEO more elegantly by default.

It would be ideal if our search logic could be exposed via a rest endpoint that we could tie into javascript, later on. Something to keep in mind.


StevenDufresne:
We will want to get #110 into master and fix the conflicts that will arise in this branch. The search bar was redone in that pull request and we should align them before we merge this.

_Also noticed that the homepage search button is weird on this branch at the moment. That should be rectified once we moved the other changes in._

StevenDufresne:
@coreymckrill 

I wonder if it makes sense to include the full list of languages in the captions and language filters. As that list grows, I think it would be unexpected to not find my language. I would expect to find my language, search, and receive a 'no results' view.

Thoughts?

coreymckrill:
@StevenDufresne I thought about that, and I'm not entirely opposed. But the reason I went with "only show languages for which there is at least one workshop" is that the language list is really a locale list and includes a lot of options that will probably never be assigned to a workshop because a more "formal" locale would most likely be used instead, like `English (Pirate)` and `Spanish (Mexico)`.

StevenDufresne:
I'm thinking the feature image/workshop is more noise than anything when searching on the archive page. After searching, I'm very interested in browsing all the workshops but the feature video takes up all the viewable space.

Thoughts?

PR #110, opened 2020-08-31
Add a searchbar in the subnav menu.

StevenDufresne:
### Description
This PR adds a search to the submenu to bring it alignment with the `wporg-plugins` experience (which this themes' design borrows from) and fixes a dead on the current search page. Once on the search page, there is currently no way to search again.

### Considerations
Ideally, we would be able to refactor the navigation & search form plugin code and reuse it, but since all the code was already loaded in the legacy styles CSS stylesheet, i just updated it here. Additionally, all the css code in the `wporg-plugins` theme is based on a `rem` unit but the theme's base font is `18px` as opposed to `16px`. 

## Screenshots
| Desktop | Mobile |
| --- | --- |
|  |   |


hlashbrooke (approved):
I like it! Nice work with the mobile layout too.

wp-content/themes/pub/wporg-learn-2020/css/components/_search-form.scss
```diff
@@ -0,0 +1,126 @@
...
+.site-header:not(.home) .search-form .search-field {
+	border: 0;
+	border-radius: 2px 0 0 2px;
+	display: inline-block;
+	font-size: 1rem;
+	padding: 5px 10px;
+	position: relative;
+}
+
+@media screen and (min-width: 48em) {
```
ryelle: I'm seeing an awkward search button size at 768px wide — maybe you should use the `$breakpoint-tablet` var, or `breakpoint` mixin here?

Screen Shot 2020-08-31 at 2 31 57 PM

Also the download button flashes visible at just this size, which is odd, but I think it comes from the legacy styles. No change requested for the download button, just a 🤨

ryelle (commented):
Works well with keyboard and a quick VoiceOver test.

I noticed some responsive issues— below about 445px, the search button wraps. Maybe some display:flex would help?

Screen Shot 2020-08-31 at 2 48 03 PM

wp-content/themes/pub/wporg-learn-2020/css/components/_search-form.scss
StevenDufresne: Yeah, I remember a meta ticket about this. I wonder if it has been fixed, and/or the copied CSS for this theme was before that fix.

StevenDufresne:
> Works well with keyboard and a quick VoiceOver test.
> 
> I noticed some responsive issues— below about 445px, the search button wraps. Maybe some display:flex would help?
> 
> Screen Shot 2020-08-31 at 2 48 03 PM

The input/button is a bit fragile (based on how it was originally written & copied) and `display: flex` would be the better implementation but it would introduce a number of issues the button. For now, I'll just shorten the input.

I have earmarked this control as one that we should move into `wporg` as a template-part.

StevenDufresne:
> > Works well with keyboard and a quick VoiceOver test.
> > I noticed some responsive issues— below about 445px, the search button wraps. Maybe some display:flex would help?
> > Screen Shot 2020-08-31 at 2 48 03 PM
> 
> The input/button is a bit fragile (based on how it was originally written & copied) and `display: flex` would be the better implementation but it would introduce a number of issues the button. For now, I'll just shorten the input.
> 
> I have earmarked this control as one that we should move into `wporg` as a template-part.

I couldn't get the existing code to comply during a bizarre breakpoint. 

I rewrote the search form CSS 😬. It should be more reliable now!

ryelle (approved):
working great now!

PR #114, opened 2020-09-03
Plugin: Disable functions that add links/references to GitHub, plus other cleanup

### LINKED ISSUE #85: Lesson plans: Disable functions that add links/references to GitHub
Importing lesson plan markdown from GitHub has been [disabled](https://github.com/WordPress/learn/commit/655065e06c1c3ec46a911d912a559e8c4bfff2c3), but there are places in the plugin, and perhaps the theme, where links and other references to the GitHub repo are showing up. One example is `WPOrg_Learn\LessonPlan::get_markdown_edit_link()`. These should be disabled or removed as well, since the canonical source of the lesson plan content is now the WP site.

coreymckrill:
This started as a simple fix for #85, removing the filters that change post edit links to go to GitHub, since markdown import is now disabled. However, that kind of snowballed into a bunch of other small cleanup and reorg commits.

* Remove the functions that modify edit links.
* Remove the entire `Lesson_Plan` class, since everything in it has now been either relocated or removed entirely.
* Remove the block of CSS that was being injected into every front end page load.
* Move the filter modifying the excerpt length over to the theme.
* Reorganize action/filter additions so they are in the same files where the corresponding functions live.
* Add a namespace to the root plugin file.
* Remove all the shortcode functionality, as it was already commented out and apparently never used.

Fixes #85 

### To Test

Generally make sure nothing is broken.

> Remove the block of CSS that was being injected into every front end page load.

This change means we need to do a visual audit of the front end of the site, in case any theme CSS we have was unknowningly assuming that the injected style block was present.


StevenDufresne (approved):
This is working as expected.

Issue #45, opened 2020-08-10
Workshop template doesn't match


CdrMarks:
I browsed to https://learn.wordpress.org/wp-admin/post.php?post=505&action=edit on production and saw the following message:

>The content of your post doesn’t match the template assigned to your post type.
Screenshot:
![image](https://user-images.githubusercontent.com/3489003/89819696-790c5600-db11-11ea-8edc-8be25b50c576.png)

StevenDufresne:
The `template_lock` that we set for workshops isn't respected. Possible related to https://github.com/WordPress/gutenberg/issues/17262.

If we can get the template lock to work, this shouldn't be an issue.

hlashbrooke:
With the latest updates to the workshop template, this is no longer an issue. Closing this issue.

PR #129, opened 2020-09-30
Workshops: Add an application form and workflow

### LINKED ISSUE #23: Workshop Presenter Application Form
To aid in the vetting and approval of applications to be a Workshop presenter, I'd like to propose a form be setup to create a Workshop Post Type (#3) with an initial, unpublished, post status.  The implementation would be similar to the CPTs in place for WordCamps and Meetups on central.wordcamp.org.  This [WordCamp Application Class code](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wordcamp-application.php) may be useful to review. 

The [current form](https://make.wordpress.org/community/2020/08/03/the-learn-wordpress-workshop-presenter-application-is-here/) asks questions that are good for vetting and approval for the presenter to start creating their video.  The WordCamp organizer form lives at https://central.wordcamp.org/wordcamp-organizer-application/.  For the Workshop presenter application form, I'd propose it live at https://learn.wordcamp.org/workshop-presenter-application/.

I believe there should be a series of Workshop post statuses setup as well.  I've reviewed the list of [WordCamp post statuses](https://github.com/WordPress/wordcamp.org/blob/f9236be586930447f1e15892d8cc311edb123a7d/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php#L166) and propose the following:
* Needs Vetting
* More Information Requested
* Declined
* Approved to Create Video
* Needs Video Review
* Workshop Published
* Workshop Unlisted

The Workshop Unlisted status could have a different name.  It would be used to indicate a Workshop has been removed typically because the content is no longer accurate.

Some questions that might determine other statuses to consider: 
* Should there be a separate status to put a update the title and description?
* Does a Workshop Presenter need to sign an agreement?
* Are there other statuses which should be considered?

coreymckrill:
Adds a custom form for applying to present a workshop. The form fields are based on the [existing form](https://learn.wordpress.org/workshop-presenter-application/) which was built with Jetpack's Contact Form block. This form, however, doesn't require the applicant to enter info like name and email, but instead requires that they log into their wp.org account first, where that data is already available.

The form is output by a `wporg-learn/workshop-application-form` custom block. Styles for the form front end are added to the Learn theme stylesheet. Field validation errors are shown inline with the fields they pertain to.

On the back end, a successful form submission will create a new workshop post with the "Needs Vetting" post status, which is set up via the Edit Flow plugin. Some of the application fields are used to populate the workshop post, while others are just information for the application reviewer. All of the fields from the original application are output in a metabox at the bottom of the Edit Workshop screen.

This is the third recent project I've done that called for validation of some kind of structured data, so I decided to build a tool that can flexibly handle a lot of different use cases, so I can stop reinventing the wheel every time. The result is the `\WordPressdotorg\Validator` class, which lives in wp.org's mu-plugins. Provide the validator instance with a schema for the structured data, run some data through it, and it will output detailed information about anything invalid about the data.

Refs #23 

## Screenshots

form1

form2

## To test

1. Make sure Edit Flow is installed and activated. You can rebuild your local environment with `yarn run create` which should take care of it.
1. Configure Edit Flow. The Custom Statuses module needs to be enabled and applied to the Workshops post type. A "Needs Vetting" status needs to be created.
1. Create a page for the application form to live. Add the Workshop Application Form block to the page and publish.
1. View the new page. While logged in, the application form should be visible. While logged out, you should be prompted to log in.
1. Try to break the form while filling it out. Of particular note are the "Other" open text inputs on the audience and experience level questions.
1. After getting a successful form submission, view the Workshops list table and find the newly created post. The title should be what you put in the "Workshop Title" field. The status should be "Needs Vetting".
1. The new workshop post should have content for the description, learning objectives, and comprehension questions. The language should be added to the Workshop Details metabox. The username of the form submitter should be in the Presenters metabox. All of the form input values should be in an "Original Application" metabox at the bottom of the screen.


StevenDufresne:
Maybe I'm looking in the wrong place but I can't find the workshop anywhere.

Notes:
- The First Name and Last Name fields are 'disabled' but they are still focusable. It took a bit to figure out why. They appeared like broken input fields.
- Some labels are really long and carry on full width. Can we maybe make the form width smaller so the labels and inputs share the same width?
- The select dropdown is clipped

coreymckrill:
@StevenDufresne 

> Maybe I'm looking in the wrong place but I can't find the workshop anywhere

Did you set up the custom "Needs Vetting" status in Edit Flow? If you don't have that, the workshop generated by the form won't appear in the list table.

> The First Name and Last Name fields are 'disabled' but they are still focusable. It took a bit to figure out why. They appeared like broken input fields.

🤔 Those fields are purely FYI so the form user can see what data is associated with their wporg account. I was thinking they should still be focusable so that screen readers will read out the values in them as you're tabbing down. I guess another option would be to output those values in spans or something else that is not an `input` at all.

> Some labels are really long and carry on full width. Can we maybe make the form width smaller so the labels and inputs share the same width?

Good call, should be fixed with 87d04a8

> The select dropdown is clipped

Should be fixed with 3c615a8

StevenDufresne:
> @StevenDufresne
> 
> > Maybe I'm looking in the wrong place but I can't find the workshop anywhere
> 
> Did you set up the custom "Needs Vetting" status in Edit Flow? If you don't have that, the workshop generated by the form won't appear in the list table.

I set up that status but I still don't see it. Maybe there's a problem with my setup? I see the custom statuses on 'pages', but I don't see them on workshops. Hmmm

> 
> > The First Name and Last Name fields are 'disabled' but they are still focusable. It took a bit to figure out why. They appeared like broken input fields.
> 
> 🤔 Those fields are purely FYI so the form user can see what data is associated with their wporg account. I was thinking they should still be focusable so that screen readers will read out the values in them as you're tabbing down. I guess another option would be to output those values in spans or something else that is not an `input` at all.

Since we can't change that information, does it make sense to show it at all? Maybe just the username so they are logged into the right account when they submit? ... 

> 
> > Some labels are really long and carry on full width. Can we maybe make the form width smaller so the labels and inputs share the same width?
> 
> Good call, should be fixed with [87d04a8](https://github.com/WordPress/learn/commit/87d04a8ac5339d20a319370380d66f69e626d92d)

👍 Looks good.
> 
> > The select dropdown is clipped
> 
> Should be fixed with [3c615a8](https://github.com/WordPress/learn/commit/3c615a8406a1319e4feda34eb90184eab4b55d97)

👍 Looks good.

Additionally, there seems to be an issue with the contrast between the required message and the background.


Lastly, the checkboxes are pretty close together, making them a bit harder to hit on mobile. Can we add some padding to the checkbox options? Maybe `padding: 4px 0;`

wp-content/plugins/wporg-learn/inc/form.php
```diff
@@ -0,0 +1,569 @@
...
+				<?php echo wp_kses_data( $args['error_message'] ); ?>
+			</span>
+		<?php endif; ?>
+		<input
+			id="<?php echo esc_attr( $args['id'] ); ?>"
+			name="<?php echo esc_attr( $args['name'] ); ?>"
+			type="<?php echo esc_attr( $args['type'] ); ?>"
+			value="<?php echo esc_attr( $args['value'] ); ?>"
+			<?php echo true === $args['required'] ? 'required' : ''; ?>
+			<?php echo true === $args['readonly'] ? 'readonly' : ''; ?>
```
ryelle, line 535: There's no indication in VoiceOver that these are not editable. Maybe just outputting this info in a span is the better option? The `disabled` attribute could also help.

wp-content/plugins/wporg-learn/inc/form.php
```diff
@@ -0,0 +1,569 @@
...
+		<?php endif; ?>
+		<?php if ( ! empty( $args['error_message'] ) ) : ?>
+			<span class="notice notice-error">
+				<?php echo wp_kses_data( $args['error_message'] ); ?>
+			</span>
+		<?php endif; ?>
+		<textarea
+			id="<?php echo esc_attr( $args['id'] ); ?>"
+			name="<?php echo esc_attr( $args['name'] ); ?>"
+			<?php echo true === $args['required'] ? 'required' : ''; ?>
```
ryelle, line 590: Could this be `readonly` too?

coreymckrill:
@StevenDufresne 

> I set up that status but I still don't see it. Maybe there's a problem with my setup? I see the custom statuses on 'pages', but I don't see them on workshops.

Ah, ok, you have to update the list of post types on the Custom Statuses options tab:

screenshot

To be clear, this will already be configured in Edit Flow on production, so this shouldn't be an issue there. But I guess maybe I should make it fall back to _Draft_ or something in case that custom status is ever changed or disabled...

wp-content/plugins/wporg-learn/js/workshop-application-form/src/edit.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import './editor.scss';
+
+export default function Edit( { className } ) {
+	return (
```
ryelle, line 8: You could use [the `Placeholder` component here](https://wordpress.github.io/gutenberg/?path=/story/components-placeholder--default) & then you wouldn't need to add any editor styles — unless you see needing styles here later?

ryelle (commented):
This is working great. I left a few comments but nothing that'd block merging. I did notice this invalid template warning, but I'm not sure if that was part of my submission or something else:

Screen Shot 2020-10-13 at 2 36 32 PM

wp-content/plugins/wporg-learn/js/workshop-application-form/src/edit.js
coreymckrill: Oh, good idea! Updated here: b162605

wp-content/plugins/wporg-learn/inc/form.php
coreymckrill: I opted to display this data in a table instead: c86fb66

wp-content/plugins/wporg-learn/inc/form.php
coreymckrill: There aren't currently any textareas in the form that should be `readonly`, but I added the option anyway in c86fb66

coreymckrill:
@ryelle 

> I did notice this invalid template warning

I think this is related to #45 which may actually be an upstream issue.

coreymckrill:
@StevenDufresne 

> the checkboxes are pretty close together, making them a bit harder to hit on mobile

Good call, the spacing is tweaked here: 3530a5c

PR #150, opened 2020-11-19
Workshops: Add field to link to a lesson/quiz

### LINKED ISSUE #142: Add field to workshops for quiz URL
Thanks to the updates in #141, we now have usable quizzes on the site :D 

I have manually linked some workshops to their quizzes by adding it into the post content ([example](https://learn.wordpress.org/workshop/intro-to-publishing-with-the-block-editor/)), but it would be great if this could either be a meta field or a custom block that comes as a part of the template. Te only data that will need t be given will be the URL for the quiz, and then it can show a button the frontend to take the quiz. I quite like the block with the button showing up above the content like it is in that example, but it might make more sense for it to be in the sidebar above the 'join a discussion group' button so long as it doesn't get lost in the visuals and can be prominent enough for people to see it.

coreymckrill:
Adds an interface for picking a lesson to link to a workshop. The lesson's post ID is stored as postmeta. Then on the front end, the Workshop Details block (which also outputs things like the duration and the language) will render a button that links to the lesson's quiz. Note that this button is part of the Workshop Details block, which also lists the info about the workshop like length and language.

Fixes #142 

## Screenshots

Metabox for linking a lesson quiz to a workshop:

The quiz button when viewing the workshop:
frontend


hlashbrooke (commented):
This looks great to me - the button placement is logical. My only thought is that maybe it isn't visible enough like it gets lost in the UI? Could we make it a 'primary' button, so it appears blue instead of grey? Could you try that out and see how it looks?

coreymckrill:
@hlashbrooke

hlashbrooke:
I love it - this looks good to go to me!

Issue #116, opened 2020-09-11
Theme: Lesson Plans archive shows some posts twice


coreymckrill:
On the [Lesson Plan archive](https://learn.wordpress.org/lesson-plans/), some of the lesson plans appear twice in the results, not necessarily adjacent to each other. For example, the "Child Themes Twenty Sixteen" lesson plan appears on the first page of the archive, and again on the second page of the archive. I've checked the URLs and they are not separate posts.

StevenDufresne:
Is this still happening? I can't reproduce the example you added.

coreymckrill:
I can't reproduce it now either. Guess I will close this as "works for me"

iandunn:
I can still see it, e.g., https://learn.wordpress.org/lesson-plans/theme/ `dashboard icons` and `static vs content`

StevenDufresne:
Yep, I see it as well, with the last 2 lesson plans on the first page. Nice catch!

coreymckrill:
Huh, I do not see doubles of either of those in the `theme` category. @iandunn are you looking at production or your sandbox?

CdrMarks:
I confirm seeing the duplicates of the ones Ian saw on production.

The first occurrences are seen on https://learn.wordpress.org/lesson-plans/theme/

The duplicate occurrences are seen on https://learn.wordpress.org/lesson-plans/theme/page/2/

iandunn:
Yeah, sorry, I assumed it only happened on pagination boundaries, but didn't mention that explicitly. I was looking at prod.

jrastaban:
I think the issue is that items are being pulled in a random order. I did some testing using ?x=x and the listed items on the page are different each time. It is the same page but if you bust the cache and you get new results.

The outcome is that the second query grabs the last 3 items from a new random query not related to what was seen on page 1 and displays those. A quick solution would be to list them in a specific order though I suspect they are random so the same ones aren't always first.

jrastaban:
I saw that the Workshops are ordered ascendingly so I went ahead and put together a patch that does so for the lesson plans.

jrastaban:
Thanks @coreymckrill for catching that what I thought was a fix wouldn't have changed anything.

~~I did some more digging. The changing order only happens on the taxonomy archive page (https://learn.wordpress.org/lesson-plans/builder/) not on the post type page (https://learn.wordpress.org/lesson-plans/)~~

It looks like this happens on the post type page too but it was harder to replicate (more aggressively cached?)

[...]

PR #160, opened 2020-12-01
Fix for duplicate lesson plans

jrastaban:
Fix for [#116 ](https://github.com/WordPress/learn/issues/116)
Adds an order argument to the lesson plan query so pagination is reliable.


coreymckrill:
@jrastaban Thanks for the patch. Upon closer inspection, it appears that `wporg_get_lesson_plans_by_tax_slugs_query` is actually an orphan function--it isn't used anywhere in the theme. So I don't think this will actually address the issue.

Issue #145, opened 2020-11-12
Add details about discussion groups in workshop pages


timiwahalahti:
_As per the [discussion in Community Team blog](https://make.wordpress.org/community/2020/10/30/getting-more-learn-wordpress-discussion-group-leaders-and-attendees/)._

On a single workshop page, add a small blurb on top of the "Join a Discussion Group" button explaining what is a discussion group. Below the button could be a small text, much like the CoC notice, saying that if there’s no scheduled discussion group for this workshop, apply to be a discussion group leader to run one.

This needs copy and the new "Become a Discussion Group Leader" page being created before fully implementing.

coreymckrill:
Putting this on hold until the "Become a Discussion Group Leader" page is ready.

hlashbrooke:
Here's a page for that: https://learn.wordpress.org/facilitate-a-discussion-group/

coreymckrill:
I think adding another text blurb to the sidebar of every workshop would get kind of messy and distracting. What if the "Join a Group Discussion" button linked to a dedicated page that explained everything and then linked out to the meetup group, and then additionally had all the info about becoming a facilitator?

hlashbrooke:
Fair point yeah.

I think it would be good if we could update that discussion group page to have information about discussion groups at the top, then a block showing the upcoming discussion groups, followed by the form as it stands now.

@coreymckrill Is that 'upcoming discussion groups' block ou made for the home page a block that we could just insert here too? Or is it exclusively tied to the homepage template?

timiwahalahti:
Agree with Corey.

I can work with the copy for that page.

coreymckrill:
> Is that 'upcoming discussion groups' block ou made for the home page a block that we could just insert here too? Or is it exclusively tied to the homepage template?

Right now it's just hard coded into the front page template, but we could make a block for it too.

hlashbrooke:
This is looking great on the homepage! If you're able to make it a block/shortcode then that would be great, but it's not essential if that will be too much to work on here.

coreymckrill:
#163 is a draft PR for adding a block, but it's probably not going to get done before the launch, so I'm going to remove this issue from that milestone.

hlashbrooke:
Works for me - this is lower priority than the other launch items.

hlashbrooke:
[...]

PR #163, opened 2020-12-05
Workshops: Add a block for outputting a list of upcoming social learning spaces

coreymckrill:
WIP

Refs #145


hlashbrooke:
Since we are now using the Meetings calendar plugin, all events (now named "Social Learning Spaces") are actually stored on the site in the `meeting` post type, so this won't require any external API calls and can be gathered from local data. That should make this kind of thing a lot easier to do!

hlashbrooke:
Opened #280 to update the social learning spaces to use the `meeting` post type instead of meetup.com events.

PR #172, opened 2020-12-12
Try to force screen readers to use select instead of select2

alexstine:
Due to the way select2 implements fake selects, it does not allow the HTML label to be assigned to the element causing an accessibility issue. This fix will allow screen readers to use select and select2 to be inaccessible by screen readers. There's a reason why the good old select combo box should be here to stay, it's accessible and it works.

I tried to test this as best I could on Mac, not sure how this will play on Windows because my computer is too old to run Docker. I am thinking accessibility will be a lot better this way.


coreymckrill:
Thanks for the PR @alexstine !

PR #175, opened 2020-12-14
Lesson Plans: Add ID to sort order to prevent "random" sorting.

### LINKED ISSUE #116: Theme: Lesson Plans archive shows some posts twice

ryelle:
Some lesson plans were created at exactly the same second, so ordering by date becomes somewhat random. This is noticeable across pages, where some lesson plans would show up on page 1 and page 2 of the archive due to how the SQL queries work.  In this PR, I've explicitly set the sort order, using date descending and ID ascending, so it will go in order of post creation.

I didn't swap fully to post ID sorting in case users want to change the post dates.

Fixes #116.

To test:

- Check on production since it has the edge-case data
- There should be no repeat lesson plans between pages
- This should apply to /lesson-plans and the lesson plans by category

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -243,16 +243,6 @@ function wporg_archive_modify_query( WP_Query $query ) {
 
 	if ( $query->is_main_query() && $query->is_post_type_archive( $valid_post_types ) ) {
 		wporg_archive_maybe_apply_query_filters( $query );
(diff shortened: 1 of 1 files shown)


jrastaban:
Fascinating! Thanks @ryelle, you are a genius!

Issue #173, opened 2020-12-14
Update copy for the 'Get Involved' block on the home page


hlashbrooke:
This is purely a copy & link update for the home page block that is titled 'Get Involved'. The block title and dashicon image are fine, but the text must be updated to the following:

> Want to get involved in creating the content for Learn WordPress?

Then the link must read:

> Learn how to contribute

And it must link to this URL: https://learn.wordpress.org/contribute/

hlashbrooke:
I tried this out in the browser inspector and it looks like, with the reduced height, the dashicon now gets cut off at the bottom (it was already off-centre since the container height is bigger than the content needs it to be), so ideally this should be vertically centre-aligned somehow:
![Screenshot 2020-12-14 at 16 35 10](https://user-images.githubusercontent.com/1543143/102037808-60b58c00-3dce-11eb-9428-f6e458b5ef84.png)

PR #180, opened 2020-12-15
Update the Get Involved copy on the homepage.

StevenDufresne:
Addresses: #173 

This updates the copy and adjusts the icon height to be dynamic. This removes the constraint that the icon needs to be square making it more flexible.


hlashbrooke (approved):
Perfect.

PR #198, opened 2021-03-18
Plugin: Stop tracking `build/` to improve devex.

### LINKED ISSUE #197: Tracking `build` clutters `git stat | diff | etc`
@coreymckrill , do you remember the reason for tracking `plugins/wporg-learn/build` in this repo?

I'm finding it distracting and frustrating to work with commands like `git stat`, `diff`, etc because most of the output is unrelated & obfuscated. I don't see an obvious benefit to it tracking it, since we're not deploying from this repo directly, but could be missing something.

Would there be any downside to removing it from this repo, but still keeping it in the SVN deploy repo?

iandunn:
Fixes #197

I didn't add anything to the README, since it's in the `create` script, and the user doesn't need to do anything different. Happy to add something, though, just not sure what it'd say.


package.json
```diff
@@ -18,7 +18,7 @@
 	"scripts": {
 		"install:env": "./bin/index.sh",
 		"install:githooks": "mv .git/hooks .git/hooks_orig && ln -s ../.githooks .git/hooks",
-		"create": "composer install && npm run install:env",
+		"create": "yarn workspace wporg-learn-plugin build && composer install && npm run install:env",
```
coreymckrill:
```suggestion
		"create": "yarn workspaces run build && composer install && yarn run install:env",
```

coreymckrill (approved):
You're right, there's not really anything to add to the readme at this point. One correction in the scripts, and then I think this is good to go.

package.json
iandunn: 👍🏻 , thanks!

Issue #195, opened 2021-03-09
Refactor workshop template to increase flexibility


iandunn:
### Background

#186 will require some changes to the block template, to add new elements to the page. Currently, the template is hardcoded in `post_content` via the `template` assigned to the CPT.

https://github.com/WordPress/learn/blob/beacb50998a874fc5f90dd134074c8a59f9b313c/wp-content/plugins/wporg-learn/inc/post-type.php#L142

The template has some "primary" user-generated content like paragraphs/lists describing Learning Outcomes and Comprehension Questions...

https://github.com/WordPress/learn/blob/beacb50998a874fc5f90dd134074c8a59f9b313c/wp-content/plugins/wporg-learn/inc/post-type.php#L169-L207

...and "secondary" content like programatically-generated workshop details, a join button, and a code of conduct link.

https://github.com/WordPress/learn/blob/beacb50998a874fc5f90dd134074c8a59f9b313c/wp-content/plugins/wporg-learn/inc/post-type.php#L208-L236

### Problems

Because of the way it's setup, changes won't apply to existing posts automatically, like they would with a traditional theme's `single-workshop.php` template. We'd have to manually update old posts to use the new template. There have already been some changes to the template, so many posts are using different templates. See https://github.com/WordPress/gutenberg/issues/7961

It also hardcodes the "secondary" content into `post_content`, when it should be applied dynamically. That content isn't supposed to change between posts (other than i18n, which can be handled dynamically). Right now there are non-English posts that have the button in English, because it wasn't translated in `post_content`.

The block template also allows the sidebar content to be removed, because `template_lock` isn't inherited to Column blocks. See https://github.com/WordPress/gutenberg/issues/8112 and https://github.com/WordPress/gutenberg/issues/9386.

### Potential Solutions

We started discussing the issue in https://wordpress.slack.com/archives/C01KC5VDQBC/p1614641586017200 and 
https://wordpress.slack.com/archives/C01KC5VDQBC/p1614645458026200.

It looks like https://github.com/WordPress/gutenberg/pull/26128 helps solve some of the locking issue, by letting us control modifications to a column.

There doesn't seem to be a straightforward solution for storing the "secondary" content outside of `post_content`, though. I looked into how FSE might help that, but it doesn't seem like the beta is far enough along to be useful in production. I'm guessing we'd likely run into problems implementing exactly what we want, and would have to change stuff once the MVP is done. I'd be happy to be corrected about that, though, if anyone sees a way I missed.

[...]

PR #200, opened 2021-03-19
Add transcripts

### LINKED ISSUE #186: Add transcript section to workshop pages
On individual workshop pages ([for example](https://learn.wordpress.org/workshop/user-management/)), it would be helpful to be able to display the video transcript in its own section for viewers to follow along or print for reference.

Would it be possible to add a section (either in the sidebar or a toggle) to display the transcript in full when activated?

In thinking through how to handle transcripts, it will be helpful to also consider cases in which the video is captioned or subtitled in multiple languages.

Just in case a visual is helpful, this is a very rough idea of where we could consider placement: 

transcript UI

iandunn:
Closes #186 

This is a simple approach that just adjusts the PHP-defined template, since #195 didn't work out. The basics are working, but I'm planning to make some other block/template adjustments that will affect this PR, and it'll be better to do those first. I'll open a PR for those later today or next week, and circle back to this after.

I started with @gwwar's idea about putting the transcript below the fold. It seems like dealing with w/ tabs inside the editor would be too complicated, but I might take a second look at that once I circle back.

This assumes that #198 will be merged first.


iandunn:
I've got it working with the transcript at the bottom of the regular content, but I'm wondering if folks would prefer to have it beside the video, so they can scroll through it and watch the video at the same time?

What do you think, @melchoyce, @varlese , @coreymckrill ?

### Below Other Content

Screen Shot 2021-04-26 at 09 03 54-fullpage

----

### Next to video (very rough sketch)

There'd be a show/hide button, and it'd be hidden by default.

Screen Shot 2021-04-26 at 09 12 50-fullpage

melchoyce:
Is the transcript in addition to video captions? If so, we won't need to include it alongside the video, underneath is fine IMO 👍

coreymckrill:
Agree with Mel, below is fine. And since the videos are hosted at wp.tv, I think they do have (at least the possibility) of transcriptions.

wp-content/plugins/wporg-learn/views/block-workshop-details.php
```diff
@@ -17,6 +23,14 @@
 					<span><?php echo esc_html( $value ); ?></span>
 				</li>
 			<?php endforeach; ?>
+
+			<?php if ( $has_transcript ) : ?>
+				<li>
+					<a class="components-button is-secondary is-small" href="#transcript">
```
coreymckrill: This button looks kind of out-of-place in the Workshop Details list. What if it matched the other detail rows more, something like this?

transcript

coreymckrill (approved):
👍 Everything works as expected. One non-blocking quibble with the "Jump to Transcript" button...

wp-content/plugins/wporg-learn/views/block-workshop-details.php
iandunn: I like that, done in 708cf87ca32cb6d665293d3da990984a7d9fd870

Issue #187, opened 2021-02-02
Add language switcher to Learn WordPress for UI translations


varlese:
All static strings for Learn WordPress are currently [available for translation via GlotPress](https://translate.wordpress.org/projects/meta/learn-wordpress/). To be able to display these strings on the front-end of the Learn WordPress site, can we implement a language switcher like the one available on https://wp15.wordpress.net/?

The language switcher should show all locale variants that have 90% or more of the strings translated. (This percentage mimics the [threshold for language packs](https://make.wordpress.org/polyglots/handbook/translating/after-your-contribution/#the-threshold-for-language-pack-creation) for plugins and themes.) Locales with a smaller amount of strings translated, then, can be hidden from the language switcher to help filter available languages for visitors to the site. 

Likewise, adding the ability to search for a language like the example language switcher will help filter choices for site visitors as the number of translated locales increases. We might also want to add the locale abbreviation for clarity, as well - i.e. `Français (fr_FR)`.

<img width="294" alt="language switcher" src="https://user-images.githubusercontent.com/1361760/106633268-a419b480-654c-11eb-8829-2d5c42309f6c.png">

[Related discussion (Polyglots P2).](https://make.wordpress.org/polyglots/2021/01/13/proposal-translating-learn-wordpress/)

tobifjellner:
> ...90% or more of the strings translated. (This percentage mimics the threshold for language packs for plugins and themes.)

On meta projects, translations are usually applied string-by-string. Then there's a delay, consisting of a cron-task for committing changes plus waiting for next deploy. (If there are no commits made for several days, and thus no deploys, then translations won't propagate.)
I'd vote for NOT setting a threshold of 90%. If at least one string is translated, then I'd say that the language is established. (Perhaps, though, it could get some kind of "badge" at the top "This UI is translated to __% and contains some strings in English. You can contribute to the translation at ______(link to language-specific translation project)".

varlese:
My one concern about not setting a threshold is that experience could come across as a little buggy (namely, when you click on the language switcher, if there are only a few strings translated it might seem like nothing changed). Including some sort of badge/tooltip with a call to action, however, sounds like it could both address that _and_ encourage contributions, which I'm a fan of 😄 

Perhaps we can start with implementing the switcher for any language with 1 < strings translated, and then, as a second step, see if we can get some design input for tooltip ideas.

hlashbrooke:
[...]

PR #202, opened 2021-03-24
Add a locale switcher and a notice when a chosen locale isn't fully translated

coreymckrill:
This adds a locale switcher as an mu-plugin. Unlike the one on the [WP15 site](https://wp15.wordpress.net/), this one lives up in the Admin Bar at the top of the screen, which shows on the front end of Learn (and other wp.org sites) even if the visitor isn't signed in. The reasoning for putting the switcher here is:

* It will be in the same place on every page. If it was lower down within the site header, it would have to be in a different place on the homepage than on internal pages.
* It is reusable across all wp.org sites, should we have other places in the future where it makes sense to use it (and we have sufficient translation).

Also unlike the one on WP15, this uses WP/Gutenberg's native `Combobox` component for selecting the locale, rather than the `select2` JavaScript library (or the `selectWoo` variant). The hope is that this will be both more accessible and more maintainable, since the future of development for select2/selectWoo is uncertain.

This also adds a dismissible notice right beneath the site header that shows when a chosen locale has less than 90 percent translation coverage. The notice explains the situation and provides links to the translation project where a visitor could add translations for their chosen locale if they want to. This works similarly to the WP15 site.

One thing this PR doesn't do is set up a system for generating translatable strings for parts of the front end UI that are stored in the database (and thus don't get picked up by the pot file generator). This would mostly be things like taxonomy term labels. This is covered in #205.

Refs #187 

## Screenshots

Screen Shot 2021-03-24 at 15 50 41

The complete flow:
https://cloudup.com/cierJZjLi4L

And here's the flow when choosing a locale that doesn't have a complete translation yet:
https://cloudup.com/c8Gp0ZFZgRO


naokomc:
Very nice! I'm excited to read that it's reusable across all wp.org sites (would be nice to have this on Rosetta, Support Forums,  HelpHub...)

I'm assuming that the combobox only appears when the language link on the top right is clicked, but I'm a little concerned the language swither on the admin bar may not be obvious enough. But since there's browser language detection, that may be ok.

I love the idea of showing a prompt to contribute! But the gray banner may be too blended. The WordPress.org homepage/plugin & theme pages have this green banner – and it's very noticeable.

image

coreymckrill:
> I'm a little concerned the language swither on the admin bar may not be obvious enough.

I have some concerns about this too, but I think this will work for a first iteration. If we later determine we need a locale option to be more prominent on the Learn site, we can add a button somewhere that simply triggers a click on the button in the adminbar. The problem on the Learn site is that there isn't a good, consistent place to have a locale button, other than the site footer, which seems like it might also be non-obvious.

> But the gray banner may be too blended.

Happy to try the green banner color instead. I'll get that updated.

Thanks!

kellychoffman:
> And here's the flow when choosing a locale that doesn't have a complete translation yet

One question: the message that shows up saying the translated is incomplete - I assume that is saying that the language has been switched but a few spots will not be translated? If so, I'd turn it into a warning label instead of a success label so that is more clear. 

Is this the case for a lot of locales? If so, we could preempt this in the modal: Add an `!` icon next to each locale that isn't complete with some copy so we're more up front with this being a possibility.

> I have some concerns about this too, but I think this will work for a first iteration. 

Agree. I think this is a solid iteration. Another future consideration could be to combine the button and the filterable dropdown, so its outside the modal.

coreymckrill:
@kellychoffman @naokomc How about this for the "incomplete translation" notice? Has the standard WP left border in orange, and then also the `:warning:` emoji so that it grabs attention.

locale-notice

kellychoffman:
Is there not a standard warning notice for the site? If no, what you have works.

coreymckrill:
> Is there not a standard warning notice for the site?

Ah, didn't even think to look if the site theme already defined notice styles. It does!

locale-notice

coreymckrill:
> Is this the case for a lot of locales? If so, we could preempt this in the modal: Add an ! icon next to each locale that isn't complete with some copy so we're more up front with this being a possibility.

I tried doing this with the ⚠️ emoji, but it caused some weird behavior in the Combobox component (which I think is related to multibyte characters, but I haven't been able to track down or replicate in Gutenberg's storybook). It also appears that we can't put HTML into the Combobox options because the markup gets escaped, so we wouldn't be able to put any kind of explanatory tooltip or anything in there. Open to other ideas, but I'm thinking for v1 we should just leave this out.

kellychoffman:
> I'm thinking for v1 we should just leave this out.

That works. 

Do we have an approx break down of complete vs. complete locales? Knowing that would help with how to surface this in future iterations, or if it's even something to solve for.

coreymckrill:
> Do we have an approx break down of complete vs. complete locales? Knowing that would help with how to surface this in future iterations, or if it's even something to solve for.

Right now there are only [7 locales](https://translate.wordpress.org/projects/meta/learn-wordpress/) that are at or above the 90% translated threshold. I anticipate this will increase over time, but it's hard to say what percentage will ever make it that far. I think you're right that it should get solved eventually.

PR #203, opened 2021-03-24
Workshops: Move layout elements from `post_content` to theme.

iandunn:
Having them in `post_content` means that we have to manually update all existing posts every time we change the layout.

See #195, https://wordpress.slack.com/archives/C01KC5VDQBC/p1616436392000400

Closes #201 

Blocker for #200

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -376,6 +376,9 @@ function save_workshop_metabox_fields( $post_id ) {
 		return;
 	}
 
+	$video_url = filter_input( INPUT_POST, 'video-url', FILTER_SANITIZE_URL );
+	update_post_meta( $post_id, 'video_url', esc_url_raw( $video_url ) );
+
 	$duration = filter_input( INPUT_POST, 'duration', FILTER_SANITIZE_NUMBER_INT, FILTER_REQUIRE_ARRAY );
 	if ( isset( $duration['h'], $duration['m'], $duration['s'] ) ) {
 		$duration = $duration['h'] * HOUR_IN_SECONDS + $duration['m'] * MINUTE_IN_SECONDS + $duration['s'];
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -68,6 +68,18 @@ function register_workshop_meta() {
 
 	register_post_meta(
 		$post_type,
+		'video_url',
+		array(
+			'description'       => __( "The URL of the Workshop's video.", 'wporg_learn' ),
+			'type'              => 'string',
+			'single'            => true,
+			'sanitize_callback' => 'esc_url_raw',
+			'show_in_rest'      => true,
+		)
+	);
+
+	register_post_meta(
+		$post_type,
 		'duration',
 		array(
 			'description'       => __( 'The duration in seconds of the workshop. Should be converted to a human readable string for display.', 'wporg_learn' ),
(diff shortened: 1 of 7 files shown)


wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -376,6 +376,9 @@ function save_workshop_metabox_fields( $post_id ) {
 		return;
 	}
 
+	$video_url = filter_input( INPUT_POST, 'video-url', FILTER_SANITIZE_URL );
+	update_post_meta( $post_id, 'video_url', esc_url_raw( $video_url ) );
```
coreymckrill: For consistency, I think it would be best to register the `video_url` key up in `register_workshop_meta()`. Then you can set the `sanitize_callback` to `esc_url_raw` and it will get applied in all current and future contexts where a value is saved. (And you wouldn't need to call it separately here.)

wp-content/plugins/wporg-learn/views/metabox-workshop-details.php
```diff
@@ -11,6 +11,18 @@
...
 
+<p>
+	<label><?php esc_html_e( 'WordPress.tv URL', 'wporg_learn' ); ?></label><br />
+	<label for="workshop-video-url">
+		<textarea
+			id="workshop-video-url"
+			name="video-url"
+			class="large-text"
+			rows="4"
+		><?php echo esc_url( $post->video_url ); ?></textarea>
```
coreymckrill, line 22: Why a textarea instead of just a text input?

wp-content/themes/pub/wporg-learn-2020/single-wporg_workshop.php
```diff
@@ -12,7 +12,19 @@
...
+			 * newer posts only store prose there. Meta data, layout, etc is handled in
+			 * `content-workshop-single.php`.
+			 */
+			$layout_hardcoded = has_block( 'core/columns' ) && has_block( 'core/separator' );
+
+			if ( $layout_hardcoded ) {
+				get_template_part( 'template-parts/content', 'workshop-single-hardcoded' );
+			} else {
+				get_template_part( 'template-parts/content', 'workshop-single' );
+			}
```
coreymckrill: Trying to think this through... if someone goes back and edits old workshop posts and removes the deprecated columns, etc., `$layout_hardcoded` would then return `false` and then the workshop would start getting rendered with the new template part right? So this avoids the need for some kind of feature flag?

If that's the idea here, I think it's great! I would maybe recommend including `has_block( 'wporg-learn/workshop-details' ) in the criteria to better ensure all of the deprecated content has been removed.

coreymckrill (approved):
👍 Looks good, and doesn't seem to break anything. A few comments below.

Also, I wonder if while we're at this, we should just get rid of the `'template_lock' => 'all'` property of the workshop CPT entirely. I think we set it originally to make sure that the stuff in the sidebar didn't get accidentally deleted, and now that the sidebar is hard-coded, it seems like more of a hinderance than anything else.

wp-content/plugins/wporg-learn/inc/post-meta.php
iandunn: 👍🏻

wp-content/plugins/wporg-learn/views/metabox-workshop-details.php
iandunn: There's never enough room to display the full URL in a `text` input, even with the `.large-text` class

wp-content/themes/pub/wporg-learn-2020/single-wporg_workshop.php
iandunn: That's right. Good idea about the details block 👍🏻 , added in 4a6a082dee49bdbc19f0d5fa3a2317f831862505

iandunn:
Removed `template_lock` in 4a6a082dee49bdbc19f0d5fa3a2317f831862505

wp-content/plugins/wporg-learn/inc/post-meta.php
iandunn: Done in 013a261

PR #205, opened 2021-04-02
Generate translatable strings for non-static front end UI

coreymckrill:
This adds a script that retrieves taxonomy term names and descriptions from Learn WP via the REST API and writes them to a file that can get picked up by the pot file generator. This way the terms will get imported into the `learn-wordpress` translation project and can get translated by community volunteers.

This new script should get run by a scheduled GitHub action and committed directly back to this repo. The generated file is outside of the wp-content directory, so it should never end up being committed to SVN or otherwise make an appearance on the production server.

This also adds the filter to dynamically translate term names and descriptions when they are output on the front end, and adds a notice on term edit screens so users know that the strings will get translated.

Refs #187

### DIFF
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -26,7 +26,8 @@ jobs:
         run: php ./bin/i18n.php
 
       - name: Commit and push
-        uses: actions-js/push@v1.2
+        # Using a specific hash here instead of a tagged version, for risk mitigation, since this action modifies our repo.
+        uses: actions-js/push@4decc2887d2770f29177082be3c8b04d342f5b64
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: try/generate-i18n # TODO change this to trunk
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -31,3 +31,4 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: try/generate-i18n # TODO change this to trunk
           message: Update translation strings
+# Test again if commit will still happen with no term changes
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -30,5 +30,5 @@ jobs:
         uses: actions-js/push@4decc2887d2770f29177082be3c8b04d342f5b64
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          branch: try/generate-i18n # TODO change this to trunk
+          branch: trunk
           message: Update translation strings
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -31,5 +31,3 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: try/generate-i18n # TODO change this to trunk
           message: Update translation strings
-
-# Test if commit will still happen with no term changes
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,8 +1,8 @@
 name: I18n
 
 on:
-  push: # TODO change this to a schedule
-    branches: [ try/generate-i18n ]
+  schedule:
+    - cron: '0 6,18 * * *'
 
 jobs:
   translation-strings:
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -31,4 +31,3 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: try/generate-i18n # TODO change this to trunk
           message: Update translation strings
-# Test minor change to terms
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -31,3 +31,5 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: try/generate-i18n # TODO change this to trunk
           message: Update translation strings
+
+# Test if commit will still happen with no term changes
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,0 +1,32 @@
+name: I18n
+
+on:
+  push:
+    branches: [ try/generate-i18n ]
+
+jobs:
+  translation-strings:
+    name: Translation strings
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set PHP version
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
+          tools: composer:v2
+
+      - name: Install dependencies
+        run: composer install
+
+      - name: Run translation script
+        run: php ./bin/i18n.php
+
+      - name: Commit and push
+        uses: actions-js/push@v1.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          message: Update translation strings
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -31,4 +31,4 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: try/generate-i18n # TODO change this to trunk
           message: Update translation strings
-# Test again if commit will still happen with no term changes
+# Test minor change to terms
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,7 +1,7 @@
 name: I18n
 
 on:
-  push:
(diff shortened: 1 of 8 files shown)


.github/workflows/i18n.yml
```diff
@@ -0,0 +1,33 @@
...
+  translation-strings:
+    name: Translation strings
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set PHP version
+        uses: shivammathur/setup-php@v2
```
iandunn, line 17: What are the security implications of using third-party code in the context of a commit workflow? 🤔 

If that were compromised, could it add malicious code to the repository, and have that be run by the i18n script on w.org? Is there a simple way to install PHP and Composer directly? 

Reinventing the wheel would definitely suck, though. I've seen some situations where GitHub now lets you do things directly in the YAML file that previously required a separate action, so that may be useful here.

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,33 @@
...
+          tools: composer:v2
+
+      - name: Install dependencies
+        run: composer install
+
+      - name: Run translation script
+        run: php ./bin/i18n.php
+
+      - name: Commit and push
+        uses: actions-js/push@v1.2
```
iandunn: Same security concerns here.

bin/i18n.php
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @return array
+ */
+function get_taxonomies( array $valid_post_types = array() ) {
+	$endpoint = ENDPOINT_BASE . 'taxonomies';
+
+	$response = Requests::get( $endpoint );
+
+	if ( 200 !== $response->status_code ) {
+		die( 'Could not retrieve taxonomy data.' );
```
iandunn, line 25: Should we log this and other errors to Slack so we're aware that it's failing?

bin/i18n.php
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/php
+<?php
+
+namespace WPOrg_Learn\Bin\I18n;
+
+use Requests;
+
+require_once dirname( __DIR__ ) . '/vendor/autoload.php';
```
iandunn, line 8: Does this script get synced to the production server? If so, it be nice to have a check here to make sure it's not executed directly via nginx.

bin/i18n.php
```diff
@@ -0,0 +1,160 @@
...
+	$file_name = 'translation-strings.php';
+	$file_header = <<<HEADER
+<?php
+/**
+ * Generated file for translation strings.
+ *
+ * Used to import additional strings into the learn-wordpress translation project.
+ *
+ * ⚠️ This is a generated file. Do not edit manually. See bin/i18n.php.
+ * ⚠️ Do not require or include this file anywhere.
```
iandunn, line 146: Kudos for the clear warnings here.

wp-content/plugins/wporg-learn/inc/admin.php
```diff
@@ -0,0 +1,57 @@
...
+	$labels   = get_taxonomy_labels( $taxonomy );
+
+	?>
+	<div class="notice notice-info is-dismissible">
+		<p>
+			<?php
+			printf(
+				wp_kses_post( __( '
+					Names and descriptions of %1$s can be translated via the Learn WordPress <a href="%2$s">translation project</a>. Once you have added or changed a term\'s name or description, it may take up to 24 hours before it is available for translation.
+				', 'wporg-learn' ) ),
```
iandunn, line 49: Kudos for the in–app documentation.

iandunn (commented):
Looks good overall, but the security questions might be worth more discussion.

.github/workflows/i18n.yml
coreymckrill: Good question. I'll have to do a little more digging, but my first thought is that we could avoid the risk if we pinned to a specific version of the action that we decide is safe. Presumably you can't change the code of an already-published version.

bin/i18n.php
coreymckrill: I suspect we'd get a notification from GitHub, but I'm not entirely sure. Logging to Slack would potentially be a lot more complicated and involve storing a token in the repo, wouldn't it?

bin/i18n.php
coreymckrill: No, this script is outside of the wp-content directory in this repo, which is the only directory that gets synced to SVN.

.github/workflows/i18n.yml
iandunn: That sounds like a good idea 👍🏻

bin/i18n.php
iandunn: Ah, good point 👍🏻

bin/i18n.php
iandunn: 👍🏻

.github/workflows/i18n.yml
coreymckrill: Thinking about security a bit more here, I don't think there's any viable threat from the PHP action above, because it wouldn't have any way to make a commit to our repo. It could do weird things with PHP or composer, but nothing would be able to leak outside of the sandbox. This action, on the other hand, gets our `GITHUB_TOKEN` which is what allows it to make commits for us. This is the one we'd most likely want to pin to a specific version.

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,33 @@
+name: I18n
+
+on:
+  schedule:
+    - cron: '0 6,18 * * *'
```
coreymckrill, line 5: This does not appear to have run over the weekend, so something must not be configured right here...

coreymckrill: Ah, it looks like scheduled actions will only run on the [most recent commit on the default branch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events). This action doesn't exist on the default branch yet, so...

.github/workflows/i18n.yml
coreymckrill: https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-release-management-for-your-custom-actions

.github/workflows/i18n.yml
iandunn: `shivammathur/setup-php@v2` installs PHP though, right? So it could install a malicious version of PHP, and when `php ./bin/i18n.php` is run, the compromised `file_put_contents()` could append `eval( 'evil payload' )` to the end of ` 'translation-strings.php`. Then w.org would `include` the compromised file, and execute the payload on production. Does that seem accurate?

coreymckrill: > Then w.org would include the compromised file, and execute the payload on production.

This part isn't accurate, because the generated file doesn't exist on production, except during the brief window when the pot file generator [clones this repo](https://dotorg.trac.wordpress.org/browser/wordpress/website/translate/bin/init-repos.sh#L28) in order to statically scan the files for translation strings. The file is never included anywhere during that process and never interpreted as PHP. The rest of the time, the generated file lives in a directory in this repo that is not synced to SVN.

coreymckrill: Er, I guess once the repo has been checked out for the pot file generation, it doesn't get deleted afterward. But it still doesn't get used for anything else or run through a PHP interpreter.

iandunn: Ah, ok, awesome 👍🏻   Then I agree that need to worry about it.

naokomc:
I noticed a Japanese series name is picked up as a translatable string. 
https://translate.wordpress.org/projects/meta/learn-wordpress/ja/default/?filters%5Boriginal_id%5D=11604737

Probably best to limit the scope to English original text only?

coreymckrill:
It might be possible to pick this up when there are non-latin characters, but in general, I think it would be hard to determine whether something is English or not.

Once the [locale switcher](https://github.com/WordPress/learn/pull/202) is deployed, perhaps another option might be to change that term to be in English and then add the Japanese translation...

naokomc:
If there is no other option to flag/hide non-EN terms, it's doable to first create an English one and adding a translation. The downside is that those English terms created for translation will not be used but need to be translated.

Though not super ideal, but strings can be manually hidden on translate.wordpress.org (by someone with higher permission - Polyglots global mentors all probably have this?) too.

coreymckrill:
> The downside is that those English terms created for translation will not be used but need to be translated.

It looks like with series names like **WordPress の基本** and **Editor de Bloques – Tutorial básico** there is the implication that all the workshops in it are in a language besides English, but the actual name of the series doesn't include that because it would be redundant. But it might actually be useful to be explicit about it for an English version of the series name. Would something like this work?

_In English_: WordPress Basics in Japanese
_Translated_: WordPress の基本

PR #214, opened 2021-04-22
Locale-related enhancements for workshops

### LINKED ISSUE #188: Sort content based on language selected
Related to #187 and [this discussion](https://make.wordpress.org/polyglots/2021/01/13/proposal-translating-learn-wordpress/).

Currently, visitors to Learn WordPress can use the filter options on the workshops page to filter content by language. To help create a more holistic, multi-lingual experience, can we explore connecting the content that appears _first_ to the language selected in the language switcher?

For example, if I select ja/Japanese from the language switcher, content in Japanese should appear first on the workshops page, prioritized by content recorded (audio) in Japanese and then content with available Japanese subtitles.

Workshop Filters

In this case, it would be ideal to filter by root language rather than locale variants, unlike the proposal for the language switcher. In other words, filter by all Spanish content (`es_*`) if a visitor selects a Spanish language variant in the language switcher (i.e., `es_VE`). The idea here is that, while the UI translations will be customized according to a visitors' language variant, they will still be able to benefit from content in their language _in general_.

### LINKED ISSUE #189: Show language shortcodes in workshop application
On the [workshop application](https://learn.wordpress.org/workshop-presenter-application/) for Learn WordPress, could we include the language shortcode in the filter/drop-down as well? (For example, if selecting `Ελληνικά` it shows as `Ελληνικά (el)`.)

Screen Shot 2021-02-10 at 16 10 59

With #187, this will keep the language switcher display consistent across the site and will also help folks submitting workshop applications to filter in a variety of ways.

Props to @webtechpooja for [sharing this feedback](https://wordpress.slack.com/archives/C02RW657Q/p1612894087180800).

### LINKED ISSUE #190: Workshop language filters should only display when a workshop is published
On the workshops landing page, the Language filter currently shows a list of languages for which there is a related workshop.

Screen Shot 2021-02-19 at 09 37 32

Currently, the languages that populate this filter are pulled regardless of the workshop post status, i.e. Needs Vetting or Published. This means that, in some cases, if there is only a pending workshop application, the language will show in the filter, but no content is available.

Screen Shot 2021-02-19 at 09 39 23

Instead, we should either:
- change the text from "No workshops found" to something like "Workshops in [locale] coming soon!"
- only show a language in the filter after there is a workshop published.

### LINKED ISSUE #193: Workshops: Filter and sort the list table by language
The list table already has a column for language. Ideally this would be one of the columns where you can click the header to sort by it. A dropdown filter up by the bulk actions would also be nice.

These features would help workshop vetters who are targeting a specific language, so they don't have to search through the entire list.

coreymckrill:
This PR provides a series of fixes and enhancements for issues related to the locale switching that was added in #202.

* [x] Sort the workshops on the archive page based on the chosen locale; Fixes #188 
* [x] Show locale codes (e.g. `en_US`) in the dropdown for the workshop application and in workshop filters; Fixes #189 
* [x] Only show locales in the workshop filters that are for published workshops; Fixes #190 
* [x] Add locale-based filtering and sorting to the workshops list table; Fixes #193

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -291,6 +291,42 @@ function wporg_archive_modify_query( WP_Query $query ) {
 add_action( 'pre_get_posts', 'wporg_archive_modify_query' );
 
 /**
+ * Modify the workshop post type archive query to prioritize workshops in the user's locale.
+ *
+ * @param array    $clauses
+ * @param WP_Query $query
+ *
+ * @return array
+ */
+function wporg_archive_query_prioritize_locale( $clauses, $query ) {
+	if ( ! $query->is_post_type_archive( 'wporg_workshop' ) || is_admin() ) {
+		return $clauses;
+	}
+
+	global $wpdb;
+
+	$locale      = get_locale() ?: 'en_US';
+	$locale_root = preg_replace( '#^([a-z]{2,3}_?)[a-zA-Z_-]*#', '$1', $locale, -1, $count );
+
+	if ( $count ) {
+		$clauses['fields']  .= ", MAX( IF( {$wpdb->postmeta}.meta_key = 'video_language' AND {$wpdb->postmeta}.meta_value LIKE '{$locale_root}%', 1, 0 ) ) AS has_language";
+		$clauses['fields']  .= ", MAX( IF( {$wpdb->postmeta}.meta_key = 'video_caption_language' AND {$wpdb->postmeta}.meta_value LIKE '{$locale_root}%', 1, 0 ) ) AS has_caption";
+		$clauses['orderby'] = 'has_language DESC, has_caption DESC, ' . $clauses['orderby'];
+
+		if ( false === strpos( $clauses['join'], "INNER JOIN {$wpdb->postmeta}" ) ) {
+			$clauses['join'] .= " INNER JOIN {$wpdb->postmeta} ON ( {$wpdb->posts}.ID = {$wpdb->postmeta}.post_id )";
+		}
+
+		if ( false === strpos( $clauses['groupby'], "{$wpdb->posts}.ID" ) ) {
+			$clauses['groupby'] = "{$wpdb->posts}.ID";
+		}
+	}
+
+	return $clauses;
+}
+add_filter( 'posts_clauses', 'wporg_archive_query_prioritize_locale', 10, 2 );
+
+/**
  * Update a query object if filter parameters are present.
  *
  * @param WP_Query $query Query object, passed by reference.
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -344,6 +344,23 @@ function wporg_archive_maybe_apply_query_filters( WP_Query &$query ) {
 	if ( is_array( $filters ) ) {
 		$filters = array_filter( $filters );
 
+		// If both language and captions filters are set, we assume an "OR" relationship.
+		if ( isset( $filters['captions'], $filters['language'] ) ) {
+			$meta_query[] = array(
+				'relation' => 'OR',
+				array(
+					'key'   => $entity_map['captions'],
+					'value' => $filters['captions'],
+				),
+				array(
+					'key'   => $entity_map['language'],
+					'value' => $filters['language'],
+				),
+			);
+
+			unset( $filters['captions'], $filters['language'] );
+		}
+
 		foreach ( $filters as $filter_name => $filter_value ) {
 			switch ( $filter_name ) {
 				case 'search':
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -293,6 +293,30 @@ add_action( 'pre_get_posts', 'wporg_archive_modify_query' );
 /**
  * Modify the workshop post type archive query to prioritize workshops in the user's locale.
  *
+ * In order to show all workshops, but with the ones that are presented/captioned in the user's locale shown first, we
+ * need to modify the posts query in ways that can't be done through the WP_Query or WP_Meta_Query APIs. Instead, here,
+ * we're filtering the individual clauses of the query to add the pieces we need.
+ *
+ * Examples, slightly truncated for simplicity:
+ *
+ * Before:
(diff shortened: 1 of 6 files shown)


wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -290,6 +290,42 @@ function wporg_archive_modify_query( WP_Query $query ) {
...
+ */
+function wporg_archive_query_prioritize_locale( $clauses, $query ) {
+	if ( ! $query->is_post_type_archive( 'wporg_workshop' ) || is_admin() ) {
+		return $clauses;
+	}
+
+	global $wpdb;
+
+	$locale      = get_locale() ?: 'en_US';
+	$locale_root = preg_replace( '#^([a-z]{2,3}_?)[a-zA-Z_-]*#', '$1', $locale, -1, $count );
```
iandunn, line 333: I don't know if it'd work in all cases, but `lang_code_iso_639_1 ` from `locales.php` might make this easier.

wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -290,6 +290,42 @@ function wporg_archive_modify_query( WP_Query $query ) {
...
+ * @return array
+ */
+function wporg_archive_query_prioritize_locale( $clauses, $query ) {
+	if ( ! $query->is_post_type_archive( 'wporg_workshop' ) || is_admin() ) {
+		return $clauses;
+	}
+
+	global $wpdb;
+
+	$locale      = get_locale() ?: 'en_US';
```
iandunn: I don't think the fallback is necessary, except maybe if we're worried about a `locale` callback making it falsy.

https://github.com/WordPress/WordPress/blob/8ef335a566cbf862263c4c6efd4e130ea976b477/wp-includes/l10n.php#L69-L71

iandunn: I'm probably wrong b/c I'm not familiar w/ the locale switching details, but would `get_user_locale()` be better here?

wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -290,6 +290,42 @@ function wporg_archive_modify_query( WP_Query $query ) {
...
+		$clauses['orderby'] = 'has_language DESC, has_caption DESC, ' . $clauses['orderby'];
+
+		if ( false === strpos( $clauses['join'], "INNER JOIN {$wpdb->postmeta}" ) ) {
+			$clauses['join'] .= " INNER JOIN {$wpdb->postmeta} ON ( {$wpdb->posts}.ID = {$wpdb->postmeta}.post_id )";
+		}
+
+		if ( false === strpos( $clauses['groupby'], "{$wpdb->posts}.ID" ) ) {
+			$clauses['groupby'] = "{$wpdb->posts}.ID";
+		}
+	}
```
iandunn, line 357: This would take me a relatively large amount of effort to understand if I ever needed to modify it. I'm wondering if it'd be simpler to do a custom `SELECT` query instead of trying to modify `WP_Query`?

Or maybe do a second query to grab the `video_language` / `video_caption_language` posts, and then use PHP to combine/dedupe/sort them w/ the `WP_Query` result?

If not, then some comments might help others grok it faster. In the docblock, maybe an example of the query before and after modification. And then maybe an inline comment to explain the use of `MAX()`

iandunn (commented):
Accomplishing the desired results is tough to implement, so kudos for making it work 👍🏻

I left a few ideas that might help, but no worries if not.

wp-content/themes/pub/wporg-learn-2020/functions.php
tellyworth: I wondered if `ORDER BY FIELD` might be a clearer option, but I don't think there's a way to do that with a wildcard `LIKE`. So that would probably require storing both the specific locale and the locale root in multiple postmeta fields. And I'm not sure if `WP_Query` even supports `ORDER BY FIELD`.

tellyworth: Ah, it looks like `ORDER BY CASE` might do it: https://stackoverflow.com/questions/3609166/mysql-order-by-like

But again I'm not sure that's even possible with `WP_Query`

wp-content/themes/pub/wporg-learn-2020/functions.php
iandunn: If that doesn't work, then it might be worth opening an issue in the GP repo, and suggest that they add a property for the locale root. I think you're right that using the root will make more sense in a lot of contexts where the locale switcher will be used in the future, so that info could come in handy.

coreymckrill: I checked `lang_code_iso_639_1` (and also `lang_code_iso_639_2`) and unfortunately neither matches `wp_locale` in all cases. So for now, this regex will have to suffice. I suspect adding a `wp_locale_root` property to every locale in GP wouldn't get much traction, but if we start needing this value a lot, I could see having a `get_locale_root()` function that makes this easy.

wp-content/themes/pub/wporg-learn-2020/functions.php
coreymckrill: Ah you're right. Fixed in 518848d

wp-content/themes/pub/wporg-learn-2020/functions.php
iandunn: 👍🏻

wp-content/themes/pub/wporg-learn-2020/functions.php
coreymckrill: In this case I think not, because the Locale Detection plugin filters on the `locale` hook, which gets applied in `get_locale()`, but not necessarily in `get_user_locale()`. I think the user locale option is stored in usermeta and is maybe only used in WP Admin?

wp-content/themes/pub/wporg-learn-2020/functions.php
coreymckrill: I added some docs in 518848d

wp-content/themes/pub/wporg-learn-2020/functions.php
iandunn: 👍🏻

wp-content/themes/pub/wporg-learn-2020/functions.php
iandunn: 👍🏻

PR #215, opened 2021-05-28
Install composer dependencies before building yarn workspaces

psrpinto:
Composer dependencies appear to be required for the `wporg-learn-2020` build to succeed. Without them, the build fails due to SASS files being erased, which results in an error:

```shell
Fatal error: Error: no mixin named breakpoint
        on line 41 of css/components/_featured-workshop.scss
        from line 3 of css/components/_components.scss
        from line 29 of css/style.scss
>> 			@include breakpoint( $breakpoint-tablet ) {
```

I wasn't able to figure out why not having composer dependencies results in the SASS files being erased, but I can consistently reproduce the issue, and can consistently verify that installing composer dependencies before building the yarn workspace fixes the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
 	"scripts": {
 		"install:env": "./bin/index.sh",
 		"install:githooks": "mv .git/hooks .git/hooks_orig && ln -s ../.githooks .git/hooks",
-		"create": "yarn workspaces run build && composer install && yarn run install:env",
+		"create": "composer install && yarn workspaces run build && yarn run install:env",
 		"wp-env": "wp-env",
 		"lint:php": "composer run lint",
 		"format:php": "composer run format"


ryelle:
You're right, the composer dependencies need to be installed first. The setup is a little convoluted. The wporg-learn theme is actually a [child theme](https://developer.wordpress.org/themes/advanced-topics/child-themes/). That means it inherits & extends from the parent theme, which is `wporg`. The wporg parent theme [is installed by composer](https://github.com/WordPress/learn/blob/be6a0ad98d39570917f2d512938baaa2cb2d24a7/composer.json#L50-L51), and then [Grunt does some magic](https://github.com/WordPress/learn/blob/be6a0ad98d39570917f2d512938baaa2cb2d24a7/wp-content/themes/pub/wporg-learn-2020/Gruntfile.js#L5-L24) to pull in the parent theme's CSS files, and everything is built together to result in the theme's stylesheet.

So you're seeing that error because the `breakpoint` mixin is defined [here in `wporg/css/tools/_breakpoint.scss`](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/css/tools/_breakpoint.scss), and without the parent theme installed, it doesn't exist.

This isn't how child themes usually work - the Grunt step here is specific to the wporg themes - but all the wporg child themes use this method. You can tell if it's a child theme by looking for "Template:" in [the theme's main `style.css`](https://github.com/WordPress/learn/blob/be6a0ad98d39570917f2d512938baaa2cb2d24a7/wp-content/themes/pub/wporg-learn-2020/style.css#L11).

ryelle (approved):

PR #216, opened 2021-05-28
Fix issue with database credentials when using wp-env

psrpinto:
Newer WordPress docker images changed database credentials, which results in the following error when running `wp-env start`:

```
mysqlcheck: Got error: 1045: Access denied for user 'example username'@'172.18.0.5'
```

This issue was fixed in https://github.com/WordPress/gutenberg/issues/29752. Upgrading `wp-env` to latest version fixes the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"license": "GPL-2.0-or-later",
 	"private": true,
 	"dependencies": {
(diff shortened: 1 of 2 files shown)


iandunn (approved):
LGTM, thanks!

PR #218, opened 2021-06-10
Use English-only terms in translation-strings.php

audrasjb:
Hi,
I spotted a few translatable strings written in non-english languages. This is pretty confusing for polyglot's translations :)


coreymckrill:
Hi @audrasjb, the file with those taxonomy translation strings is [auto-generated](https://github.com/WordPress/learn/blob/cc52cdb948ffe7eee387112b7d909d088b2c7ba1/bin/i18n.php) by a scheduled GitHub action, so editing it manually won't fix it. One potential solution would be to change the labels for those taxonomy terms to be in English from within the Learn site itself. More discussion on that [here](https://github.com/WordPress/learn/pull/205#issuecomment-814639380).

PR #231, opened 2021-08-20
Add "WordPress Version" and "Included Content" taxonomies

### LINKED ISSUE #224: Taxonomies for content tracking
As per [this proposal](https://make.wordpress.org/training/2021/05/28/proposal-an-audit-tool-for-learn/) for a content audit tool on Learn, we need to implement two new taxonomies that will be made available to the following four post types: `lesson-plan`, `wporg_workshop`, `course` and `lesson`.

The taxonomies are:
* A taxonomy for indicating which WordPress version the content applies to or has been tested up to.
* A taxonomy for which UI elements or dashboard screens the content applies to.

I'm happy for them to be named whatever you feel is appropriate - something like `wp_version` and `interface` would work, but I'm not sure what naming conventions or standards you have for these things. The taxonomies will only need to be registered once and then applied to all of the post types listed above for them all to make use of.

Let me know if you need any further explanation or information.

coreymckrill:
These taxonomies are registered to by visible in WP Admin, including in menus and list tables, but not queryable on the front end.

Fixes #224 

Each of the post types that gets associated with these new taxonomies will have two new columns in their list tables. For several of the post types, and particularly Lesson Plans, this makes the tables very crowded. Perhaps there are some other taxonomies or other columns that we could remove from the list tables to make more room?

lesson-plan-listtable

(Note this is with a fairly wide screen resolution already)

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/taxonomy.php
+++ b/wp-content/plugins/wporg-learn/inc/taxonomy.php
@@ -21,6 +21,8 @@ function register() {
 	register_lesson_level();
 	register_workshop_series();
 	register_workshop_topic();
+	register_wp_version();
+	register_included_content();
 }
 
 /**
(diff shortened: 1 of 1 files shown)


hlashbrooke:
This is great! I'm very happy with this. Regarding the columns - I'm not sure what's most useful or not with lesson plans exactly as I haven't worked directly with them for too long, but individual users are able to choose which columns they want to be displayed or not using the Screen Options, so I don't mind including everything then people can choose what they have displayed.

I also wasn't to check if there are filters for these taxonomies at the top of the list table? Are you able to filter the table by any of the terms in either taxonomy?

coreymckrill:
> I also wasn't to check if there are filters for these taxonomies at the top of the list table? Are you able to filter the table by any of the terms in either taxonomy?

The list table functionality only generates a dropdown for the default Category taxonomy, not any custom ones. You can, however, filter the table by a term by clicking on it within a table row.

hlashbrooke:
OK - that seems like a reasonable solution to me. If we find that we need more dynamic filtering after we work with this, then we can open a new issue with more specific details.

hlashbrooke (approved):

PR #233, opened 2021-08-25
Use a literal percent sign, to make it easier for translators

dd32:
This string is commonly translated to `... 100% by <a href="%s"> ...` by translators, which results in the string not displaying on the site, due to a mismatched printf parameters, and generating lovely PHP Warnings.
```
E_WARNING: printf(): Too few arguments in wp-content/plugins/wporg-learn/views/locale-notice.php:24
```
Screen Shot 2021-08-25 at 12 34 01 pm

When a translator translates it to that, they get a Translation warning, but upon correcting it to `%%` they also get a translation warning, it's a loose:loose situation for the translators.

Since the string doesn't make much difference in English, translating it to be `100%` seems like it'll solve a few issues here.

### DIFF
--- a/wp-content/plugins/wporg-learn/views/locale-notice.php
+++ b/wp-content/plugins/wporg-learn/views/locale-notice.php
@@ -15,9 +15,9 @@ defined( 'WPINC' ) || die();
 		<?php
 		printf(
 			wp_kses_post(
-				/* translators: %s is a URL. */
+				/* translators: %s is a URL. If you translate 'percent' to '%' please encode it as '%%' and discard the GlotPress warning. */
 				__(
-					'⚠️ The translation for this locale is incomplete. Help us get to 100%% by <a href="%s">contributing a translation</a>.',
+					'⚠️ The translation for this locale is incomplete. Help us get to 100 percent by <a href="%s">contributing a translation</a>.',
 					'wporg-learn'
 				)
 			),
--- a/wp-content/plugins/wporg-learn/views/locale-notice.php
+++ b/wp-content/plugins/wporg-learn/views/locale-notice.php
@@ -17,7 +17,7 @@ defined( 'WPINC' ) || die();
 			wp_kses_post(
 				/* translators: %s is a URL. */
 				__(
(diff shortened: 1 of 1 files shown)


coreymckrill:
Just noting that there's a paradox here, in that any locale where this string actually gets translated probably won't ever show this message anyway 😄 

For the locales that aren't translated, this message will appear, and in English, so the double percent is going to be a little bit weird. Could we instead add a `translators` note explaining how to avoid the error messages?

dd32:
A translators note that is basically `If you translate 'percent' to '%', please encode it as %%` would also suffice.

I can go through the existing translations and correct them easily enough, I just felt that something needed changing here :)

wp-content/plugins/wporg-learn/views/locale-notice.php
```diff
@@ -17,7 +17,7 @@
 			wp_kses_post(
 				/* translators: %s is a URL. */
 				__(
-					'⚠️ The translation for this locale is incomplete. Help us get to 100 percent by <a href="%s">contributing a translation</a>.',
+					'⚠️ The translation for this locale is incomplete. Help us get to 100%% by <a href="%s">contributing a translation</a>.',
```
dd32:
```suggestion
					'⚠️ The translation for this locale is incomplete. Help us get to 100 percent by <a href="%s">contributing a translation</a>.',
```

wp-content/plugins/wporg-learn/views/locale-notice.php
```diff
@@ -17,7 +17,7 @@
 			wp_kses_post(
 				/* translators: %s is a URL. */
```
dd32:
```suggestion
				/* translators: %s is a URL. If you translate 'percent' to '%' please encode it as '%%' and discard the GlotPress warning. */
```

dd32 (commented):
Replace the 'percent' => '%' change with an additional translation instruction.

dd32:
I've gone through and updated the translations for the 14 locales (inc 5 spanish dialects) that had translated it "badly".

coreymckrill (approved):

PR #234, opened 2021-08-28
Add an "Expiration" field to the block editor for some post types

### LINKED ISSUE #225: Manual expiration date field
As outlined in the [audit tool proposal](https://make.wordpress.org/training/2021/05/28/proposal-an-audit-tool-for-learn/), we would like to have an additional custom date field for each content type (workshops, lesson plans, lessons and courses) that allows us to specify an 'expiration date' for the content. This wouldn't actually cause the content to expire or be removed, but it would be for internal management of content that needs to be re-checked on a certain date.

coreymckrill:
Adds a control to the "Status & Visibility" panel, just below the Publish date, for adding an expiration date. It functions the same way as setting the publish date: Click the label that either says "No expiration date" or has the current expiration date, which reveals a dropdown with a date picker.

Fixes #225 

## Screenshots

No expiration date set:
expire-1

Choosing an expiration date:
expire-2

With an expiration date set:
expire-3

Filtering the list table for expired posts:
expire-4

## To test

1. Create some posts of one of the supported post types: lesson plans, workshops, courses, or lessons
2. Edit a post and set an expiration date for it that is in the future
3. Go to the list table for that post type and see the "post state" shows how long until it expires
4. Go back and change the expiration date to some time in the past
5. Now the "post state" should be red and say "Expired"
6. Try clicking the "Expired" view link at the top of the list table


hlashbrooke (approved):
This is perfect - love it!

PR #238, opened 2021-09-20
Adding local bio check first, then .org bio.

### LINKED ISSUE #237: Allow speaker bio override from local user profile
Workshops currently display the full bio of individual speakers displayed dynamically from their WordPress.org profiles. In some cases this could potentially become problematic:
* The bio includes swear words
* The bio promotes non-GPL compatible companies or products
* The bio is a nonsense block of text that isn't meant to be taken seriously

In order to provide some control over this situation, could we please add a feature that allows us to override the displayed bio with one added to the user in the Learn WordPress dashboard itself?

The logic flow would then be to check for a bio in the user profile on the site itself and then, only if it does not exist or returns an empty string, then fetch the bio dynamically from profiles.wordpress.org.

ronalfy:
Resolves #237

Reverses the current logic to:

1. Check for the existence of a local bio.
2. If no local bio found, tries to find .org bio.

No UI pieces were added to this PR.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -631,16 +631,18 @@ function wporg_get_workshop_presenters( $workshop = null ) {
  * This relies on the availability of the `bpmain_bp_xprofile_data` table, so for local environments
  * it falls back on values in `usermeta`.
  *
- * @param WP_User $user
+ * @param WP_User $user The user to retrieve a bio for.
  *
  * @return string
  */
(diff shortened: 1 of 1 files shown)


hlashbrooke:
This looks good to me from what I can tell. @coreymckrill could you please review and if you're happy then you can go ahead and merge this.

coreymckrill (approved):
👍 Worked as expected testing on my sandbox!

@ronalfy it would be great to update the doc block for the function to reflect this change in functionality.

Thanks for the PR!

ronalfy:
@coreymckrill you're welcome! I've updated the internal docs.

PR #240, opened 2021-09-20
Add Fallback Social Image (OpenGraph) to Theme

### LINKED ISSUE #166: Placeholder fallback image
At a wider level, Learn could use brand assets that inherit .org but also are unique to Learn.  As a starting point, can we define a fallback image for link previews (when no featured image is present) that includes the WordPress logo but overtly uses "Learn WordPress" and/or the link?

ronalfy:
Resolves #166

This PR adds basic image OpenGraph tags if a featured image is not present.


hlashbrooke:
This looks straightforward to me - I'm not fully sure how to test it properly, so @coreymckrill could you please have a look and merge if it's all good.

coreymckrill (commented):
I think we'll want to host this image on the WordPress.org CDN instead of in the theme directory. Let me figure out where it needs to go and I'll get you a new URL. Stay tuned...

ronalfy:
Thanks @coreymckrill please let me know how to proceed when ready.

coreymckrill (changes requested):
Hey @ronalfy thanks for this PR!

Jetpack is already adding a default `og:image` when there isn't a featured image available, so with this current approach I think we could end up with two `og:image` meta tags on the same page.

WordCamp.org has an [alternative approach](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/jetpack-tweaks/sharing.php#L7) that will probably work here as well: add a filter on the `jetpack_open_graph_image_default` hook and provide an alternate image URL (in our case it will be `https://s.w.org/images/learn-thumbnail-fallback.jpg`). We probably don't need to bother with the separate meta tags for the width and the height.

Note that the whole WordPress.org network has an [mu-plugin](https://github.com/WordPress/wordpress.org/blob/master/wordpress.org/public_html/wp-content/mu-plugins/pub/jetpack-social-metadata.php#L9) that is already filtering that hook at the default priority, so we'll need to set the priority here higher than `10`.

ronalfy:
@coreymckrill thanks for the feedback. PR updated.

coreymckrill (approved):
👍 Looks good!

That filter at the end to enable the OpenGraph output isn't necessary on production, but I can move it to `0-sandbox.php` in a separate commit.

PR #244, opened 2021-10-05
Add custom roles, capability types for lesson plans / workshops

### LINKED ISSUE #220: Custom user roles for Learn WordPress
As part of [this discussion](https://make.wordpress.org/training/2021/05/27/proposal-adding-custom-user-roles-to-learn-wordpress/), I'd like to propose two custom user roles for Learn WordPress. The goal would be to add in these two new user roles to help create more granular user access to better reflect the structure and permissions needed for contributing.

The two new proposed roles would be:
- Workshop Reviewers, which would provide "Deputy"-level access for reviewing workshop applications
- Lesson Plan Editors, which would provide edit access to all Lesson Plans on the site

With this proposal, the full list of user roles on Learn would look like: 

| User Role  | Permissions |
| ------------- | ------------- |
| Author  | Edit access to specific Workshops and specific Lesson Plans |
| Lesson Plan Editors  | Edit access to all Lesson Plans  |
| Editors | Edit access to all Lesson Plans, all Workshops, and Courses | 
| Workshop Reviewers | Edit access to all content types, deputy-level access to meta field for application vetting, and ability to add new users to the site |
| Administrators | Full site access and permissions |

### LINKED ISSUE #223: Add `Review Notes` panel to workshop post type
> Workshop Reviewers: When reviewing Meetup and WordCamp applications, Community Deputies have a custom meta box to share vetting notes. Sometimes this information is confidential, especially when a proposed application/applicant may not meet the program guidelines. The current workflow for evaluating workshop submissions is very similar to this process, and the Learn site would benefit from a role for workshop reviewers to have access to a confidential meta box for communicating review notes with one another.
>
> https://make.wordpress.org/training/2021/05/27/proposal-adding-custom-user-roles-to-learn-wordpress/

#220 is tracking the user role functionality, which can be done independent of adding notes. It will need to be done first, though.

coreymckrill:
Adds two new custom roles, Lesson Plan Editor and Workshop Reviewer, to the Learn site, which necessitates using custom capability types for the Lesson Plan and Workshop post types. Also introduces a custom capability for viewing/managing internal notes on the Workshop post type (enabled by the Internal Notes plugin), because only Admins/Workshop Reviewers should be able to see the notes, not Editors.

Fixes #220 
Fixes #223 

## To test

Lesson Plan Editor role

1. Create a new user on your test site and assign the Lesson Plan Editor role.
2. Log into WP Admin on the site with that user.
3. You should have access to the Lesson Plans menu in the left column, but no other post types.
4. You should be able to add new lesson plans as well as edit and delete existing ones.
5. You should be able to assign terms to lesson plans from any of the associated taxonomites. However, you should not be able to access the screens for adding/deleting/editing terms within those taxonomies.

Workshop Reviewer role

1. You'll need to temporarily install the Internal Notes plugin on your test site for this. You can check it out via [dotorg svn](https://dotorg.svn.wordpress.org/wordpress/website/wp-content/plugins/wporg-internal-notes) if you have access, or get it from [GitHub](https://github.com/WordPress/wporg-internal-notes), but then you'll have to run the build script.
1. Create two new users on your test site: one with the Workshop Reviewer role and one with the normal Editor role.
2. Log into WP Admin as the Workshop Reviewer.
3. You should be able to manage/edit all post types.
4. You should be able to change roles of other users on the test site (on the production multisite this [will allow](https://wordpress.org/support/article/roles-and-capabilities/#promote_users) "Add Existing User")
5. When you edit a workshop post, you should see the icon in the upper right to toggle the Internal Notes sidebar.
6. Now log out and log back in as the Editor user.
7. When editing a workshop post, you should no longer see the icon for the Internal Notes sidebar.

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/capabilities.php
+++ b/wp-content/plugins/wporg-learn/inc/capabilities.php
@@ -87,7 +87,7 @@ function map_meta_caps( $required_caps, $current_cap, $user_id, $args ) {
 				$object = get_post_type_object( $post_type );
 				if ( user_can( $user_id, $object->cap->edit_posts ) ) {
 					$required_caps[] = $object->cap->edit_posts;
-					break 2;
+					break 2; // Breaks out of the foreach and the switch.
 				}
 			}
 
@@ -97,6 +97,7 @@ function map_meta_caps( $required_caps, $current_cap, $user_id, $args ) {
 		case 'read-internal-notes':
 		case 'create-internal-note':
 		case 'delete-internal-note':
+			// Override the meta caps set up in the Internal Notes plugin, specifically for the workshop post type.
 			$parent = get_post( $args[0] ?? 'none' );
 			if ( $parent && 'wporg_workshop' === get_post_type( $parent ) ) {
 				$required_caps = array( 'manage_workshop_internal_notes' );
--- a/wp-content/plugins/wporg-learn/inc/capabilities.php
+++ b/wp-content/plugins/wporg-learn/inc/capabilities.php
@@ -8,6 +8,7 @@ defined( 'WPINC' ) || die();
  * Actions and filters.
  */
 add_filter( 'user_has_cap', __NAMESPACE__ . '\set_post_type_caps' );
+add_action( 'init', __NAMESPACE__ . '\add_or_update_lesson_plan_editor_role' );
 
 /**
  *
@@ -39,4 +40,71 @@ function set_post_type_caps( $user_caps ) {
 	}
 
 	return $user_caps;
+}
+
+/**
+ * Add the Lesson Plan Editor role if it doesn't exist yet, or ensure it has the correct capabilities.
+ *
+ * Once a role has been added, and is stored in the database, it can't be changed using `add_role` because it
+ * will return early.
+ *
+ * @return \WP_Role|null
+ */
+function add_or_update_lesson_plan_editor_role() {
+	$role_caps = get_lesson_plan_editor_role_caps();
+	$lpe_role  = get_role( 'lesson_plan_editor' );
+
+	if ( is_null( $lpe_role ) ) {
+		$lpe_role = add_role(
+			'lesson_plan_editor',
+			__( 'Lesson Plan Editor', 'wporg-learn' ),
+			$role_caps
+		);
+	} else {
+		$caps_to_remove = array_diff(
+			array_keys( $lpe_role->capabilities, true, true ),
+			array_keys( $role_caps, true, true )
+		);
+
+		foreach ( $caps_to_remove as $remove ) {
+			$lpe_role->remove_cap( $remove );
+		}
+
+		$caps_to_add = array_diff(
+			array_keys( $role_caps, true, true ),
+			array_keys( $lpe_role->capabilities, true, true )
+		);
+
+		foreach ( $caps_to_add as $add ) {
+			$lpe_role->add_cap( $add );
+		}
+	}
+
+	return $lpe_role;
+}
+
+/**
+ * Generate a list of capabilities for the Lesson Plan Editor role.
+ *
+ * @return array
+ */
+function get_lesson_plan_editor_role_caps() {
+	$cap_args = array(
+		'capability_type' => array( 'lesson_plan', 'lesson_plans' ),
+		'capabilities'    => array(),
+		'map_meta_cap'    => true,
+	);
+	$cap_map = (array) get_post_type_capabilities( (object) $cap_args );
+
+	$editor_caps = get_role( 'editor' )->capabilities;
+	$role_caps   = array();
+
+	// Same caps as the editor, but only for lesson plans, not posts.
+	foreach ( $cap_map as $primative_cap => $mapped_cap ) {
+		if ( isset( $editor_caps[ $primative_cap ] ) && $editor_caps[ $primative_cap ] ) {
+			$role_caps[ $mapped_cap ] = true;
+		}
+	}
+
+	return $role_caps;
 }
--- a/wp-content/plugins/wporg-learn/inc/capabilities.php
+++ b/wp-content/plugins/wporg-learn/inc/capabilities.php
@@ -150,6 +150,9 @@ function add_or_update_lesson_plan_editor_role() {
 /**
  * Generate a list of capabilities for the Lesson Plan Editor role.
  *
+ * The Lesson Plan Editor should have full editing caps for the Lesson Plan post type, but no other post types.
+ * They can assign/unassign terms from the various taxonomies to lesson plans, but they can't add/edit/delete terms.
+ *
  * @return bool[]
  */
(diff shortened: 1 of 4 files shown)


iandunn:
> You'll need to temporarily install the Internal Notes plugin on your test site for this

I'd assume we'd want it to be installed permanently in the local env; is that not the case?

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * @return mixed
+ */
+function map_meta_caps( $required_caps, $current_cap, $user_id, $args ) {
+	switch ( $current_cap ) {
+		case 'edit_any_learn_content':
+			$required_caps       = array();
+			$learn_content_types = array( 'lesson-plan', 'wporg_workshop', 'course', 'lesson' );
+
+			foreach ( $learn_content_types as $post_type ) {
```
iandunn, line 86:
```suggestion
			// Grant `edit_any_learn_content` when the user has `edit_posts` for any of our custom post types.
			foreach ( $learn_content_types as $post_type ) {
```

WP caps are always convoluted, but I also had to go through each line slowly to figure out what this function was doing. If there isn't a clearer way to structure it, then commenting _what_ it's doing would definitely help.

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+				$object = get_post_type_object( $post_type );
+				if ( user_can( $user_id, $object->cap->edit_posts ) ) {
+					$required_caps[] = $object->cap->edit_posts;
+					break 2;
```
iandunn:
```suggestion
					return $required_caps;
```

In general I think single points of return are simpler, but in this case an explicit `return` here feels much simpler than `break 2`.

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+				}
+			}
+
+			$required_caps[] = 'edit_posts';
```
iandunn:
```suggestion
			$required_caps[] = 'do_not_allow';
```

This achieves the same result, right? If so, then I think it'd be much more obvious this way.

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+			break;
+
+		case 'read-internal-notes':
+		case 'create-internal-note':
+		case 'delete-internal-note':
+			$parent = get_post( $args[0] ?? 'none' );
```
iandunn:
```suggestion
			$parent = isset( $args[0] ) ? get_post( $args[0] ) : false;
```

`'none'` is to prevent `get_post()` from returning the `$GLOBALS['post']`, right? 

If so, this feels more explicit to me.

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+add_filter( 'user_has_cap', __NAMESPACE__ . '\set_caps_for_internal_notes' );
+add_filter( 'map_meta_cap', __NAMESPACE__ . '\map_meta_caps', 20, 4 ); // Needs to fire after meta caps in wporg-internal-notes.
+add_action( 'init', __NAMESPACE__ . '\add_or_update_lesson_plan_editor_role' );
+add_action( 'init', __NAMESPACE__ . '\add_or_update_workshop_reviewer_role' );
+
+/**
+ * Assign custom post type caps to roles based on their equivalents for the 'post' post type.
+ *
+ * Example: If a user has the `edit_others_posts` cap (from Editor role), this will also give them
+ * the equivalent `edit_others_lesson_plans` and `edit_others_workshops` caps.
```
iandunn, line 20: The example is a really helpful comment, kudos 👍🏻

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+		$cap_args = array(
+			'capability_type' => $capability_type,
+			'capabilities'    => array(),
+			'map_meta_cap'    => true,
+		);
+		$cap_map = (array) get_post_type_capabilities( (object) $cap_args );
+
+		foreach ( $user_caps as $cap => $bool ) {
+			if ( $bool && isset( $cap_map[ $cap ] ) ) {
+				$user_caps[ $cap_map[ $cap ] ] = true;
```
iandunn:
```suggestion
		foreach ( $user_caps as $cap_slug => $granted ) {
			if ( $granted && isset( $cap_map[ $cap_slug ] ) ) {
				$user_caps[ $cap_map[ $cap_slug ] ] = true;
```

`$cap` feels ambiguous to me, and `$bool` describes the data type, rather than what the variable means.

wp-content/plugins/wporg-learn/inc/capabilities.php
```diff
@@ -0,0 +1,237 @@
...
+	return $wr_role;
+}
+
+/**
+ * Generate a list of capabilities for the Workshop Reviewer role.
+ *
+ * The Workshop Reviewer should have all the same caps as the Editor role, with the addition of `promote_users`
+ * (normally reserved for the Admin role), so that they can add workshop presenters as new users on the site.
+ *
+ * This also gives them the cap to manage internal notes on workshop posts. (See `set_caps_for_internal_notes` above.)
```
iandunn, line 229: Those are helpful comments, thanks!

iandunn (approved):
LGTM overall 👍🏻 

Caps are always hard to work with, so I left a few suggestions that might make things a bit more clear

wp-content/plugins/wporg-learn/inc/capabilities.php
coreymckrill: 👍 Good point

wp-content/plugins/wporg-learn/inc/capabilities.php
coreymckrill: Feels wrong to return early, but I added a comment to clarify what we're breaking out of in bf24d13

wp-content/plugins/wporg-learn/inc/capabilities.php
coreymckrill: Correct. Although your revision wouldn't prevent returning the global in all cases, because `get_post` uses `empty()` as the check, so if the arg was set but with something false-y, we'd still get it. I used a variation of your suggestion in b7f90c1

coreymckrill:
> I'd assume we'd want it to be installed permanently in the local env; is that not the case?

🤔 I suppose that it should be in there, since we have code in this codebase that is specifically for it (though nothing will fatal if it's not there). I'll add it in a separate PR, though.

PR #246, opened 2021-10-08
Add 'Other Contributors' field to workshop

### LINKED ISSUE #242: New 'Other Contributors' field for workshops
As per the conclusion of the discussion here: https://make.wordpress.org/training/2021/09/21/should-learn-wordpress-contributors-be-gpl-compliant/#comment-3159

Please add a new field to workshops labelled "Other Contributors" that can take a comma-separated list of WordPress.org usernames. These will be contributors to the workshop who are not seen/heard in the video itself.

On the front end, these will be listed directly below the current area where workshop presenters are displayed with the title "Other Contributors". Each contributor must be listed only with their name that links to their WordPress.org profile (no bio, avatar or anything else).

psrpinto:
Fixes #242

## What can still be improved
- In the admin, the `Other contributors` meta box is displayed before the `Presenters`, even though, in the code it appears after. Not sure why that is or how to fix.
- ~~In the frontend, the styling could probably be improved. For example, there's too much white space, IMHO. It's also possible that I haven't used the correct markup, which could explain the excessive whitespace.~~

### Admin
backend

### Frontend
Screenshot 2021-10-11 at 14 52 19


ashfame:
Reviewed it, LGTM!

wp-content/themes/pub/wporg-learn-2020/template-parts/content-workshop-single.php
```diff
@@ -81,6 +81,28 @@ class="wp-block-button__link"
 				</section>
 			<?php endforeach; ?>
 
+			<section class="row workshop-page_section"">
+				<div class="col-4">
+					<h2><?php esc_html_e( 'Other Contributors', 'wporg_learn' ); ?></h2>
+				</div>
+			</section>
```
coreymckrill: This probably doesn't need all the wrapping elements. It might be fine with just the `h2`. Related, I wonder if there should also be a "Presenters" `h2` above the list of presenters. It looks kind of weird to me to have a heading for one but not the other. Thoughts about that @hlashbrooke?

wp-content/themes/pub/wporg-learn-2020/template-parts/content-workshop-single.php
```diff
@@ -81,6 +81,28 @@ class="wp-block-button__link"
...
+						get_template_part(
+							'template-parts/component',
+							'workshop-other-contributor',
+							array(
+								'other_contributor' => $other_contributor,
+							)
+						);
+						?>
+					</div>
+				</section>
```
coreymckrill: Since these "other" contributors are supposed to be listed "only with their name that links to their WordPress.org profile (no bio, avatar or anything else)", it might look better to display them as a single comma-separated list string... in which case we probably wouldn't need the `component-workshop-other-contributor.php` file at all.

coreymckrill (commented):
@psrpinto thanks so much for this PR! The code looks good, though I agree that there's excessive whitespace and maybe the markup and layout could be tweaked a bit (see inline comments below).

wp-content/themes/pub/wporg-learn-2020/template-parts/content-workshop-single.php
hlashbrooke: Agreed - add a 'Presenters' heading above the first list with the full bios, etc. and then 'Other Contributors' for this second section.

wp-content/themes/pub/wporg-learn-2020/template-parts/content-workshop-single.php
hlashbrooke: Yeah - a comma-separated list would be best here.

wp-content/themes/pub/wporg-learn-2020/template-parts/content-workshop-single.php
psrpinto: Makes sense, done :)

wp-content/themes/pub/wporg-learn-2020/template-parts/content-workshop-single.php
psrpinto: 👌

psrpinto:
I've addressed remarks and updated the screenshot in the PR description. The whitespace situation didn't change much though, I'm available to try to improve it if requested :)

coreymckrill:
@psrpinto Thanks! I tried tweaking the markup a little bit to try and fix the whitespace, but it looks like it will be more complicated because of the... "interesting" grid framework that the theme is using. I'd rather not tie up your time with a bunch more back-and-forth on this, so I think I'll just add more commits directly to this PR when I figure it out.

coreymckrill:
^ With the above commit:

contributors

hlashbrooke (commented):
This looks great to me!

hlashbrooke (approved):

psrpinto:
Thanks @coreymckrill, looks great!

PR #247, opened 2021-10-08
 define a print media stylesheet for lessons and workshops pages 

### LINKED ISSUE #148: Print style sheet for lesson plans
For lesson plans especially, can we have a print style sheet?

### LINKED ISSUE #229: Print stylesheet for workshops
Related to #148 and #155.

Could we add a print stylesheet for workshops? This would be especially useful as we are going to be adding full transcripts for all videos going forward, so having everything in the post content (aside from the actual video) in a printable view would be great. The print stylesheet could exclude the sidebar.

ashfame:
I took a very lean approach of only showing the relevant stuff and hiding everything else, including the header, navigation, footer & sidebar. This requires good testing with elaborated content, I only had little content that I entered manually.

Few things:
- I don't like committing files that are generated automatically, but since existing styles are committed in repo, I committed `print.css` as well.
- `print.css` is not minified, as I am not sure how to set that up in Gruntfile.

Current plan is to iterate over edge cases post deployment to production, where a lot of elaborated content would be a good test bench for it.

This fixes #148 and #229


wp-content/themes/pub/wporg-learn-2020/css/print.scss
```diff
@@ -0,0 +1,136 @@
+// Set custom margin
+@page {
+	margin: 1cm;
+}
+// Hide stuff that's not relevant on print media
+#wporg-header,
+#wporg-footer,
+#head-search,
+#wporg-header-menu,
+.site-header {
```
coreymckrill: A couple other things to hide:

```suggestion
#wporg-header,
#wporg-footer,
#head-search,
#wporg-header-menu,
.site-header,
.workshop-page_video,
.wporg-learn-locale-notice {
```

wp-content/themes/pub/wporg-learn-2020/css/print.scss
```diff
@@ -0,0 +1,136 @@
...
+}
+// Avoid page break in mid of elements
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+ul,
+img {
```
coreymckrill:
```suggestion
h1,
h2,
h3,
h4,
h5,
h6,
ul,
pre,
img {
```

wp-content/themes/pub/wporg-learn-2020/css/print.scss
```diff
@@ -0,0 +1,136 @@
...
+// Handle super-long piece of text, like URLs that are too big to fit on a single line,
+// by breaking them instead of bleeding over the edge
+p a {
+	word-wrap: break-word;
+}
+// Set the URL along with breadcrumbs, wherever available
+.bbp-breadcrumb::after {
+	content: "learn.wordpress.org";
+	float: right;
+}
```
coreymckrill: 🤔 Instead of showing the breadcrumb, but hiding the site title, what if we switched that? Hide the breadcrumb, but show the site title at the top, along with this URL? Then there would be something at the top of workshops too.

wp-content/themes/pub/wporg-learn-2020/css/print.scss
```diff
@@ -0,0 +1,136 @@
...
+	.workshop-page_sidebar {
+		display: none;
+	}
+
+	article.wporg_workshop section {
+		padding: 0;
+	}
+
+	.wp-block-column {
+		flex-grow: 1;
```
coreymckrill, line 167: This seems to work in Chrome for expanding the content column to the full page width, but it doesn't look like it's working in Firefox...

coreymckrill (commented):
Thanks for this PR @ashfame! 👍 Looking really good, especially in Firefox. Chrome seems to have less respect for the rules about not breaking the page in specific places like after headings or the middle of some elements 😕 

Aside from the comments below, one other thing I noticed, in both FF and Chrome, is that on single posts (both lesson plans and workshops), the post title gets centered, while everything else stays aligned left. I don't see any rules in the print stylesheet that would cause that, so maybe it's a browser default in print view? I think the post titles should probably be aligned left, even if we have to add explicit rules to do that.

wp-content/themes/pub/wporg-learn-2020/css/print.scss
ashfame: Definitely worked for me in Firefox v93 as that's what I used. Perhaps because of some other content in conjunction it doesn't? I didn't test anything in Chrome. I thought I will do it in the end but then forgot :sweat:

wp-content/themes/pub/wporg-learn-2020/css/print.scss
ashfame: The markup was so different for each view, that I didn't try hard enough to achieve this. I will take another look.

wp-content/themes/pub/wporg-learn-2020/css/print.scss
ashfame: Works fine on both - Firefox & Chrome!

wp-content/themes/pub/wporg-learn-2020/css/print.scss
ashfame: Done! Ready for another look.

ashfame:
@coreymckrill Yeah, Chrome does seem to have less respect towards page break governing CSS properties. The page title being centered in Firefox, its just the effect of default margin, but when its set to none, it goes to left on its own. Chrome on the other hand, doesn't do it & has less respect for our defined margins as well. But I believe, its a trivial thing and heading of the page being centrally aligned in Chrome isn't a big pain to put more time into.

If everything looks ok here, I would love to get this deployed and then check things once with the live data.

wp-content/themes/pub/wporg-learn-2020/css/print.scss
coreymckrill: This looks great now!

wp-content/themes/pub/wporg-learn-2020/css/print.scss
coreymckrill: 🤔 I'm still seeing this issue in FF v93. But I'm fine to go ahead and merge this for now and then iterate.

ashfame: I noticed the issue on live and its still not reproducible on my local for some reason.
This rule gets applied on live `.wp-block-columns:not(.is-not-stacked-on-mobile) > .wp-block-column[style*="flex-basis"]` which makes `flex-grow: 0` and if that rule is removed, it all works. There is already a `.wp-block-columns:not(.is-not-stacked-on-mobile) > .wp-block-column` rule that makes `flex-grow: 1` but I have no context of how this works. Perhaps an easy fix is to just redefine the rule in print stylesheet? Or alternatively make the existing stylesheet not apply on print media at all, which might make us handle typography on our end from start. Just wanted to post my findings but unsure of whats the best way to move forward.

ashfame: @coreymckrill Just want to make sure you have seen my comment above, but I am unsure of how to fix it.

coreymckrill: @ashfame I think this rule is getting added by the Gutenberg plugin, so my guess is that you have an older version of Gutenberg on your local dev environment?

Either way, I think your idea to redefine the rule in the print stylesheet is probably the best option.

ashfame: @coreymckrill Repo has Gutenberg `v11.5.1` so perhaps that needs updating? I made the change to cover that class explicitly in https://github.com/WordPress/learn/pull/253

coreymckrill: Looks great! Thanks for following up.

PR #248, opened 2021-10-09
Lesson Plans: Add a "series" taxonomy and related functions and templates

### LINKED ISSUE #147: Lesson Plans: Create 'series' taxonomy to group lesson plans together
Lesson Plan + Lesson Plan + Lesson Plan = Workshop (the term the team originally used before Community Video Workshops were created).  We can adjust the word "workshop" but the idea here remains for filtering https://wp.invisionapp.com/share/BNP0V7QXRF8#/screens/330323718

coreymckrill:
Adds a new "series" taxonomy for lesson plans that mirrors the series taxonomy/functionality for workshops.

Fixes #147


coreymckrill:
@hlashbrooke if we use the same "Series" taxonomy for lesson plans that we're already using for workshops, it would mean that both post types would make use of the same pool of term names, and a series could potentially have both workshops and lesson plans in it. Is that desirable?

It will probably also make things a little more complicated in the theme templates. The taxonomy archive view for Series, for example, would mix workshops and lesson plans together, unless we specifically customize the template to show them separately.

coreymckrill:
Just realized, it would also mean that if you are on the Lesson Plan archive view, and you want to filter by series, it would include terms that are assigned to workshops, but not lesson plans. There probably wouldn't be a good way to avoid that.

hlashbrooke:
Hmm - that's a good point! Might be better to add a new taxonomy for lesson plans then I suppose. If that's not a major change at this point, could you head in that direction?

PR #249, opened 2021-10-12
Import test content from live site via REST

### LINKED ISSUE #792: Question: can raw post data be safely made public?
There is an issue #194 for the creation of export data from learn.wordpress.org for the development environment. The idea being that anyone who wants to develop or test a change to this repo can run an environment locally _with a copy of real post data from production_. There are a couple of ways of implementing this (see #249), but resolving it either way requires answering an important question:

**Can we safely export raw post content (lesson plans, workshops, etc) from production without leaking sensitive data?**

Obviously PII metadata like author emails, comment IPs and so on will need to be redacted. I'm referring here specifically to the raw Gutenberg blocks in Lesson Plan and Workshop posts that would normally only be visible to admin users via the post editor in wp-admin. Ordinarily the raw blocks are not made public, only the rendered output. Solving the test data issue requires making exposing those raw blocks to the public one way or another.

Are there any Blocks in use that could potentially contain something in its content or attributes that needs to be kept private, like an email address or API key? Could that become an issue in future? Is there something we could do to mitigate the risk of inadvertently including sensitive data in Block data?

tellyworth:
This imports local content for testing purposes, fetched from the REST API on the live learn web site.

Using the REST API has some shortcomings and advantages over other methods such as WXR:
* It probably doesn't include any PII or sensitive data
* It might miss some important metadata or formatting

Still WIP.

### DIFF
--- a/bin/import-test-content.php
+++ b/bin/import-test-content.php
@@ -1,0 +1,75 @@
+#!/usr/bin/php
+<?php
+
+namespace WPOrg_Learn\Bin\ImportTestContent;
+
+/**
+ * CLI script for generating local test content, fetched from the live learn.wordpress.org site.
+ *
+ * This needs to be run in a wp-env, for example:
+ *
+ * yarn run wp-env run cli "php bin/import-test-content.php"
+ */
+
+// This script should only be called in a CLI environment.
+if ( 'cli' != php_sapi_name() ) {
+    die();
+}
+
+
+$opts = getopt( '', array( 'post:', 'url:', 'abspath:', 'age:' ) );
+
+require dirname( dirname( __FILE__ ) ) . '/wp-load.php';
+
+
+function sanitize_meta_input( $meta ) {
+	$meta = (array( $meta ) );
+	foreach ( $meta as $k => $v ) {
+		if ( is_array( $v ) ) {
+			$meta[ $k ] = implode( ',', $v );
+		}
+	}
+
+	return $meta;
+}
+
+function import_rest_to_posts( $rest_url ) {
+	$response = wp_remote_get( $rest_url );
+	if ( is_wp_error( $response ) ) {
+		die( $response->get_error_message() );
+	}
+
+	$body = wp_remote_retrieve_body( $response );
+	$data = json_decode( $body );
+
+	foreach ( $data as $post ) {
+		echo "Got {$post->type} {$post->id} {$post->slug}\n";
+
+
+		// Surely there's a neater way to do this.
+		$newpost = array (
+			'import_id' => $post->id,
+			'post_date' => date( 'Y-m-d H:i:s', strtotime($post->date) ),
+			'post_name' => $post->slug,
+			'post_title' => $post->title,
+			'post_status' => $post->status,
+			'post_type' => $post->type,
+			'post_title' => $post->title->rendered,
+			'post_content' => $post->content->rendered, // TODO: can we re-parse this with parse_plocks() / serialize_block() etc?
+			'post_parent' => $post->parent,
+			'comment_status' => $post->comment_status,
+			'meta_input' => sanitize_meta_input( $post->meta ),
+		);
+
+		$r = wp_insert_post( $newpost, true );
+
+		if ( is_wp_error( $r ) ) {
+			die( $r->get_error_message() );
+		}
+
+		echo "Inserted $post->type $post->id as $r\n";
+	}
+}
+
+import_rest_to_posts( 'https://learn.wordpress.org/wp-json/wp/v2/wporg_workshop' );
+import_rest_to_posts( 'https://learn.wordpress.org/wp-json/wp/v2/lesson-plan' );
--- a/bin/import-test-content.php
+++ b/bin/import-test-content.php
@@ -13,7 +13,7 @@ namespace WPOrg_Learn\Bin\ImportTestContent;
 
 // This script should only be called in a CLI environment.
 if ( 'cli' != php_sapi_name() ) {
-    die();
+	die();
 }
 
 
@@ -22,11 +22,16 @@ $opts = getopt( '', array( 'post:', 'url:', 'abspath:', 'age:' ) );
 require dirname( dirname( __FILE__ ) ) . '/wp-load.php';
 
 if ( 'local' !== wp_get_environment_type() ) {
-	die( "Not safe to run on " . get_site_url() );
+	die( 'Not safe to run on ' . esc_html( get_site_url() ) );
 }
 
+/**
+ * Sanitize postmeta from the rest API for the format required by wp_insert_post.
+ *
+ * @return array An array suitable for meta_input.
+ */
 function sanitize_meta_input( $meta ) {
-	$meta = (array( $meta ) );
+	$meta = array( $meta );
 	foreach ( $meta as $k => $v ) {
 		if ( is_array( $v ) ) {
 			$meta[ $k ] = implode( ',', $v );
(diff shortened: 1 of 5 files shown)


tellyworth:
This now includes server-side code to add a new `export` context to the rest API, which adds an extra `content_raw` field containing the raw blocks.

To test this end-to-end you'd need to add the export code to a sandbox, then run an import on a local development box. You can test the REST API part locally if content exists by fetching the endpoints and checking for the `content_raw` items in json:

```
http://localhost:8888/wp-json/wp/v2/wporg_workshop?context=export
http://localhost:8888/wp-json/wp/v2/lesson-plan?context=export
```

I think the remaining work to do is in import-test-content.php. @iandunn if you can suggest a better and more complete way to insert the content than [what I've done here](https://github.com/WordPress/Learn/pull/249/files#diff-2b807766c529810531819f12d8b732bb1ce5117b50acb57ba6f6672147649548R57-R72) I'd love to hear your ideas.

bin/import-test-content.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	$body = wp_remote_retrieve_body( $response );
+	$data = json_decode( $body );
+
+	foreach ( $data as $post ) {
+		echo "Got {$post->type} {$post->id} {$post->slug}\n";
+
+
+		// Surely there's a neater way to do this.
```
iandunn, line 65: Were you thinking there might be a more performant way to insert them? This seems fine to me for this kind of situation 👍🏻  It's straight-forward, and I'm guessing it doesn't take too long to run.

If it becomes an issue, we could look and see how more modern importers handle it, like the one Ryan McCue proposed for Core, or Gary's work on WPCOM importers.

bin/import-test-content.php
```diff
@@ -0,0 +1,83 @@
...
+			'post_status' => $post->status,
+			'post_type' => $post->type,
+			'post_title' => $post->title->rendered,
+			'post_content' => ( $post->content_raw ?? $post->content->rendered ),
+			'post_parent' => $post->parent,
+			'comment_status' => $post->comment_status,
+			'meta_input' => sanitize_meta_input( $post->meta ),
+		);
+
+		$r = wp_insert_post( $newpost, true );
```
iandunn:
```suggestion
		$new_post_id = wp_insert_post( $newpost, true );
```

wp-content/plugins/wporg-learn/inc/export.php
```diff
@@ -0,0 +1,109 @@
...
+function register_raw_content_for_post_type( $post_type ) {
+
+	register_rest_field(
+		$post_type,
+		'content_raw',
+		[
+			'get_callback' => __NAMESPACE__.'\show_post_content_raw',
+			'schema'       => [
+				'type' => 'string',
+				'context' => [ 'export' ]
```
iandunn:
```suggestion
				'context' => [ 'wporg_export' ]
```

This might be a bit more future-proof, in case Core adds an `export` context in the future.

wp-content/plugins/wporg-learn/inc/export.php
```diff
@@ -0,0 +1,109 @@
...
+			]
+		]
+	);
+
+	add_filter( "rest_{$post_type}_item_schema", __NAMESPACE__.'\add_export_context_to_schema' );
```
iandunn: What's the benefit of the `add_export_context_to_schema` wrapper, as opposed to registering `update_schema_array_recursive` as the callback?

wp-content/plugins/wporg-learn/inc/export.php
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return array_unique( $block_names );
+}
+
+function show_post_content_raw( $object, $field_name, $request ) {
+	// Blocks that are allowed to be revealed via the export context.
+	// Posts that contain any other blocks will not expose raw content.
+	$allowed_blocks = [
+		// core/* assumed safe by default; is this a reasonable assumption?
```
iandunn: 🤔 , I'd probably err on the side of caution and safelist them too. It's hard do have confidence about that without manually checking each one, and we don't know what will be added in the future.

wp-content/plugins/wporg-learn/inc/export.php
```diff
@@ -0,0 +1,109 @@
...
+		$blocks = parse_blocks( $post->post_content );
+		$block_names = get_all_block_names( $block_names );
+
+		foreach ( $block_names as $block_name ) {
+			// Allow all core blocks
+			if ( 'core/' === substr( $block_name, 0, 5 ) ) {
+				continue;
+			}
+			// If it contains a disallowed block, then return no content.
+			// Better to raise an error instead?
```
iandunn, line 142: Maybe replace it with a `wp:paragraph` that says something like:

> The content of this block has been redacted because it might contain sensitive information.

wp-content/plugins/wporg-learn/inc/export.php
```diff
@@ -0,0 +1,109 @@
...
+		}
+		if ( 'context' === $key && in_array( 'view', $value ) ) {
+			$value[] = 'export';
+		}
+	}
+}
+
+function get_all_block_names( $blocks ) {
+	$block_names = [];
+
```
iandunn:
```suggestion
    if ( ! $blocks ) {
        return array();
    }
```

I got an error on my sandbox w/ out this, because `$blocks` was `null` for https://learn.wordpress.org/wp-json/wp/v2/wporg_workshop?context=export

The workshop content is in the response, though, so I'm guessing it's just empty in one of the recursive calls on some posts, or something like that.

wp-content/plugins/wporg-learn/inc/export.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+/**
+ * Allow some raw data to be exposed in the REST API for certain post types, so that developers can import
+ * a copy of production data for local testing.
```
iandunn, line 5:
```suggestion
 * a copy of production data for local testing.
 *
 * ⚠️ Be careful to only expose public data!
```

bin/import-test-content.php
```diff
@@ -0,0 +1,83 @@
...
+$opts = getopt( '', array( 'post:', 'url:', 'abspath:', 'age:' ) );
+
+require dirname( dirname( __FILE__ ) ) . '/wp-load.php';
+
+// Refuse to run on a production site.
+$run_on_hosts = [
+	'localhost',
+	'127.0.0.1',
+];
+if ( !in_array( parse_url( get_site_url(), PHP_URL_HOST ), $run_on_hosts ) ) {
```
iandunn:
```suggestion
if ( 'local' !== wp_get_environment_type() ) {
```

This might be more consistent with how our other codebases do similar checks.

bin/import-test-content.php
```diff
@@ -0,0 +1,83 @@
...
+
+// This script should only be called in a CLI environment.
+if ( 'cli' != php_sapi_name() ) {
+    die();
+}
+
+
+$opts = getopt( '', array( 'post:', 'url:', 'abspath:', 'age:' ) );
+
+require dirname( dirname( __FILE__ ) ) . '/wp-load.php';
```
iandunn, line 22: `wp eval-file <filename>` will run a PHP file and load WP for you. Probably the same outcome, but WP-CLI's custom bootstrapper might fix a few hard-to-find bugs in some edge cases, etc. Nothing tangible in this case, just a good habit IMO.

bin/import-test-content.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	return $meta;
+}
+
+function import_rest_to_posts( $rest_url ) {
+	$response = wp_remote_get( $rest_url );
+	if ( is_wp_error( $response ) ) {
+		die( $response->get_error_message() );
+	}
```
iandunn:
```suggestion
	$response = wp_remote_get( $rest_url );
	$status_code = wp_remote_retrieve_response_code( $response );

	if ( is_wp_error( $response ) ) {
		die( $response->get_error_message() );
	} elseif ( 200 !== wp_remote_retrieve_response_code( $response ) ) {
		die( "HTTP Error $status_code \n" );
	}
```

I applied the branch to my sandbox, and the API endpoints are working. I get a `403 Forbidden` error when trying running this locally, though. I think it's just the same problem that we've been seeing in other situations, caused by recent changes to the sandboxes.

iandunn (approved):
I think extending the REST API `view` context is a great idea, and safelisting the specific post types, blocks, etc that we expect adds a good amount of extra safety. I don't see anything in the API endpoints that looks sensitive 👍🏻 

I left some some comments/suggestions, but none are blocking

wp-content/plugins/wporg-learn/inc/export.php
tellyworth: Personal preference. I dislike having public / externally callable functions that are directly recursive. It feels too much like exposing internal implementation details in an interface. I think I'll rename the recursive functions to add a leading underscore, that'll make my intention more clear.

bin/import-test-content.php
tellyworth: Yeah you're probably right. It's not performance that worried me, it's having to explicitly transform fields in the json to differently-named fields in the php array. That's not necessarily a bad thing though since it gives us control over exactly what is and is not inserted.

wp-content/plugins/wporg-learn/inc/export.php
iandunn: 👍🏻

bin/import-test-content.php
iandunn: Ah, I see. Yeah, some clever mapping thing would probably make the code more complicated 👍🏻

jonathanbossenger:
So cool to see this merged, thanks @tellyworth

PR #253, opened 2021-10-19
fix print area not spanning entire width of the workshops page

ashfame:
Based on discussion https://github.com/WordPress/learn/pull/247#pullrequestreview-775372869

Handling a gutenberg class that only shows up on live, limiting width of workshops detail page when printing.


coreymckrill:
Looks like this fixes it 👍 

Thanks @ashfame!

PR #254, opened 2021-10-20
Make previous/next workshop navigation more accessible.

alexstine:
Add specific link text and eliminate arrow icon.


wp-content/themes/pub/wporg-learn-2020/template-parts/component-series-navigation.php
```diff
@@ -35,4 +35,20 @@
 			<?php endif; ?>
 		</li>
 	</ul>
+	<ul class="row video-grid screen-reader-text">
+		<li class="col-6 video-grid_item previous-series-item">
+			<?php if ( $previous ) : ?>
+				<a class="video-grid_item_link" href="<?php echo esc_url( get_the_permalink( $previous ) ); ?>">
+					<?php echo wp_kses_post( 'Previous Workshop: ' . get_the_title( $previous ) ); ?>
```
coreymckrill: @alexstine This template part is now used for both workshops and lesson plans, as they both have a series taxonomy. Does it reduce accessibility to just use "Previous" and "Next" instead of including the post type label?

alexstine:
@coreymckrill It would have been more accessible to leave as is, but this will still be a big improvement over what is there now. I can work on this further later, this will be fine as is.

PR #258, opened 2021-10-23
Adds a print button to Workshops and Lesson plans

### LINKED ISSUE #251: Add a 'print' button to Lesson Plans
Now that we have print stylesheets live for lesson plans and workshops, there's a lot of value in calling attention to this for lesson plans specifically. Adding a "print this page" link to the lesson plan template would be really valuable - it can appear underneath the lesson plan meta data in the sidebar and, if my memory serves correctly, it should just have to trigger `window.print()` since all we need to do is use the default browser printing function.

jonathanbossenger:
Adds a Print View item and button to Workshops, under the Transcript item

post_type-workshop-p-5-preview-true

Adds a Print View link to Lesson plans, below the details but above the Suggestions in the sidebar

post_type-lesson-plan-p-7-preview-true

Closes #251

### DIFF
--- a/wp-content/plugins/wporg-learn/views/block-workshop-details.php
+++ b/wp-content/plugins/wporg-learn/views/block-workshop-details.php
@@ -34,14 +34,14 @@ $has_transcript = false !== strpos( $post->post_content, 'id="transcript"' );
 					</span>
 				</li>
 			<?php endif; ?>
-            <li>
-                <b><?php esc_html_e( 'Print View', 'wporg-learn' ); ?></b>
-                <span>
+			<li>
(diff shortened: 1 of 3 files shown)


jonathanbossenger:
@coreymckrill unless I missed it (and apologies if I did) I don't see a generic 'javascript' file to add general element based actions (ie onclick events) for the print view link/button, so I've just implemented this into on each element for now. If there's somewhere specific you'd like me to move this to, please do let me know.

tellyworth:
Yeah I think onclick is fine here, it's clear enough what's intended, and nice and simple. Given that there's no better place to put it let's not overcomplicate things.

jonathanbossenger:
Thanks @tellyworth 

`let's not overcomplicate things.` my thinking as well 😀

PR #271, opened 2021-11-02
Fix up link text on front page for screen readers.

alexstine:
Make link text more descriptive. Eliminate those icons that just don't read well with screen readers.


hlashbrooke:
This looks good to me - thanks @alexstine! I'll wait for @tellyworth to approve it before merging.

PR #284, opened 2021-11-24
Add courses to homepage

### LINKED ISSUE #283: Add courses to homepage
We now have two _real_ courses live on Learn WordPress (not just the collections of workshop videos we had before): https://learn.wordpress.org/courses/

Now that we have this, and are working on more courses, now is a good time to add the courses to the homepage of Learn!

My idea would be to lay the home page out in this order:

1. Workshop / Lesson plan block
2. Courses
3. Workshops
4. Social learning spaces
5. Get involved
6. Submit an idea

The only new section there is for courses, but it also rearranges things a little by moving the 'get involved' block lower down the page.

The styling of the courses would simply be identical to the current courses on learn.wordpress.org/courses and just who the latest 2 courses with a 'view all' link.

Happy to take other ideas on this layout and order!

hlashbrooke:
Adds courses to homepage and moves 'get involved' block lower down on the page.

Adds new template part component `course-grid`

Closes #283

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/front-page.php
+++ b/wp-content/themes/pub/wporg-learn-2020/front-page.php
@@ -35,7 +35,7 @@ get_header(); ?>
 		<section>
 			<div class="row align-middle between section-heading">
 				<h2 class="h4 section-heading_title"><?php esc_html_e( 'Recent Courses', 'wporg-learn' ); ?></h2>
-				<a class="section-heading_link" href="/courses"><span aria-hidden="true"><?php esc_html_e( 'View All »', 'wporg-learn' ); ?></span><span class="screen-reader-text"><?php esc_html_e( 'View All Courses', 'wporg-learn' ); ?></span></a>
+				<a class="section-heading_link" href="/courses"><span aria-hidden="true"><?php esc_html_e( 'View All Courses', 'wporg-learn' ); ?></span><span class="screen-reader-text"><?php esc_html_e( 'View All Courses', 'wporg-learn' ); ?></span></a>
 			</div>
 
 			<?php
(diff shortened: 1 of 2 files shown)


hlashbrooke:
This PR uses existing templates wherever possible - the new template component was needed purely to wrap the query loop.

wp-content/themes/pub/wporg-learn-2020/front-page.php
```diff
@@ -32,15 +32,30 @@
...
-				<div class="graphic"><span class="dashicons dashicons-welcome-learn-more"></span></div>
-				<div>
-					<h2 class="h3"><?php esc_html_e( 'Get Involved', 'wporg-learn' ); ?></h2>
-					<p><?php esc_html_e( 'Want to get involved in creating the content for Learn WordPress?', 'wporg-learn' ); ?></p>
-					<a href="//learn.wordpress.org/contribute"><?php esc_html_e( 'Learn how to contribute', 'wporg-learn' ); ?></a>
-				</div>
+		<section>
+			<div class="row align-middle between section-heading">
+				<h2 class="h4 section-heading_title"><?php esc_html_e( 'Recent Courses', 'wporg-learn' ); ?></h2>
+				<a class="section-heading_link" href="/courses"><span aria-hidden="true"><?php esc_html_e( 'View All »', 'wporg-learn' ); ?></span><span class="screen-reader-text"><?php esc_html_e( 'View All Courses', 'wporg-learn' ); ?></span></a>
```
tellyworth: Re the double arrow: I think translators prefer that you either use an entity like `&raquo;`, or leave it out entirely. Could this be simplified by just using `View All Courses` for visible and screen reader alike?

hlashbrooke: Yeah, I'll make that change. I had copied this from the workshops block below this, so I'll update that one to also remove the double arrow.

PR #290, opened 2021-11-30
Adding links to workshop & lesson plan meta data

### LINKED ISSUE #192: Workshop & Lesson Plans: Add links to meta info in the sidebar
It'd be useful for users if the meta info on the sidebar of workshops (topic, level, language, and captions) is linked.
image

We already have filters for them (except for levels), so how about linking to the filtered result page?
image
e.g. Send users to https://learn.wordpress.org/workshops/?series=&topic=7&language=&captions= when "Block Editor" topic is clicked.

hlashbrooke:
This commit enhances the function that generates the workshop meta data block to fetch the IDs as well, and then link each meta item to its filtered archive page.

Closes #192

### DIFF
--- a/wp-content/plugins/wporg-learn/views/block-workshop-details.php
+++ b/wp-content/plugins/wporg-learn/views/block-workshop-details.php
@@ -20,26 +20,28 @@ $has_transcript = false !== strpos( $post->post_content, 'id="transcript"' );
 			<?php foreach ( $fields as $key => $field ) : ?>
 				<li>
 					<b><?php echo esc_html( $field['label'] ); ?></b>
-					<?php
-					$display = '';
-					foreach ( $field['value'] as $field_key => $value ) {
-						$url = '';
-						if ( ! empty( $field['param'] ) ) {
-							$url = trailingslashit( site_url() ) . 'workshops/?' . $key . '=' . $field['param'][ $field_key ];
-						}
+					<span>
+						<?php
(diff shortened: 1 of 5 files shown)


hlashbrooke:
I'm happy with this solution - it works well and allows us to add more here in the future if we need to do so. I'm not 100% sure if I did it all correctly, so happy to be shown improvements if needed!

wp-content/plugins/wporg-learn/views/block-workshop-details.php
```diff
@@ -17,10 +17,31 @@
...
+							if ( ! empty( $field['param'] ) ) {
+								$url = trailingslashit( site_url() ) . 'workshops/?' . $key . '=' . $field['param'][ $field_key ];
+							}
+
+							if ( 0 < $i ) {
+								echo ', ';
+							}
+
+							if ( $url ) {
+								echo sprintf( '%1$s' . esc_url( $url ) . '%2$s' . esc_html( $value ) . '%3$s', '<a href="', '">', '</a>' );
```
tellyworth: What's the reason for using sprintf() in this way? (also a similar line added in component-taxonomy-item.php)

hlashbrooke: A naked echo felt...weird to do for some reason, so honestly I don't have a better reason than that.

PR #292, opened 2021-12-16
Add WordPress version to filters & display on Lesson Plans and Workshops

### LINKED ISSUE #289: Add workshop filter for WordPress version
We have the `wporg_wp_version` taxonomy in the dashboard so we can tag content for specific versions of WordPress and it would be great if we could add this to the workshop filters so that people could find content that is version-specific.

If we could have this live before the release of 5.9 then it can be used in the about page and announcement posts and we can link directly to the filtered content.

ryelle:
Fixes #289 — This adds the WP version to Lesson Plans and Workshops, and adds filtering on that value.

Display on lesson plan
Screen Shot 2021-12-16 at 5 04 51 PM

Display on lesson plan list
Screen Shot 2021-12-16 at 5 03 32 PM

Display on workshop
Screen Shot 2021-12-16 at 5 04 20 PM

Filter in lesson plan sidebar
Screen Shot 2021-12-16 at 5 03 54 PM

Filter in workshop header
Screen Shot 2021-12-16 at 5 04 13 PM

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/blocks.php
+++ b/wp-content/plugins/wporg-learn/inc/blocks.php
@@ -75,7 +75,9 @@ function workshop_details_render_callback( $attributes, $content ) {
 	$topic_ids = wp_get_post_terms( $post->ID, 'topic', array( 'fields' => 'ids' ) );
 	$level     = wp_get_post_terms( $post->ID, 'level', array( 'fields' => 'names' ) );
 	$captions  = get_post_meta( $post->ID, 'video_caption_language' );
-	$versions  = wp_get_post_terms( $post->ID, 'wporg_wp_version', array( 'fields' => 'names' ) );
+
+	$version_ids    = wp_get_post_terms( $post->ID, 'wporg_wp_version', array( 'fields' => 'ids' ) );
+	$version_names  = wp_get_post_terms( $post->ID, 'wporg_wp_version', array( 'fields' => 'names' ) );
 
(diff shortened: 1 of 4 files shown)


hlashbrooke (changes requested):
This looks great! It's exactly what I was hoping to see 😄 

There are three areas where it isn't quite working as expected:

1. The version filter doesn't actually apply on the workshops archive (although it does work on the lesson plans archive) - it updates the URL, but it just shows all the workshops.
2. If a lesson plan has no version tagged, then it still shows "WordPress version" on the lesson plan archive, but there's no data - that needs to be hidden on the archive page if there is no version specified.
3. The version numbers on the single workshop page should link to the filtered archive page for that version. The other meta data items do that and it works correctly on lesson plans, so we need to do it for the WP version on single workshops too.

ryelle:
> The version filter doesn't actually apply on the workshops archive

😅 it worked before I added it to the lesson plans. Should be working now.

> If a lesson plan has no version tagged, then it still shows "WordPress version" on the lesson plan archive, but there's no data - that needs to be hidden on the archive page if there is no version specified.

The change I just made also hides it on the single lesson plan page, is that OK? or should it say something else on single pages?

> The version numbers on the single workshop page should link to the filtered archive page for that version

Done ✅

hlashbrooke (approved):

hlashbrooke:
This all works perfectly now - ready to ship!

PR #301, opened 2022-01-11
Accessibility fixes for single Workshops

alexstine:
- Add heading 3 for screen readers to Workshop Details block.
- Make sure complementary doesn't get appended to every widget.

Let's hold off on merging. I can't test these yet due to not having a mirror image of live. Hard to say if it will help or make the problems worse.

### DIFF
--- a/wp-content/plugins/wporg-learn/views/block-workshop-details.php
+++ b/wp-content/plugins/wporg-learn/views/block-workshop-details.php
@@ -15,6 +15,7 @@ $has_transcript = false !== strpos( $post->post_content, 'id="transcript"' );
 ?>
 
 <div class="wp-block-wporg-learn-workshop-details">
+	<h3 class="screen-reader-text">Workshop Details</h3>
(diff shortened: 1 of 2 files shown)


hlashbrooke:
I've tested this locally for me and it works - the display looks the same and I see the screen reader text in there. I'm not sure of any additional ramifications for this, but it looks straightforward enough to merge. A review from @tellyworth would be handy though :)

alexstine:
@hlashbrooke I am concerned with not being able to test the widgets area. I am 90% certain it will work, but would rather not chance it.

ryelle (approved):
This adds the header to the workshop details block, and changes the nested `aside`s in the widgets to be `div`s, which fixes the multiple complementary bug in my testing.

I'm not sure what "worse problems" might be, but I think this would be good to merge and we can always iterate.

alexstine:
Sounds good. Let's try it.

PR #303, opened 2022-01-13
Filters: Strip out empty `wp_version` to fix filtering results with no version

### LINKED ISSUE #302: Applying filtering doesn't work properly for Workshop page
Link: [https://learn.wordpress.org/workshops/](https://learn.wordpress.org/workshops/)

How to reproduce the issue:

- select any option other than 'WordPress Version' in the filter (say language: English)
- Apply filter

It shows no results even though there are a lot of it. 

Apparently if the WP version part is removed from the URL, the search works fine. 
Link: [https://learn.wordpress.org/workshops/?series=&topic=&language=en_US&captions=](https://learn.wordpress.org/workshops/?series=&topic=&language=en_US&captions=)

ryelle:
Fixes #302.

The fix for an issue in #292 caused this value to always be an array, so `array_filter` does not clear it out. This causes the query to try looking for a version that doesn't exist, returning no posts. The fix adds an extra step in the filter-to-query processing, to remove this value if it's actually meant to be empty.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -511,6 +511,10 @@ function wporg_archive_maybe_apply_query_filters( WP_Query &$query ) {
 
 	if ( is_array( $filters ) ) {
 		$filters = array_filter( $filters );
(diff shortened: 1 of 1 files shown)


hlashbrooke (approved):
This works for me!

PR #306, opened 2022-01-18
Accessibility fixes for single Workshops Part 2

alexstine:
- Add better link screen reader text to Transcript/Print View links.
- Make title on list item heading 3.

### DIFF
--- a/wp-content/plugins/wporg-learn/views/block-workshop-details.php
+++ b/wp-content/plugins/wporg-learn/views/block-workshop-details.php
@@ -50,9 +50,8 @@ $has_transcript = false !== strpos( $post->post_content, 'id="transcript"' );
 				<li>
 					<b><?php esc_html_e( 'Transcript', 'wporg-learn' ); ?></b>
 					<span>
-						<a class="components-button is-secondary is-small" href="#transcript">
-							<?php esc_html_e( 'View ', 'wporg-learn' ); ?>
-							<span class="screen-reader-text"><?php esc_html_e( 'Transcript', 'wporg-learn' ); ?></span>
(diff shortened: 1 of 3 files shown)


wp-content/plugins/wporg-learn/views/block-workshop-details.php
```diff
@@ -60,6 +61,7 @@
 				<b><?php esc_html_e( 'Print View', 'wporg-learn' ); ?></b>
 				<span>
 						<a class="components-button is-secondary is-small" href="#" onclick="window.print();">
```
ryelle: Would an `aria-label` on the `a` tag work here? So we could keep the translatable string as one unit ("Print View"), rather than splitting it?

alexstine: Done.

ryelle:
@alexstine The `aria-label` change is good to merge, but can you split out the `h3` change into a separate PR? It's a good idea, but it's causing a visual style difference, and the change would also need to be made in `component-featured-workshop.php`.

alexstine:
@ryelle What about this approach I just committed? Makes the code slightly bigger but I am afraid if I don't do it this way, it might not ever get done. No styling adjustments should be necessary. I can't edit styles as I don't have vision anymore to make those types of changes.

I also completely missed component-featured-workshop.php. I made a couple of tweaks there, let me know your thoughts.

Thanks.

ryelle:
I've merged this in [this commit](https://github.com/WordPress/learn/commit/b1931327587c911be48ee1f720c292ae4db7ba10). I ended up going back to wrapping the title in `h3`, and took care of the styling fixes. It'll get deployed tomorrow.

Thanks for your work!

alexstine:
@ryelle Thanks for handling this!

PR #311, opened 2022-02-15
Add a quick script to sync to production svn.

dd32:
I wanted a single script I could run outside of SVN/Git and have it sync current trunk from Git to the WordPress.org SVNs. That's what this is.

This might be better living in https://github.com/WordPress/wporg-repo-tools technically but this seems like an obvious place for a sync script.

### DIFF
--- a/bin/sync-to-production.sh
+++ b/bin/sync-to-production.sh
@@ -9,14 +9,53 @@ svn co https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-c
 
 cd /tmp/learn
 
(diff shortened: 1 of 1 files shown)


coreymckrill:
Yeah, eventually something like this would be better in `wporg-repo-tools`, but since that isn't integrated with this repo yet, this is probably fine as a prototype.

dd32:
I'm closing this as I've committed a variant of this onto WordPress.org instead.

See `bin/sync/learn.sh` if you have a WordPress.org sandbox.

PR #658, opened 2022-03-03
Fix breadcrumbs not always shown

### LINKED ISSUE #84: Theme: Breadcrumbs not always shown
_Copied from https://meta.trac.wordpress.org/ticket/5373_

https://learn.wordpress.org/lesson-plan/high-level-overview/ is showing breadcrumbs correctly, but clicking on 'Lessons Plans', the breadcrumbs disappear.

Capture

outdoor2kode:
**Ref**: #84 

**Description:**
Adjust crumbs trail based on the type of current page.
Archive pages, Lesson Plan, Workshop, Course post type pages, and search pages are now identified explicitly.

**Screencast of Results:**
https://user-images.githubusercontent.com/18050944/156573078-4ccda801-463b-4478-9416-5b92b0bdf279.mp4

**TODO:**
Test in the remote sandbox - I haven't had access permission to it yet.
You can see that in the screencast starting at time 1:10, there's a weird behavior happening when clicking into `Courses`. That's probably caused by the local env provision or testing data.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/archive-course.php
+++ b/wp-content/themes/pub/wporg-learn-2020/archive-course.php
@@ -16,6 +16,8 @@ $category_description = '';
 get_header(); ?>
 
 <main id="main" class="site-main">
+	<?php get_template_part( 'template-parts/component', 'breadcrumbs' ); ?>
+
 	<section>
 		<div class="row align-middle between section-heading section-heading--with-space">
 			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
--- a/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
+++ b/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
@@ -14,20 +14,30 @@ $crumbs = array(
 	),
 );
 
-// Get information about the post title.
-$cpt_object = get_post_type_object( get_post_type( get_queried_object() ) );
-
-if ( 'lesson-plan' === get_post_type() ) {
+if ( is_post_type_archive() ) {
 	array_push( $crumbs, array(
-		'label' => ucfirst( $cpt_object->labels->name ),
-		'url'   => home_url( $cpt_object->has_archive ),
-	) );
-}
+		'label' => post_type_archive_title( '', false ),
+		'url' => '',
+	));
+} elseif ( is_singular() ) {
+	if ( is_single() ) {
+		$cpt_object = get_post_type_object( get_post_type() );
 
-array_push( $crumbs, array(
-	'label' => get_the_title(),
-	'url'   => '',
-) );
+		array_push($crumbs, array(
+			'label' => $cpt_object->label,
+			'url'   => home_url( $cpt_object->has_archive ),
+		));
+	}
+	array_push( $crumbs, array(
+		'label' => get_the_title(),
+		'url'   => '',
+	) );
+} elseif ( is_search() ) {
+	array_push( $crumbs, array(
+		'label' => get_search_query(),
+		'url' => '',
+	));
+}
 ?>
 
 <div class="clearfix">
--- a/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
+++ b/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
@@ -12,6 +12,8 @@ namespace WordPressdotorg\Theme;
 get_header(); ?>
 
 <main id="main" class="site-main">
+	<?php get_template_part( 'template-parts/component', 'breadcrumbs' ); ?>
+
 	<section>
 		<div class="row align-middle between section-heading section-heading--with-space">
 			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
--- a/wp-content/themes/pub/wporg-learn-2020/archive-wporg_workshop.php
+++ b/wp-content/themes/pub/wporg-learn-2020/archive-wporg_workshop.php
@@ -19,6 +19,8 @@ $is_filtered = $wp_query->get( 'wporg_archive_filters' );
 get_header(); ?>
 
 <main id="main" class="site-main">
+	<?php get_template_part( 'template-parts/component', 'breadcrumbs' ); ?>
+
 	<section>
 		<div class="row align-middle between section-heading section-heading--with-space">
 			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
--- a/wp-content/themes/pub/wporg-learn-2020/search.php
+++ b/wp-content/themes/pub/wporg-learn-2020/search.php
@@ -18,6 +18,8 @@ $search_query = sprintf(
 get_header(); ?>
 
 	<main id="main" class="site-main type-page" role="main">
+		<?php get_template_part( 'template-parts/component', 'breadcrumbs' ); ?>
+
 		<section>
 			<div class="row align-middle between section-heading section-heading--with-space">
 				<h1 class="section-heading_title h2"><?php echo esc_html( $search_query ); ?></h1>
--- a/wp-content/themes/pub/wporg-learn-2020/single-wporg_workshop.php
+++ b/wp-content/themes/pub/wporg-learn-2020/single-wporg_workshop.php
@@ -9,6 +9,8 @@ get_header();
 ?>
 
 	<main id="main" class="site-main page-full-width" role="main">
+		<?php get_template_part( 'template-parts/component', 'breadcrumbs' ); ?>
+
 		<?php
 		while ( have_posts() ) {
 			the_post();


wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
```diff
@@ -14,20 +14,35 @@
 	),
 );
 
-// Get information about the post title.
-$cpt_object = get_post_type_object( get_post_type( get_queried_object() ) );
-
-if ( 'lesson-plan' === get_post_type() ) {
+if ( is_post_type_archive() ) {
```
outdoor2kode, line 17: Check if the current page is an archive page.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
```diff
@@ -14,20 +14,35 @@
...
 	array_push( $crumbs, array(
-		'label' => ucfirst( $cpt_object->labels->name ),
+		'label' => post_type_archive_title( '', false ),
+		'url' => '',
+	));
+} elseif ( is_singular( 'lesson-plan' )
+	|| is_singular( 'wporg_workshop' )
+	|| is_singular( 'course' ) ) {
```
outdoor2kode: Explicitly check every custom post type we have for now.
Not sure if it's a good idea. For me and a newcomer, it's informative and straightforward.
But would need extra work when there are new custom post types added.
Or we could just move the logic of this part into `else` and explicitly identify the `page` post type and then move the logic in the current `else` into it instead.
Open to discussion.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
```diff
@@ -14,20 +14,35 @@
...
+	$cpt_object = get_post_type_object( get_post_type() );
+
+	array_push($crumbs, array(
+		'label' => $cpt_object->label,
```
outdoor2kode: remove `ucfirst` here as it's already the upper case for the first char. and it might cause problems when it comes to a different locale

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
```diff
@@ -14,20 +14,35 @@
...
+if ( is_post_type_archive() ) {
 	array_push( $crumbs, array(
-		'label' => ucfirst( $cpt_object->labels->name ),
+		'label' => post_type_archive_title( '', false ),
+		'url' => '',
+	));
+} elseif ( is_singular( 'lesson-plan' )
+	|| is_singular( 'wporg_workshop' )
+	|| is_singular( 'course' ) ) {
+	$cpt_object = get_post_type_object( get_post_type() );
```
renintw, line 25: Have tried to look for a WP API to retrieve the post type label of the current post instead of getting a post type object. But I couldn't find it, and `has_archive` still needs to be retrieved from cpt_object later, so cpt_object is retained.

outdoor2kode: Have tried to look for a WP API to retrieve the post type label of the current post instead of getting a post type object. But it seems there's no such function and `has_archive` still needs to be retrieved from cpt_object later, so cpt_object is retained.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
```diff
@@ -14,20 +14,35 @@
...
+
+	array_push($crumbs, array(
+		'label' => $cpt_object->label,
 		'url'   => home_url( $cpt_object->has_archive ),
+	));
+	array_push( $crumbs, array(
+		'label' => get_the_title(),
+		'url'   => '',
+	) );
+} elseif ( is_search() ) {
```
outdoor2kode, line 35: Check if the current page is a search page.

hlashbrooke:
This looks great from the screencast! I can't speak for the code itself and will let someone else review that, but the output looks perfect to me.

iandunn:
> weird behavior happening when clicking into Courses. That's probably caused by the local env provision or testing data.

I think you're right, it sounds like the that field doesn't exist. Maybe the table exists but it's empty?

You might be able to fix it by running this in [wp shell](https://developer.wordpress.org/cli/commands/shell/), or temporarily adding it to a file in `mu-plugins/`:

```php
dbDelta( wp_get_db_schema( 'blog' ) );
```

hlashbrooke:
Try adding the courses each to a 'course category' in the dashboard - the layout of the courses page relies on that now and that DB error is likely due to the courses not having a category assigned.

wp-content/themes/pub/wporg-learn-2020/archive-course.php
```diff
@@ -16,6 +16,8 @@
 get_header(); ?>
 
 <main id="main" class="site-main">
+	<?php get_template_part( 'template-parts/component', 'breadcrumbs' ); ?>
```
iandunn, line 19: I like that you made a separate/atomic commit for updating these files, so that the the history is clean 👍🏻

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
iandunn: That's a good question! I think we probably don't want the extra work you mentioned, since there's a chance we won't remember to come back here and update it when we add a new post type. I think the logic will work for all post types that are visible on the front end.

For the `page` case, I think it'd be good to have it documented somehow, otherwise our future selves won't know what specifically is being affected by that condition.

So I'd probably collapse single posts and pages like the following. It reduces the duplicate code, and feels more "self documenting". 

```
if ( is singular ) {
	if ( is single ) {
		get post type object
		push archive link onto array
	}

	push title onto array
}
```

Pages aren't explicitly mentioned, but it's implied because pages are the only thing that match `is_singular()` but not `is_single()`. A comment could explicitly mention them too, though, if you find that helpful.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
iandunn: That seems good to me 👍🏻

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
iandunn: That's a good point about locales, it looks like there may be some problems w/ the multibyte ones 👍🏻

iandunn (approved):
This looks good to me 👍🏻 

I left a one comment, but don't feel strongly about it. Feel free to stick w/ the current approach if you think it's best. 

Feel free to merge once you're happy with it.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
renintw: Thanks @iandunn, the explanation about extra work needed and `page` case makes a good point!

I've revised the logic based on your suggestion, it's indeed clearer and more explicit.

I'm not going to add extra comments here for the moment as it seems to me that it might bring extra confusion by adding some short and incomplete descriptions. I think the explanation in the WP API document would be clear enough to understand.

renintw: Here's the new [commit](https://github.com/WordPress/learn/pull/658/commits/e5dd69c2c7f495a50c9fcd35e59d780db48a8738) and the new [screencast](https://d.pr/v/yPBmOL) of the revised logic - everything still works as expected.

outdoor2kode: Thanks @iandunn, the explanation about the extra work needed and the `page` case makes a good point, thanks for clarifying!

I've revised the logic based on your suggestion, and I'm not going to add extra comments in the code for the moment as it seems to me that it might bring extra confusion by adding some short and incomplete descriptions. I think the explanation in the WP API document would be clear enough to understand.

Here's the new [commit](https://github.com/WordPress/learn/pull/658/commits/e5dd69c2c7f495a50c9fcd35e59d780db48a8738) and the new [screencast](https://d.pr/v/yPBmOL) of the revised logic - everything still works as expected.

iandunn: Looks good!

Does the final `else` condition still do anything, now that pages are covered by the `is_singular()` case?

If not, I'd lean towards removing it, to avoid clutter and confusion. If something we didn't think of does trigger it, then `get_the_title()` might not work anyway.

coreymckrill (approved):
👍 I agree with @iandunn 's [comments](https://github.com/WordPress/learn/pull/658#discussion_r819182883). Otherwise, I think this is good to go.

wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
outdoor2kode: > If something we didn't think of does trigger it, then get_the_title() might not work anyway.

Ah, you're right. Let me remove it. And then I'll merge the PR.
Thanks!

alexstine:
Related: https://github.com/WordPress/learn/pull/660

PR #660, opened 2022-03-09
Fix accessibility of breadcrumbs

alexstine:
- Add dedicated navigation landmark before the main landmark.
- Add translatable "Breadcrumbs" aria-label.


outdoor2kode:
Hi @alexstine, thanks for helping fix the accessibility!

Both navigation landmark and the Breadcrumbs aria-label are working well in my testing.

However, the style seems a bit off after the change. The breadcrumb is sticking to the leftmost edge of the window instead of aligned with the main section below. 
image
I think this is because you've moved the `get_template_part( 'template-parts/component', 'breadcrumbs' );` outside of the main section, making it unable to get the style from the selector `site-main`

Could you try to fix it, thanks 🙂

alexstine:
Hello @renintw 

> However, the style seems a bit off after the change. The breadcrumb is sticking to the leftmost edge of the window instead of aligned with the main section below.

This is something I will need your help fixing. I cannot see so I cannot fix styling.

I also removed the `role="main"` from the `<main>` tags. This used to not be safe but now it should be supported in most browsers with screen readers.

Thanks.

outdoor2kode:
> This is something I will need your help fixing. I cannot see so I cannot fix styling.

No problem 🙂 I'll push the fixed commit later.
 
> I also removed the `role="main"` from the `<main>` tags. This used to not be safe but now it should be supported in most browsers with screen readers.

Ah, thanks for this. I did leave some comments about this originally, but I removed them at the end of the day.  
Since I found in [here](https://www.w3.org/TR/html-aria/#don-t-add-redundant-roles) saying that this kind of redundancy will likely NOT have any unforeseen side effects, other than unnecessarily making the markup more verbose, and incorrectly signaling to other authors that this practice is useful.

Also, both these two sites [site 1](https://caniuse.com/html5semantic) and [site 2](https://www.accessibilityoz.com/2020/02/html5-sectioning-elements-and-screen-readers/#:~:text=%E2%80%9Carticle%2C%20landmark%E2%80%9D-,Conclusion,-With%20only%20four) suggest that HTML5 sectioning elements are still not fully supported by all screen readers or web browsers, it is still much safer to use Aria roles even in case of simple pages. And if HTML5 sectioning elements are being used, the page should be manually tested with various screen readers to make sure the tags are correctly interpreted.

That being said, just like what you mentioned that the HTML5 sectioning elements should already be supported in most browsers and screen readers, and the conclusion in those two sites does show that only a small number of browsers and screen readers will be affected. So I'm ok with the changes here.

Thanks 🙂

alexstine:
@renintw Yeah, a lot has changed with the great world of HTML5. Roles on header, navigation, main, aside, and footer are just getting in the way these days. Latest versions of browsers and NVDA, JAWS, and Voiceover support it fine.

This TRAC ticket has more info if you are interested.

https://core.trac.wordpress.org/ticket/54079

outdoor2kode:
@alexstine Thanks for sharing it, it's helpful!

PR #670, opened 2022-03-23
Term Translations: Include the URL to where the translation is used

dd32:
Currently Term names & descriptions specify that it's a term description, and the kind of string it is, but they don't link to where the translation is actually used

For example:
Screen Shot 2022-03-23 at 5 42 38 pm

It seems reasonable to include the URL to the term in a comment here, so that translators can backtrack to where it's actually used.

### DIFF
--- a/bin/i18n.php
+++ b/bin/i18n.php
@@ -115,8 +115,11 @@ function main() {
 		$label = addcslashes( $tax_label, "'" );
 
 		foreach ( $terms as $term ) {
+
 			$name = addcslashes( $term['name'], "'" );
-			$file_content .= "_x( '{$name}', '$label term name', 'wporg-learn' );\n";
+			$link = addcslashes( $term['link'], '*' );
+
+			$file_content .= "/* translators: {$link} */\n_x( '{$name}', '$label term name', 'wporg-learn' );\n";
 
 			if ( 'cli' === php_sapi_name() ) {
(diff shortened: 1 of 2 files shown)


dd32:
This does end up having some URLs that aren't valid, for example:
`https://learn.wordpress.org/wporg_wp_version/5-6`

But is helpful for other terms such as this:
Screen Shot 2022-03-23 at 6 02 07 pm

naokomc:
Including as much context as possible is always helpful.
Having non-valid URLs in some parts can get confusing, but if that's the by-product that has to happen, I still prefer having additional context personally.

jeroenrotty:
I agree with Nao's point - I prefer to have as much context and a URL where it's actually used to see what the context is on that page/section/tutorial/...

NekoJonez:
Can't agree more with @naokomc & @jeroenrotty! Context is king in translation.

psmits1567:
I also agree with @naokomc and @jeroenrotty 
The more info the better the translation

outdoor2kode:
Thanks @dd32. The changes look good to me. I think the link to where the translation is used does give more context and is helpful.

outdoor2kode (approved):

outdoor2kode:
Added URLs to the new terms or descriptions. 
Forced push to rebase the trunk.

PR #745, opened 2022-04-18
Course titles: Render as heading 3 for better heading order in archives

alexstine:
Under the Recent Courses (heading 2), change those titles to heading 3 for better heading order.

https://learn.wordpress.org


alexstine:
@hlashbrooke @tellyworth Hopefully a quick one for review/send live. Looks like it has h4 class so hoping this doesn't break styling.

hlashbrooke:
Looks good - thanks @alexstine!

PR #767, opened 2022-05-29
Create content-error-report issue template

Piyopiyo-Kitsune:
This is an issue template for adding Learn WordPress Content Error Reports from Help Scout to GitHub.

### DIFF
--- a/.github/ISSUE_TEMPLATE/content-error-report
+++ b/.github/ISSUE_TEMPLATE/content-error-report
@@ -4,13 +4,24 @@ about: Learn WordPress Content Error Report template
 title: Learn WordPress Content Error Report: Ticket #
 labels: Help Scout, Error Report
 assignees: ''
-
 ---
 
+<!--
+
+Thank you for taking the time to report a content error. Please include the title and the URL to the content in question in the Error Description, and a description of the error.
+
+This can be done either using the markdown format, for example, (Creating Custom Post Types Without Code)[https://learn.wordpress.org/workshop/creating-custom-post-types-without-code/], or by use the **Add a link** icon in the issue editor toolbar. 
+
+Alternatively you can share the link to the page as is.
+-->
+
 # Error Description
-Write a description of the topic here.
+- Content URL: 
+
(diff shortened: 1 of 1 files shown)


azhiya:
Hey @courtneyr-dev I'm assuming we can just squash and merge this.

jonathanbossenger:
@azhiya I've opened PR #772  to fix the errors with the PHPCS linter, which was causing the check failure. Once those checks pass, I'll merge the fix.

jonathanbossenger:
@Piyopiyo-Kitsune nice work. 

Could I make one suggestion, can we update this to ask the reporter to add a link to the piece of content being reported on. This will make it a lot easier for folks working on content fixes to quickly click the relevant link to see the content with the error. 

We can use HTML opening and closing tags to also include some instructions for bug reporters, which won't display in the final bug report. So for example we could include something like this in the Error Description section:

```
<!--
Please include the title and the URL to the content in question in the Error Description.

This can be done either using the markdown format, for example, (Creating Custom Post Types Without Code)[https://learn.wordpress.org/workshop/creating-custom-post-types-without-code/].

Alternatively, use the **Add a link** icon in the issue editor toolbar
-->
```

jonathanbossenger:
@azhiya @courtney-pk the [PR to fix the failing tests](https://github.com/WordPress/Learn/issues/772) has been merged, so it will be safe to merge this PR as well.

azhiya:
@jonathanbossenger sorry I'm asking again, don't want to do make a mistake but can I go ahead and merge this.

jonathanbossenger:
@azhiya sorry for the delay here (I thought I had replied). Yes, it is safe to merge. Unfortunately, because the bug was only fixed after this PR was created, it still uses the older code when running the checks. Squash and merge away! 😊

azhiya:
All done @Piyopiyo-Kitsune and @jonathanbossenger. That was oddly satisfying :-)

jonathanbossenger:
"Feel the power!" 😅

PR #790, opened 2022-06-21
Accessibility fixes for social learning spaces on front page

alexstine:
* Add heading 3 structure.
* Move timestamp after initial title.
* Improve link text.


hlashbrooke:
Thank you @alexstine, as always :)

Issue #810, opened 2022-07-08
Bug report: learn.wordpress filter dropdown faulty 


westnz:
## Description
The filter dropdown on the WordPress Learn website is not working properly. When you click on a filter it closes. 

## Step-by-step reproduction instructions
Please list the steps needed to reproduce the bug. For example:
1. Go to https://learn.wordpress.org/workshops/
2. Click on one of the filters

## Expected behaviour

I expected the dropdown filter to stay open.

## Actual behaviour

The filter dropdown closes when you click on it.

## Screenshots or screen recording (optional)

https://user-images.githubusercontent.com/89751452/177898053-e55c7d5e-809c-4dab-b0e9-377ee208bae2.mov

## Environment information
- I am using Chrome
- I am using a Macbook Pro

adamwoodnz:
I can't reproduce this in my local environment, but I'll try my sandbox now. Production has many more plugins activated than local though. @westnz do you know if any of them have been added/activated recently?

westnz:
Thanks @adamwoodnz.  We don’t have any new plugins on there that I’m aware of.

westnz:
I could reproduce it when I tried it in Safari as well, and without being logged in.

vagelisp:
I can reproduce this on both Chrome and Safari on a Macbook Pro.
As you can see in the video, you can keep it open if you click on the right side of the arrow, but if you click anywhere else, the selector picks the first available option. Clear filter [x] also doesn't work.

https://user-images.githubusercontent.com/9304640/178074109-c87f6aa5-2a18-45aa-983d-7a96a77136fb.mov

adamwoodnz:
Yep definitely not arguing about reproducibility! Broken in all browsers in prod for me, and also in my sandbox. I'm continuing to debug this week.

kaitohm:
I've confirmed this has now been fixed in Chrome, Firefox and Safari. Closing.

PR #814, opened 2022-07-12
Fix workshop filters

adamwoodnz:
See #810 

As this is a critical bug it is desirable to get basic functionality back while we look at a long term fix.

This PR temporarily disables the enhancement of the Workshop filter menus, leaving them as native select elements. The search and clear functionality is removed but they continue to provide filtering as expected, with mouse or keyboard:

Learn filters


wp-content/themes/pub/wporg-learn-2020/functions.php
```diff
@@ -58,16 +58,20 @@ function wporg_learn_scripts() {
 		true
```
adamwoodnz, line 58: This parameter being `true` tells WordPress to load the script in the footer. Without `wp_footer` being called this never happens. The `wporg-filters` script below is enqueued the same way and never loads.

https://developer.wordpress.org/reference/functions/wp_enqueue_script/

hlashbrooke:
This looks good to me and works on my local. Will need someone to test in a sandbox before deploying.

adamwoodnz:
> This looks good to me and works on my local. Will need someone to test in a sandbox before deploying.

👍 I can do that

Issue #734, opened 2022-04-11
Uneven "Join this space" button sizes in mobile device


bhaveshd36:
The "Join this space" button does not look good in mobile devices due to the different button size based on the discussion event content. Should we have a constant "Join this space" button size?
Page Link: [https://learn.wordpress.org/](https://learn.wordpress.org/)
Screenshot: [https://prnt.sc/nZddYmYeGoG0](https://prnt.sc/nZddYmYeGoG0)

PR #816, opened 2022-07-13
Solve button alignment issue

varshildoshi151189:
Solution/Patch for the Button alignment issue.
#734 

CleanShot 2022-07-13 at 10 35 56


hlashbrooke:
Thanks for the PR @varshildoshi151189!

Could you please check the errors there - looks like they're to do with linting, but I haven't dived into them yet.

@adamwoodnz Can you test this on a sandbox? The events on this page are pulled from the .org database so I don't see them on my local and I don't have a sandbox to test on.

adamwoodnz:
Nice work @varshildoshi151189 !

I've found that the layout can be a little consistent using `flex-wrap: wrap` with content of mixed lengths, where some buttons have dropped below and others remain above:

Screen Shot 2022-07-18 at 4 38 54 PM

I've added a commit using `flex-direction: column` to fix this:

Screen Shot 2022-07-18 at 4 43 40 PM

Screen Shot 2022-07-18 at 4 43 58 PM

I should be able to merge this tomorrow, let me know if you see any issues.

wp-content/themes/pub/wporg-learn-2020/css/components/_workshop-discussion-events.scss
```diff
@@ -1,10 +1,6 @@
 .wporg-learn-workshop-discussion-events {
 	:last-child {
 		margin-bottom: 0;
-
-		@media only screen and (max-width: $breakpoint-tablet) {
```
adamwoodnz: This was actually targeting every `:last-child` in `.wporg-learn-workshop-discussion-events`, not only the buttons as I think was the intention.

wp-content/themes/pub/wporg-learn-2020/css/components/_workshop-discussion-events.scss
```diff
@@ -44,5 +41,10 @@
 	.wp-block-button__link:visited {
 		margin: 0;
 		padding: 0.5rem;
+		white-space: nowrap;
```
adamwoodnz, line 45: Keep the buttons a consistent width, so that they don't get squashed due to the flex layout

Issue #889, opened 2022-08-17
[Bug report] UI issue in submit button and pagination in the Course Category page


tahmidulkarim:
## Description
The submit button and spacing around the text and pagination in not correct on the course category pages.

## Step-by-step reproduction instructions

1. Go to https://learn.wordpress.org/course-category/contributing-to-wordpress/
2. Scroll down to the bottom
3. Just above the footer you can see the section - _Have an Idea for a Tutorial? Let us know!_ which has styling issue.

## Expected behaviour
There should be proper padding around the text, pagination and the button. Also, the padding inside the button should be correctly set.

It displays correctly on this page - https://learn.wordpress.org/tutorials/

Here is a screenshot of the correct layout - 
<img width="1439" alt="Screenshot at Aug 17 19-54-23" src="https://user-images.githubusercontent.com/15308261/185152157-b46f0b14-0436-4388-b8ce-3ec6516e1f93.png">

## Actual behaviour
There is no padding around the text and the button.

The course category is created from the Sensei LMS plugin so it overwriting the default styles set in the wporg-learn-2020 theme. The category archive page loads some extra styles from Sensei LMS which is creating this issue. The buttons style issue has to be solved from the Sensei LMS plugin but the padding around the text and pagination should be solved from the theme.

## Screenshots 
<img width="1439" alt="learn_wporg" src="https://user-images.githubusercontent.com/15308261/185154120-58a42315-14eb-47ab-b9d5-e27b0252ec83.png">

## Environment information
- Chrome - version 102
- Device - MacBook Air M1, 2020 (MacOs - 11.5.2)

kaitohm:
Update: The PR above has corrected the padding around the text. But the padding within the button remains to be fixed.

jonathanbossenger:
There have been a few Sensei updates since this last comment, so this issue needs to be retested to verify if it's still a valid bug.

BhargavBhandari90:
I think this bug is not a bug anymore. We don't have **_Have an Idea for a Tutorial?_** section anymore. Also there is no pagination now.

<img width="1461" alt="Image" src="https://github.com/user-attachments/assets/84834269-40b2-454e-8801-417a7d149dc4" />

kaitohm:
Thanks for checking, @BhargavBhandari90 🙌  I'll close this as fixed.

PR #890, opened 2022-08-17
Fixed Spacing issue in course category page

tahmidulkarim:
See #889

The PR adds a **:not** pseudo class in the existing style to check if the page is the course category page (.tax-course-category) class, and in that case the padding 0 won't be applied, that is it will load the default padding. 

Here is a screenshot of the fix -

fixed_spacing


hlashbrooke:
Thanks for the contribution @tahmid-ul! This looks great!

Fixes #889

PR #921, opened 2022-09-06
Add Lesson Plan Details and Actions blocks

### LINKED ISSUE #274: Convert lesson plan details into block widget
The lesson plans have a static widget that shows lesson plan details.

Pull request #239 converts the sidebars into dynamic widget areas, but that widget is still hardcoded static content. This needs to be converted to a block widget. V1 would be fine to have the widget with no options so we can at least move it around, but v2 could have some options for what to show.

adamwoodnz:
See #274 

This ports the Lesson Plan details and actions into 2 block widgets, so that they can be moved in the sidebar by an admin via the widget editor. The structure of the code is copied from the [Workshop Details block](https://github.com/WordPress/Learn/tree/trunk/wp-content/plugins/wporg-learn/js/workshop-details).

There are minimal changes on the frontend; buttons styles have been reused for the Actions block and HTML semantics and accessibility have been improved in the Details block. See screenshots below.

Some validation has been added to the existing Workshop Details block widget to stop the Widgets editor crashing if it is added to the sidebar for an incompatible post type.

### Screenshots

| Widget Editor | Validation |
|-|-|
| Screen Shot 2022-09-20 at 5 54 00 PM | Screen Shot 2022-09-20 at 5 52 59 PM |

| Desktop before | Desktop after |
|-|-|
| Screen Shot 2022-09-19 at 11 53 48 AM | Screen Shot 2022-09-19 at 11 54 02 AM |

| Mobile | Print |
|-|-|
| Screen Shot 2022-09-19 at 11 40 43 AM | Screen Shot 2022-09-19 at 11 41 27 AM |

### How to test this PR
1. Visit http://localhost:8888/wp-admin/widgets.php and expand the Lesson Plan widget.
2. There should be 2 new blocks available; Lesson Plan Details and Lesson Plan Actions. Add them and test reordering the sidebar, updating and checking the frontend display, comparing to prod.
3. Test posts with different media; slides linked and slides downloadable.
4. Test with a small screen < 768px; the details should be shown but no actions.
5. Test the print button; only the details should be shown and no other sidebar content.
6. Test adding the blocks to the sidebars for other post types. They should not be displayed on the frontend and the Widget editor should not crash on reload.


wp-content/themes/pub/wporg-learn-2020/template-parts/component-taxonomy-item.php
```diff
@@ -1,34 +0,0 @@
-<?php
-/**
- * Template part for displaying lesson/workshop details on the content-single part.
```
adamwoodnz, line 3: This was only actually used for Lesson Plan details

wp-content/plugins/wporg-learn/inc/blocks.php
```diff
@@ -24,10 +24,63 @@
...
+/**
+ * Render the block content (html) on the frontend of the site.
+ *
+ * @param array  $attributes
+ * @param string $content
+ * @return string HTML output used by the block
+ */
+function lesson_plan_details_render_callback( $attributes, $content ) {
+	$post    = get_post();
+	$details = wporg_learn_get_lesson_plan_taxonomy_data( get_the_ID(), 'single' );
```
adamwoodnz, line 137: This function exists in [functions.php](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2020/functions.php#L230) and it doesn't feel quite right to me for the plugin to depend on it. The function is also called by [the code which renders the details in Lesson Plan archive cards](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php#L34) so it can't be moved. Should it be duplicated in the plugin, or is there prior art for explicit code sharing between the theme and a plugin?

.stylelintrc
```diff
@@ -1,5 +1,5 @@
 {
-	"extends" : "stylelint-config-wordpress/scss",
```
adamwoodnz, line 2: This is deprecated, see https://www.npmjs.com/package/stylelint-config-wordpress

Additionally switching to this new package allows the VSCode stylelint extension to pick up failures, allowing early detection.

wp-content/plugins/wporg-learn/inc/blocks.php
```diff
@@ -71,6 +186,10 @@ function register_workshop_details() {
  * @return string HTML output used by the block
  */
 function workshop_details_render_callback( $attributes, $content ) {
+	if ( get_post_type() !== 'wporg_workshop' ) {
```
adamwoodnz, line 189: The lack of validation here recently led to an incident where the Widgets editor crashed due to this block mistakenly being added to the Lesson Plan sidebar.

wp-content/plugins/wporg-learn/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+	extends: 'plugin:@wordpress/eslint-plugin/recommended',
```
adamwoodnz, line 2: Allows VSCode to highlight problems. Linting remains the same.

package.json
```diff
@@ -24,7 +24,11 @@
 		"create": "composer install && yarn workspaces run build && yarn run install:env",
 		"wp-env": "wp-env",
 		"lint:php": "composer run lint",
-		"format:php": "composer run format"
+		"format:php": "composer run format",
+		"lint:css": "yarn workspace wporg-learn-plugin lint:css && yarn workspace wporg-learn-theme lint:css",
+		"lint:js": "yarn workspace wporg-learn-plugin lint:js",
+		"watch:theme": "yarn workspace wporg-learn-theme start",
+		"watch:plugin": "yarn workspace wporg-learn-plugin start"
```
adamwoodnz, line 31: Shortcuts for the scripts listed https://github.com/WordPress/Learn/blob/trunk/readme.md#linting and https://github.com/WordPress/Learn/blob/trunk/readme.md#development

hlashbrooke:
I've tested this all locally, and it works as expected.

This is great @adamwoodnz - it came together really well and looks perfect! I'm happy to merge and deploy this.

wp-content/plugins/wporg-learn/inc/blocks.php
adamwoodnz: @tellyworth @ryelle I'd like your input on this please 🙏

wp-content/plugins/wporg-learn/js/hooks.js
```diff
@@ -0,0 +1,27 @@
...
+	} );
+
+	const widgetId = attributes.__internalWidgetId;
+
+	// __internalWidgetId is not stable.
+	// If it is undefined we can't look up the widget sidebar type to check compatibility
+	// so we have to return true and let the backend validation handle the incompatible type.
+	if ( widgetId === undefined ) {
+		return true;
+	}
```
adamwoodnz: @ryelle any opinion or suggestions for improving this? I'm guarding against it being removed but I'm not sure whether I should be referencing it at all and I haven't found any other way of finding the current block in the list of widget blocks so far...

wp-content/plugins/wporg-learn/inc/blocks.php
ryelle: I think we tend to go the other way — functions defined in plugins can be used in themes. That feels more natural to me, since a theme is more likely to be changed (in a redesign), but the data structure would stay the same.

I wouldn't consider this a blocker right now, though, since you know the theme shouldn't change.

wp-content/plugins/wporg-learn/js/hooks.js
ryelle: I don't think this is working correctly, I can add the block from any widget area & while I saw the warning once, it doesn't seem consistent.

I pushed up a commit that changes this to check the block ancestry, since technically each widget area is a `core/widget-area` block, and the `id` attribute of that block is the sidebar name. I also changed the hook name to `useIsBlockInSidebar`, to reflect better what it's doing. This seems to work for widgets and the regular post editor, even if the block is nested in other blocks (group, etc).

If I misunderstood the purpose of this function, feel free to drop my commit :)

adamwoodnz: Code wise this looks great thanks, lots for me to learn about these APIs!

I'll test it out...

adamwoodnz: > This seems to work for widgets and the regular post editor,

I assumed incorrectly that a block with `category: 'widgets'` would only be allowed in the widget editor, it's a shame that's not the case. Are you aware of a programmatic way of blocking the post editor, or is my approach of using a message all that should be expected?

ryelle: No, categories are just categories — you can see the "widgets" category in the post editor too. Looks like [the `ancestor` property](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-metadata/#ancestor) in `block.json` could work, if you add `ancestor: [ "core/widget-area" ]` it _should_ restrict it to the widget editor.

adamwoodnz: Decided there's really no reason why this info couldn't be displayed in a post, so it won't be blocked, but I've just added a new commit to show different messages depending on where it's used.

adamwoodnz: | Lesson Plan Post | Workshop Post | Lesson Plan Sidebar | Workshop Sidebar |
|-|-|-|-|
| Screen Shot 2022-09-28 at 5 30 12 PM | Screen Shot 2022-09-28 at 5 30 28 PM | Screen Shot 2022-09-28 at 5 30 48 PM | Screen Shot 2022-09-28 at 5 31 04 PM |

ryelle (approved):
Looks good & works as described 👍🏻

Issue #811, opened 2022-07-11
Add language meta field for lesson plans


hlashbrooke:
Workshops have an autocomplete meta field for the workshop language, but lesson plans do not. We need to add this same meta field to lesson plans and then filter the lesson plan archive page based on the language the site is currently being viewed in (which is how workshops currently work).

hlashbrooke:
Once we have this in place for courses, online workshops and lesson plans, then I'd like to integrate this meta field with the existing locale switcher so that content in the selected locale is given priority in all queries.

piermario:
Hello @hlashbrooke! Have you considered adding a language filter in the lesson plan result page sidebar as well? Would that be harder than implementing the language priority function linked to the existing locale switcher? Thanks for your consideration. :)

hlashbrooke:
Yeah - we have that for Tutorials and it makes sense to have it on Lesson Plans - I actually thought we did have it, but just checked now and I see we don't.

piermario:
Thanks for checking! A boy can dream, then. 😇

adamwoodnz:
Done in #1030

PR #931, opened 2022-09-11
Add a language taxonomy to Learn

jrastaban:
See #811 

The training team needs a way to support courses (course) and online workshops (meeting) in multiple languages. Adding a language taxonomy will allow users to more easily find resources in their preferred language.

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/taxonomy.php
+++ b/wp-content/plugins/wporg-learn/inc/taxonomy.php
@@ -18,6 +18,7 @@ function register() {
 	register_lesson_duration();
 	register_lesson_group();
 	register_lesson_instruction_type();
+	register_lesson_language();
 	register_lesson_level();
 	register_lesson_plan_series();
 	register_workshop_series();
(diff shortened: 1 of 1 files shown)


hlashbrooke:
This is a great idea to have! There is, however, already a "Language" meta field for tutorials (`wporg_workshop`), which autocompletes from a standardised list of available locales. I feel like it would be better to apply that meta field to the other post types as well instead of introducing a new taxonomy.

The downside of a meta field is there's no immediate archive available like with a taxonomy, but that's easy enough to create with a quick custom filter.

jrastaban:
It looks like `wporg_workshop` is using two fields, `video_language` and `video_caption_language`. A more generic language field would make more sense for the other learning types. Would you like to see a new meta field that mirrors it's functionality but is used across all 3 of these types? Maybe a `learning_language` meta field?

hlashbrooke:
That makes sense to me - the caption language won't need to apply to other post types, but a more generic `learning_language` sounds like a good way to go. I'm sure it could copy the same functionality as the existing language selector for tutorials, so hopefully a lot of that code could be reused or referenced.

jrastaban:
Makes sense, let me see what that would take to implement.

burtrw:
This is a bit over my head, but if it was the separate taxonomy unique to Courses (like the current WP Versions taxonomy), I believe this will play nice with the Course Grid block and filters - so it will be easy to display or filter courses by language with out of the box tools. 

But maybe custom course listings for the learn site isn't too hard to do, so the meta field may also make more sense.

jrastaban:
From a discussion with the training team here at WCUS, the desire is a way to start segmenting videos by language. While lesson plans have a similar interface with the workshops, the other two post types do not and so we don't know how this will function on the front end yet.

Is a PR that just adds a meta field to the backend of post types something that can be accepted? It wouldn't close out the issue above until the front end was added but would allow the team to begin organizing while a front end decision is finalized..

hlashbrooke:
Yes - to my mind, an initial PR to add the meta field in the dashboard will be a great start. Then we can start using it before adding in frontend filters.

adamwoodnz:
Language meta field added to all post types in #1030

Issue #925, opened 2022-09-08
Add Code Syntax Block upgrades to Learn


hlashbrooke:
DevHub recently received upgrades to the Code Syntax Block that make code blocks much more useful and easier to use. We'd like to get these improvements added to Learn as well.

Here's an example of the upgrades in action: https://developer.wordpress.org/block-editor/reference-guides/data/data-core-blocks/

And this is where I'm told the upgrades were implemented: https://github.com/WordPress/wporg-developer/blob/34e370515f48dab03a0c4fd444bff1c4e151f52c/source/wp-content/themes/wporg-developer/js/function-reference.js

adamwoodnz:
Reopening to fix some lint issues

PR #934, opened 2022-09-12
Fix lint failures from function reference addition

adamwoodnz:
See #925 

Some of the code added in https://github.com/WordPress/Learn/commit/3a509e356b3a661157a5f8a7d8b1cf7465c03468 introduced lint failures which are now blocking new PRs.


phpcs.xml.dist
```diff
@@ -146,7 +147,7 @@
 	     provided as a comma-delimited list. -->
 	<rule ref="WordPress.WP.I18n">
 		<properties>
-			<property name="text_domain" type="array" value="wporg-learn,wporg_learn" />
+			<property name="text_domain" type="array" value="wporg-learn,wporg_learn,wporg" />
```
adamwoodnz: @iandunn alternatively we could change the references in `functions.php` to `wporg-learn`, I wasn't 100% sure what is correct.

tellyworth: Looks like it should be `wporg-learn` for the Learn theme and plugin. `wporg` for mu-plugins. There are a few instances of `wporg` textdomain in the theme's `functions.php`, those are probably mistakes.

It's interesting that the code uses a mishmash of `wporg-learn` and `wporg_learn`. I guess it works but that seems like it must be a mistake? This says textdomains must use dashes not underscores: https://codex.wordpress.org/I18n_for_WordPress_Developers#Text_Domains. I wonder if there's a reason for it.

I'd be tempted to remove `wporg` from the property here and fix those lint errors (there are only a handful). And leave the dash/hyphen thing for a later PR.

tellyworth (approved):
Works fine for me. The textdomain thing isn't a blocker, that could be a followup PR if you want to get this in.

phpcs.xml.dist
outdoor2kode: > Looks like it should be wporg-learn for the Learn theme and plugin. wporg for mu-plugins. There are a few instances of wporg textdomain in the theme's functions.php, those are probably mistakes.

👍 From the search result, `'wporg'` all appears in the files under `mu-plugins/`, I believe the `'wporg'` in `themes/pub/wporg-learn-2020/functions.php` can be changed to `'wporg-learn'` safely.

> It's interesting that the code uses a mishmash of wporg-learn and wporg_learn. I guess it works but that seems like it must be a mistake? This says textdomains must use dashes not underscores: https://codex.wordpress.org/I18n_for_WordPress_Developers#Text_Domains. I wonder if there's a reason for it.

I can find both of them work on https://translate.wordpress.org/locale/zh-tw/default/meta/learn-wordpress/, but not sure why it could work. Since I thought only the text domain specified in the theme [style.css](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2020/style.css#L10) would take effect, perhaps this is not how it really works under the hood.

outdoor2kode (approved):

phpcs.xml.dist
tellyworth: I'm assuming something (in core? GlotPress?) magically replaces the underscore with a dash, so it appears to work ok. We should follow the recommendations though.

PR #935, opened 2022-09-12
Improved copy button styling

jonathanbossenger:
Improves the new Copy button styling for the Code block

Before:

before

After:

after

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/css/objects/_prism.scss
+++ b/wp-content/themes/pub/wporg-learn-2020/css/objects/_prism.scss
@@ -217,7 +217,7 @@ pre.line-numbers > code {
 .wp-code-block-button-container {
 	display: flex;
 	justify-content: right;
-	padding: 0.5rem;
+	padding: 1rem;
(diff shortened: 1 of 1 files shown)


wp-content/themes/pub/wporg-learn-2020/css/objects/_prism.scss
```diff
@@ -217,7 +217,7 @@ pre.line-numbers > code {
 .wp-code-block-button-container {
 	display: flex;
 	justify-content: right;
-	padding: 1rem;
+	padding: 0.5rem;
```
iandunn: This all looks good, but I think it'd be ideal to leave the original values, and put the changes at the buttom in the "Learn customizations" section. That'll make it easier to sync w/ the original in the future, if we ever want to.

jonathanbossenger: Good point, thanks @iandunn moved the Learn specific additions to the bottom of the file.

iandunn:
This is deployed now

PR #961, opened 2022-09-21
Fix/944 lesson plan landing page a11y

### LINKED ISSUE #944: Accessibility improvements on Lesson Plan landing page
https://learn.wordpress.org/lesson-plans/ could benefit from the following a11y improvements:

1. The dashicons associated with topics/audiences, etc., should use aria-hidden="true" to prevent the possibility they could be read as characters.
2. Focus states are kind of minimal; the single-pixel dotted outline offers poor visibility. (They also end up looking accidental, rather than designed.)
3. The duration filters would benefit from human-readable labels, e.g. '15 minutes'.  Also, "any" and "over 60m" are not the same thing, so that label feels ambiguous.
4. "Search lesson plans" has a label of "Search for". This results in reading that's "Search landmark. Search for: Search lesson plans". It's very verbose. Would be better if there was just a visible label with "Search lesson plans" and no placeholder.
5. Hit area for links is small and should be enlarged to cover entire tile:

Screen Shot 2022-09-16 at 7 19 56 AM

See https://wordpress.slack.com/archives/C02RP4X03/p1663279177628089

adamwoodnz:
Fixes #944 

### Tasks

- [x] The dashicons associated with topics/audiences, etc., should use aria-hidden="true" to prevent the possibility they could be read as characters.
- [x] Focus states are kind of minimal; the single-pixel dotted outline offers poor visibility. (They also end up looking accidental, rather than designed.)
- [x] The duration filters would benefit from human-readable labels, e.g. '15 minutes'.  Also, "any" and "over 60m" are not the same thing, so that label feels ambiguous.
- [x] "Search lesson plans" has a label of "Search for". This results in reading that's "Search landmark. Search for: Search lesson plans". It's very verbose. Would be better if there was just a visible label with "Search lesson plans" and no placeholder.
- [x] Hit area for links is small and should be enlarged to cover entire tile.

### Details and screenshots

All the interactive button-like elements across the page have have large padding for increased hit area, and consistent focus styles inherited from `button`:

Screen Shot 2022-09-27 at 3 07 18 PM

The search field label has been made visible and the placeholder removed:

| Lesson Plans | Courses | Tutorials |
|-|-|-|
| Screen Shot 2022-09-27 at 3 11 15 PM | Screen Shot 2022-09-27 at 3 11 04 PM | Screen Shot 2022-09-27 at 3 11 30 PM |

The card grid used for Courses and Lesson Plans has been migrated to use CSS grid which is a vastly simpler implementation. It did require some markup changes to move the title and descriptions out of the card grid, but semantically it makes more sense for only `card` elements to be within a `card-grid`.

This fixes issues with the current Courses grid at certain screen widths, and introduces more breakpoints for the 4 column grid, so it now changes from 4 to 2 to 1.

| Desktop | iPad | Mobile |
|-|-|-|
| screencapture-localhost-8888-lesson-plans-2022-09-27-15_30_05 | screencapture-localhost-8888-lesson-plans-2022-09-27-15_41_42 | screencapture-localhost-8888-lesson-plans-2022-09-27-15_43_11 |


wp-content/themes/pub/wporg-learn-2020/css/components/_card.scss
```diff
@@ -105,41 +106,83 @@
...
-			}
+
+			@include card-grid_2();
 		}
+	}
+
+	@media only screen and (max-width: $breakpoint-mobile) {
+		&_4 {
 
-		&.gap-64 {
```
adamwoodnz, line 140: Unused class

wp-content/themes/pub/wporg-learn-2020/css/components/_course-archive.scss
```diff
@@ -1,4 +0,0 @@
-h2.course-category-header,
-div.course-category-description {
-	flex-basis: 100%;
```
adamwoodnz, line 3: These have now moved out of the card grid and are just standard block elements

wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
```diff
@@ -45,25 +45,23 @@
 			'orderby'    => 'id',
 		) );
 		?>
-		<div class="row lp-taxonomy categories">
+		<div class="lp-taxonomy">
 			<div class="card-grid card-grid_4">
-				<h2 class="h4 lp-taxonomy-header"><?php echo esc_html__( 'Topic', 'wporg-learn' ); ?></h2>
-				<div class="lp-taxonomy-description"><?php echo esc_html__( 'Browse lesson plans by their high-level topic.', 'wporg-learn' ); ?></div>
+				<header class="card-grid-header">
```
adamwoodnz, line 50: These non `card` elements within the `card-grid` cause responsive issues. Wrapping them with this div means we can control their behaviour and force them to always be full width, no matter what width the cards are set to

.stylelintrc
```diff
@@ -1,5 +1,5 @@
 {
-	"extends" : "stylelint-config-wordpress/scss",
+	"extends" : "@wordpress/stylelint-config/scss",
```
adamwoodnz: Cherry picked from another PR https://github.com/WordPress/Learn/pull/921/files#r969131435

package.json
```diff
@@ -24,7 +24,11 @@
 		"create": "composer install && yarn workspaces run build && yarn run install:env",
 		"wp-env": "wp-env",
 		"lint:php": "composer run lint",
-		"format:php": "composer run format"
+		"format:php": "composer run format",
+		"lint:css": "yarn workspace wporg-learn-plugin lint:css && yarn workspace wporg-learn-theme lint:css",
+		"lint:js": "yarn workspace wporg-learn-plugin lint:js",
+		"watch:theme": "yarn workspace wporg-learn-theme start",
+		"watch:plugin": "yarn workspace wporg-learn-plugin start"
 	},
```
adamwoodnz: Also cherry picked: https://github.com/WordPress/Learn/pull/921/files#r973834595

hlashbrooke:
This works well on my local.

Regarding the duration labels - I assume the fix there is to simply rename them in the dashboard? They're a taxonomy, so that's easy to do.

adamwoodnz:
> Regarding the duration labels - I assume the fix there is to simply rename them in the dashboard? They're a taxonomy, so that's easy to do.

Yep I can update once it's landed, easy

wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
```diff
@@ -113,7 +109,11 @@
 					) );
 
 					foreach ( $levels as $level ) : ?>
-					<li><a href="<?php echo esc_url( get_post_type_archive_link( 'lesson-plan' ) ) . '?level[]=' . esc_html( $level->term_id ); ?>"><?php echo esc_html( $level->name ); ?><span class="dashicons dashicons-arrow-right-alt2"></span></a></li>
+					<li>
+						<a class="button" href="<?php echo esc_url( get_post_type_archive_link( 'lesson-plan' ) ) . '?level[]=' . esc_html( $level->term_id ); ?>">
+							<?php echo esc_html( $level->name ); ?><span class="dashicons dashicons-arrow-right-alt2"></span>
```
ryelle: I think these dashicons will need `aria-hidden` too? They aren't read aloud by VoiceOver but since you were given that feedback for the topic/audience ones, they could be by another screen reader. Maybe add `aria-hidden` to all the places dashicons are used.

```suggestion
							<?php echo esc_html( $level->name ); ?><span aria-hidden="true" class="dashicons dashicons-arrow-right-alt2"></span>
```

ryelle (approved):
This looks good — I left one comment about dashicons in other sections on the page, but once that's addressed this is good to merge 👍🏻

wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
adamwoodnz: Done, thanks

Issue #942, opened 2022-09-15
Allow ordering of Lesson Plan Topic and Audience tiles on landing page


adamwoodnz:
Admins need to be able to reorder the Lesson Plan Topic and Audience taxonomies so that they are displayed in a particular order on https://learn.wordpress.org/lesson-plans/

The Topic tiles now use the shared `topic` taxonomy and the original `wporg_lesson_category` taxonomy has been deprecated. During migration we found that the General topic was no longer being displayed first so some custom logic has been added to ensure this. See: https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2020/functions.php#L1211

If custom ordering is implemented this logic should be replaced.

![Screen Shot 2022-09-16 at 10 06 40 AM](https://user-images.githubusercontent.com/1017872/190517178-8ebb23b7-1fbe-42fb-8def-dbd405f06e59.jpg)

See design for rough order: https://www.figma.com/file/UtmG8KQv3Csy8hNtuAWkwZ/Learn-WordPress?node-id=1094%3A671

courtneyr-dev:
https://learn.wordpress.org/pathways-to-learn-wordpress/ is a beginning order of content, but always new features will come needing to be added between other things.

PR #973, opened 2022-09-28
Interim fix for the order of the Topic tiles on the Lesson Plan landing page

adamwoodnz:
See #942 

The 'Topic' tiles on the Lesson Plan landing page are currently ordered by `id ASC` and this results in the 'General' tile being last. @hlashbrooke's [suggestion for a quick fix](https://wordpress.slack.com/archives/C02RW657Q/p1663278070703839) was to simply reverse the order.

Before:
Screen Shot 2022-09-28 at 8 44 56 PM

After:
Screen Shot 2022-09-28 at 8 44 31 PM

 A more robust custom ordering solution will be developed later.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
+++ b/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
@@ -40,9 +40,11 @@ if ( '' === get_query_var( 'search' ) && empty( $_GET ) && is_post_type_archive(
 
 		<hr>
 		<?php
-		$categories = get_terms( 'wporg_lesson_category', array(
+		$categories = get_terms( array(
+			'taxonomy'   => 'wporg_lesson_category',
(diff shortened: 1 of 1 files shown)


wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
```diff
@@ -40,9 +40,11 @@
 
 		<hr>
 		<?php
-		$categories = get_terms( 'wporg_lesson_category', array(
+		$categories = get_terms( array(
+			'taxonomy'   => 'wporg_lesson_category',
```
adamwoodnz, line 44: Since 4.5.0, taxonomies should be passed via the ‘taxonomy’ argument in the $args array

https://developer.wordpress.org/reference/functions/get_terms/#description

hlashbrooke (approved):
Looks good and works for me.

PR #974, opened 2022-09-29
Breadcrumbs: Separators should be commas for screen readers, not icons

### LINKED ISSUE #977: Breadcrumbs: Screen reader announces arrows
## Description

NVDA on Windows reads double right pointed bracket for the arrow icons that are in between each breadcrumb.

1. Launch NVDA.
2. Open Firefox or Google Chrome.
3. Navigate to a course or workshop on Learn.
4. Notice how the breadcrumbs have the arrow read out as some weird name.

## Expected behavior

The screen reader should pause in between each link to avoid sounding like a huge run-on sentence. It should not be reading the visual arrows.

## Actual behavior

It reads the arrows.

## Environment information
- Latest Firefox/Google Chrome
- Windows 10 Professional

alexstine:
Fixes #977

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
+++ b/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
@@ -55,7 +55,7 @@ if ( is_post_type_archive() ) {
 		}
 
 		if ( $x < $crumb_length - 1 ) {
-			echo '<span class="screen-reader-text">,</span> <span aria-hidden="true" class="bbp-breadcrumb-sep">»</span> ';
+			echo '<span class="screen-reader-text">, </span><span aria-hidden="true" class="bbp-breadcrumb-sep"> » </span>';
 		}
 	}
 	?>
--- a/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
+++ b/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
@@ -55,7 +55,7 @@ if ( is_post_type_archive() ) {
 		}
 
 		if ( $x < $crumb_length - 1 ) {
-			echo ' <span class="bbp-breadcrumb-sep">»</span> ';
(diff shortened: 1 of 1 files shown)


adamwoodnz:
Hi @alexstine thanks for this. I'm don't doubt there is an issue here but I'm struggling to understand the symptoms with the current implementation, as I don't seem to be able to reproduce anything with VoiceOver. I'm almost certain that it's my lack of know how so I'd appreciate your help please.

Would you mind creating a bug report with more details? That'll help me learn and be better equipped to reproduce and fix this type of issue in future.

https://github.com/WordPress/Learn/issues/new?assignees=&labels=Awaiting+Triage%2C+%5BType%5D+Bug&template=bug-report-template.md&title=Bug+report

alexstine:
@adamwoodnz You are not going to experience this bug quite like me because Mac Voiceover works much differently than the Windows screen readers. This just happens to be learned experience and that comes with time. I opened the issue but I doubt it will help you.

https://github.com/WordPress/Learn/issues/977

Unfortunately, it is hard to predict how screen readers will read any type of visual icon so it is best to just eliminate them all together. Using a comma will make the screen reader natural pause in between each link. However, the natural pause thing is not an issue with Voiceover because you are navigating element by element where NVDA can navigate multiple elements at once based on CSS display property/DOM.

This is just one you'll have to trust me on. Most of the time I open PRs like this one myself because it is hard to test for this and hard to tell others to find issues this specific.

Thanks.

adamwoodnz (approved):

adamwoodnz:
> Unfortunately, it is hard to predict how screen readers will read any type of visual icon so it is best to just eliminate them all together. Using a comma will make the screen reader natural pause in between each link. However, the natural pause thing is not an issue with Voiceover because you are navigating element by element where NVDA can navigate multiple elements at once based on CSS display property/DOM.

Super useful to know this! I'm going to try to get NVDA set up locally on some sort of VM in future.

Thanks for the ticket, good to have the issue documented for our future selves and others 🙂 

Code wise this looks good, I'll get it deployed 👍

PR #997, opened 2022-10-03
Fix node sass fatal

outdoor2kode:
Fixes https://github.com/Automattic/neso/issues/11

Since the original issue was about node-sass not supporting arm env, we need to update all packages that have node-sass as their dependencies.

## Testing Results
* No errors were found after running the build scripts below (except the warning of division from wporg's sass):
`yarn workspace wporg-locale-switcher build`
`yarn workspace wporg-learn-theme build`
`yarn workspace wporg-learn-plugin build`

* No other errors were found in dev tools (local env).

* The layout of the sites looks all good.


.github/workflows/linters.yml
```diff
@@ -20,23 +20,25 @@ jobs:
...
-          ${{ runner.os }}-build-
-          ${{ runner.os }}-
+          ${{ runner.os }}-yarn-
 
-    - name: Use Node.js 12.x
+    - name: Use Node.js 14.x
       uses: actions/setup-node@v1
       with:
-        node-version: 12.x
+        node-version: 14.x
```
ryelle: Later versions of the `setup-node` action include [an optional yarn cache](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data), and can read from [a local `.nvmrc` file](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file) — for example, in [the wporg-main-2022 actions](https://github.com/WordPress/wporg-main-2022/blob/455a91e3165744864fe5b17933324a62b0be8d0d/.github/workflows/linters.yml#L19-L23). I like the `.nvmrc` file — it documents the node version and if/when we update, it only needs to be changed once.

outdoor2kode: Thanks! I've updated the actions.
Curious why we're using a certain commit of setup-node instead of something like `@v3` in [the wporg-main-2022 actions](https://github.com/WordPress/wporg-main-2022/blob/455a91e3165744864fe5b17933324a62b0be8d0d/.github/workflows/linters.yml#L19-L23)? Maybe it didn't support it back then?

ryelle: I pulled it from the [wordpress-develop](https://github.com/WordPress/wordpress-develop/) repo. I believe they review the actions for security, and using a specific SHA ensures the version used is the one that was reviewed.

outdoor2kode: Just found that they set up a dependabot to update the action/setup-node every time it has a new release.
https://github.com/WordPress/wordpress-develop/pull/3350/files

I'm thinking that it would be safe and up-to-date as well by following its [instruction](https://github.com/actions/setup-node#usage) to use `@v3` here.

PR #1011, opened 2022-10-06
Move presenter notification to Learn-specific file

### LINKED ISSUE #940: Move presenter notification to Learn-specific file
A new Profiles notification was [added](https://github.com/WordPress/wordpress.org/pull/84) when a contributor is added as a presenter on a tutorial.

I missed it during review, but the code should go in [plugins/wporg-learn/inc/profiles.php](https://github.com/WordPress/Learn/blob/0a66440da1192158ff181104162fbcd8fa9facf6/wp-content/plugins/wporg-learn/inc/profiles.php), since it's specific to Learn. `wporg-profiles-wp-activity-notifier.php ` is a general plugin that runs on most w.org sites.

Not a big deal, and definitely not urgent, but it'd help to keep all the Learn-specific activity in one place.

cc @renintw

outdoor2kode:
Fixes https://github.com/WordPress/Learn/issues/940

Related: https://github.com/WordPress/wordpress.org/pull/96

Tested on the sandbox, and everything works as expected.

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/profiles.php
+++ b/wp-content/plugins/wporg-learn/inc/profiles.php
@@ -1,9 +1,10 @@
 <?php
-/*
+/**
  * Add activity, etc to profiles.wordpress.org when users do noteworthy things.
  */
 
 namespace WPOrg_Learn\Profiles;
+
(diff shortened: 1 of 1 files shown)


iandunn (approved):
👍🏻

PR #1026, opened 2022-10-17
Adds a bordered-image utility class

### LINKED ISSUE #1025: Feature request - utility class for images
In some cases (mostly when capturing UI from the Block Editor) screenshots may contain a lot of the color white.

When included in the content on Learn, this might cause the image to blend in with the white background, and may therefore not be as impactful for the content.

It would therefore be ideal if there was a utility CSS class, that could be applied to such images in the editor so that they appear with a border.

```
border: 1px solid #ddd;
```

jonathanbossenger:
Fixes #1025 

In some cases, screenshots may contain a lot of the color white.

This might cause the image to blend in with the white background, and may therefore not be as impactful for the content.

This class can be applied to a specific image block to add a slight border around the image

### DIFF
--- a/wp-content/mu-plugins/pub/locale-switcher/package.json
+++ b/wp-content/mu-plugins/pub/locale-switcher/package.json
@@ -14,6 +14,12 @@
 		"packages-update": "wp-scripts packages-update"
 	},
 	"devDependencies": {
-		"@wordpress/scripts": "12.1.1"
+		"@wordpress/scripts": "^19.2.2"
+	},
+	"dependencies": {
+		"@wordpress/components": "^21.3.0",
+		"@wordpress/element": "^4.18.0",
+		"@wordpress/i18n": "^4.20.0",
+		"@wordpress/url": "^3.21.0"
 	}
 }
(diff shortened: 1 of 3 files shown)


hlashbrooke:
This is a great idea! Definitely happy to merge this, although I see there's a linting error there that I don't understand. @adamwoodnz could you have a look at that?

adamwoodnz:
Something happened with building an unrelated package in the repo so I've added a commit to fix this. Probably worth a smoke test of the locale switcher before merging this.

jonathanbossenger:
> Something happened with building an unrelated package in the repo so I've added a commit to fix this.

Aha, that explains the errors I was getting when trying to test locally, it works now, thanks @adamwoodnz 

I've tested the fix locally, and it works on the image block perfectly.

Adding-a-Contact-Form-to-Your-Site-learn-wordpress

>Probably worth a smoke test of the locale switcher before merging this.

I'm not sure how to test the local switcher. Is that a separate piece of functionality, or a different PR you're referring to?

jonathanbossenger:
@adamwoodnz is there any way this could be merged during the course of this week? The main reason I ask is that I'd like to use it for my block theme course, which we're aiming to publish next Monday.

adamwoodnz (approved):
👍 Works for me, including the locale switcher

Issue #1009, opened 2022-10-06
Standardise taxonomies and meta fields across post types


hlashbrooke:
We have multiple post taxonomies and meta fields on different post types - it would be great if those were standardised so we could more easily filter content using the same functions.

Fields to be standardised: Language, Topic

Language can be a new `language` meta field applied to all post types.

`topic` taxonomy currently only applied to `wporg_workshop` should be extended to apply to all post types.

Once these are added we can migrate all the post content to use the new shared taxonomy and meta field:

- Lesson plan `category` -> `topic`
- Tutorial `video_language` meta field -> `language`

PR #1030, opened 2022-10-18
Standardise taxonomies and meta fields across post types

adamwoodnz:
See #1009 

This PR sets up the new taxonomies we will be using across all post types:

1. A new `wporg_language` taxonomy to be populated with all the possible locales
2. Extends the `topic` taxonomy (currently only applied to tutorials) to all post types

Once these are in place we can start migrating all the post content to use them:

- Lesson plan `category` -> `topic`
- Tutorial `video_language` meta field -> `wporg_language`
- Course `course-category` -> `topic`

After that the old taxonomies and meta fields can be removed (or possibly hidden from the UI if there is any value in keeping them around).

A number of frontend tasks will then follow to switch filters and queries to use the new taxonomies.

#### Admin tables
| Lesson Plans | Languages |
|-|-|
| Screen Shot 2022-10-20 at 2 16 16 PM | Screen Shot 2022-10-20 at 2 16 43 PM | 

#### Posts

| Lesson Plan | Tutorial |
|-|-|
| Screen Shot 2022-10-20 at 2 25 20 PM | Screen Shot 2022-10-20 at 2 26 26 PM | 

### How to test

1. Load the admin and check that Lesson Plans, Tutorials, Meetings and Courses all have Topic and Language taxonomies.
2. Edit different post types adding Languages (see screenshots above for proposed data), and Topics (should already be plenty of terms from Tutorials).
3. Smoke test frontend after adding new terms to post content.


wp-content/plugins/wporg-learn/inc/taxonomy.php
```diff
@@ -565,6 +567,54 @@ function register_included_content() {
...
+		'hierarchical'      => true,
+		'public'            => true,
+		'query_var'         => 'wporg_language', // Prevent collisions with query params in the archive filter.
+		'show_ui'           => true,
+		'show_admin_column' => true,
+		'show_in_nav_menus' => true,
+		'show_tagcloud'     => false,
+		'show_in_rest'      => true,
+		'capabilities'      => array(
+			'assign_terms' => 'edit_any_learn_content', // See \WPOrg_Learn\Capabilities\map_meta_caps.
```
adamwoodnz: Once all the locales from `WordPressdotorg\Locales\{ get_locales_with_english_names };` have been added as terms, the intention is to only allow super admins to add new languages:

```suggestion
			'assign_terms' => 'edit_any_learn_content', // See \WPOrg_Learn\Capabilities\map_meta_caps.
			'edit_terms'     => 'manage_network',
```

wp-content/mu-plugins/pub/locale-switcher/package.json
```diff
@@ -14,6 +14,12 @@
 		"packages-update": "wp-scripts packages-update"
 	},
 	"devDependencies": {
-		"@wordpress/scripts": "12.1.1"
+		"@wordpress/scripts": "^19.2.2"
```
adamwoodnz: The build for this package was failing and upgrading this package to match the others in this repo fixed it. 

From what I can tell it seems updating the translation strings downgraded it 🤔 
https://github.com/WordPress/Learn/blame/trunk/wp-content/mu-plugins/pub/locale-switcher/package.json

wp-content/plugins/wporg-learn/inc/taxonomy.php
```diff
@@ -386,11 +387,11 @@ function register_workshop_series() {
...
 
 /**
- * Register the Workshop Topic taxonomy.
+ * Register the Topic taxonomy.
  */
-function register_workshop_topic() {
+function register_topic() {
 	$labels = array(
-		'name'                       => _x( 'Topics', 'Topic Plans associated to tutorial.', 'wporg-learn' ),
+		'name'                       => _x( 'Topic', 'Topics associated with the content.', 'wporg-learn' ),
```
ryelle: Typically this is the plural name, is there a reason to switch it to singular? (this is _very_ nitpicky, I know 😅 )

ryelle (dismissed):
This is purely from reading the code, I haven't tested it locally, but it looks good 👍🏻

wp-content/mu-plugins/pub/locale-switcher/package.json
ryelle: That's so strange, I can't find anything in the workflow that would do that. I wonder if the same thing will happen when you merge this PR.

ryelle:
Sorry for un-approving, but I went back to the original locale PR. In #112, the wporg `locales` plugin was added to handle generating the locale names based on the standardized languages set in GlotPress (see [pub/locales.php](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/locales.php) & [pub/locales/locales.php](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/locales/locales.php)). This is kept up to date with the GlotPress projects, so when new languages are added (or if a language name updates), they're populated throughout the site. If the language is added as a separate taxonomy here, it will probably become unsynced from that list at some point.

I also found the code I was thinking of when we spoke in slack – [in `wporg_archive_query_prioritize_locale`](https://github.com/WordPress/Learn/blob/19f573e52cd4f52e3fa4a5af5ef8095f086749d4/wp-content/themes/pub/wporg-learn-2020/functions.php#L446), the language is used to sort workshops.

> In order to show all workshops, but with the ones that are presented/captioned in the user's locale shown first, we need to modify the posts query in ways that can't be done through the WP_Query or WP_Meta_Query APIs. Instead, here, we're filtering the individual clauses of the query to add the pieces we need.

I'm not sure if it's _working_, or if you want it to keep working this way, but the way the taxonomy tables are set up would make this impossible (or at least much more complex).

You might be better off trying to expand the current language meta to the other post types after all.

adamwoodnz:
> I'm not sure if it's _working_, or if you want it to keep working this way, but the way the taxonomy tables are set up would make this impossible (or at least much more complex).

Yeah I think this is something I don't want to mess with! Thanks for alerting me to this.

@hlashbrooke here is the use case for sorting by language. I think we should change tack and add more meta fields to the other post types using the same code.

wp-content/plugins/wporg-learn/views/locales-list.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace WPOrg_Learn\Locale;
+
+defined( 'WPINC' ) || die();
+
+/** @var array $locales */
+?>
+
+<script type="text/javascript" id="wporg-learn-locales">
+	var wporgLearnLocales = [
```
adamwoodnz: Maybe there is a better way to namespace this?

adamwoodnz:
UPDATE: The Language taxonomy has now been dropped in favour of a new `language` meta field available on all post types. This is visible as a column on the admin tables and is editable via a post setting:

Screen Shot 2022-10-28 at 3 13 43 PM

This can be tested by editing, refreshing and returning to the admin table. `meeting` needs to be tested in sandbox.

NOTE: Tutorials already have a `video_language` meta field so that hasn't been touched. Next step for this is to migrate the data from the old field to the new and then update all the filtering and querying references to use the new field.

wp-content/plugins/wporg-learn/inc/taxonomy.php
adamwoodnz: I think I thought it was strange for the column to be named as plural, but it's multi select so seems correct 🤷 

Reverted, thanks

wp-content/plugins/wporg-learn/js/language-meta/index.js
```diff
@@ -4,33 +4,29 @@
 import { PanelRow, SelectControl } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
-import { useState, useRef } from '@wordpress/element';
+import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { registerPlugin } from '@wordpress/plugins';
 
+// Global wporgLearnLocales, see render_locales_list() in wp-content/plugins/wporg-learn/inc/post-meta.php
+// eslint-disable-next-line no-undef
```
outdoor2kode: It can also add `/* global wporgLearnLocales */` to the top of this file to avoid the warning.

wp-content/plugins/wporg-learn/views/locales-list.php
outdoor2kode: trivial: maybe we could use `const` here?

outdoor2kode (approved):
Nice work! 👍  I've tested locally and I can see `language` meta field correctly showing in the editor and `Topics` showing on the bulk editing screen / editor / admin table.

wp-content/plugins/wporg-learn/js/language-meta/index.js
adamwoodnz: Thanks, done

wp-content/plugins/wporg-learn/views/locales-list.php
adamwoodnz: I was thinking this wasn't going to be compiled so was best in es5, but you're right, our browser support policy doesn't require it.

Issue #1037, opened 2022-10-20
Migrate Lesson Plan `wporg_lesson_category` taxonomy usage to `topic` taxonomy 


adamwoodnz:
Depends on #1009 

Once all the Lesson Plan `wporg_lesson_category` data has been moved into the common `topic` taxonomy we need to:

1. Add icon and sticky meta fields to the `topic` taxonomy like the current `audience` and `wporg_lesson_category` taxonomies, and switch landing page Topic tiles over to use `topic`:

![Screen Shot 2022-11-04 at 1 07 46 PM](https://user-images.githubusercontent.com/1017872/199857833-4e23ab6a-1ff5-4239-8508-ba2a24300de6.jpg)

![Screen Shot 2022-11-04 at 1 05 36 PM](https://user-images.githubusercontent.com/1017872/199857640-59ebbb08-a61e-450f-83bd-7ccd6f1d60d9.jpg)

2. Remove the `wporg_lesson_category` taxonomy

hlashbrooke:
Confirming that the lesson plan categories have all been migrated to the topic taxonomy and the content has been added to the appropriate topics, so everything is prepped for the three tasks in this issue to be worked on.

adamwoodnz:
All the Topics required for the Lesson Plan landing pages have now had icons added and been made sticky, so this PR can land with no obvious change on the frontend, apart from possibly the order of the tiles changing.

PR #1072, opened 2022-11-03
Migrate `wporg_lesson_category` taxonomy usage to `topic`

adamwoodnz:
See #1037 

### Areas affected

1. Lesson Plan landing page Topic tiles display `topic` terms now, not `wporg_lesson_category`
3. The Lesson Plan Category taxonomy is no longer registered

### How to test
1. Ensure there is no Lesson Plan Category taxonomy
3. View the Lesson Plan landing page at `/lesson-plans/` and ensure the correct sticky terms are displayed with the correct icon, and the General topic is first

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/admin.php
+++ b/wp-content/plugins/wporg-learn/inc/admin.php
@@ -14,10 +14,12 @@ defined( 'WPINC' ) || die();
  */
 add_action( 'admin_notices', __NAMESPACE__ . '\show_term_translation_notice' );
 add_filter( 'manage_wporg_workshop_posts_columns', __NAMESPACE__ . '\add_workshop_list_table_columns' );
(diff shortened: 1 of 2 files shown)


hlashbrooke:
Would it be possible to deploy just the functionality for making topics sticky and adding icons first, then we can get that setup, and then deploying the rest of the code that removes the lesson category taxonomy and switched the landing page to use topics.

If it's not possible to deploy them separately like that, then we can collab so that the migration of all that can happen immediately after deploying.

adamwoodnz:
> Would it be possible to deploy just the functionality for making topics sticky and adding icons first, then we can get that setup, and then deploying the rest of the code that removes the lesson category taxonomy and switched the landing page to use topics.

Great idea!

adamwoodnz:
> Would it be possible to deploy just the functionality for making topics sticky and adding icons first

New PR: https://github.com/WordPress/Learn/pull/1077

PR #1117, opened 2022-11-23
Stop filtering tutorials by locale if feed

### LINKED ISSUE #1116: Non-English tutorials do not show in the tutorials RSS feed
The default view on the site shows only English tutorials until you filter the archive by language. It looks like this has also affected the tutorials RSS feed, so that it only shows English tutorials: https://learn.wordpress.org/feed/?post_type=wporg_workshop - the latest published tutorial is non-English, but it doesn’t show up in the RSS feed.

This is likely related to the filter based on the local switcher, so we probably need to implement something like `if ( ! is_feed() )` as a conditional for that filter.

adamwoodnz:
Fixes #1116


hlashbrooke (approved):
Good to go!

Issue #1006, opened 2022-10-06
Show video length on tutorials archive page


hlashbrooke:
It would be great to show the video length on the tutorials archive page. This is stored in a meta field on the tutorial post type.

adamwoodnz:
Is that this page @hlashbrooke ?

![Image](https://user-images.githubusercontent.com/1017872/194987833-1a848715-9064-4479-aa7a-0f023938b15a.jpg)

hlashbrooke:
Yes, that's the page.

PR #1125, opened 2022-11-28
Make tutorial length label translatable

adamwoodnz:
See #1006 

I missed translation of the labelling text 'Length:'


hlashbrooke (approved):
Oh - good catch - looking good now.

PR #1135, opened 2022-11-30
Adding average learner data to course sidebar

### LINKED ISSUE #1089: Display course metrics on the frontend
Sensei tracks number of learners, passing metrics, etc. natively, but since this is only available in the dashboard it is not public information. In the interests of transparency and openness, I think it would be a great idea to make this available on the frontend.

We could either do this on a dedicated page where all the courses are listed, or have it in the sidebar of each course. The sidebar option would be my preference.

The data we could probably make available is:
* Number of learners enrolled in the course
* Number of learners who completed the course (possibly displayed as a percentage of people who have completed it)
* The average grade received for the course as a whole
* The average number of days it took learners to complete the course

That data is all available in the dashboard already, so getting it out on the frontend will most likely be a matter of finding the right functions to get it and using them in the right place.

If anyone has a good reason why we _shouldn't_ make this data public, please say so! I only see it as a positive step, but there could be sides to this that I'm not seeing.

_Thanks to @mikemcalister for asking about this [on Twitter](https://twitter.com/mikemcalister/status/1592533666580004866)._

hlashbrooke:
This PR adds the average learner data for a course to the sidebar on the single course page.

Some questions from my side that I'm uncertain about:
* I did everything in the `sidebar-course.php` template file, but should the logic perhaps be added to a function elsewhere?
* Should this instead be defined as a block that can be moved around as a widget?
* Is the HTML semantic and accessible?

I'm happy to make changes accordingly but thought I'd get the initial PR up first for feedback.

Fixes #1089

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/blocks.php
+++ b/wp-content/plugins/wporg-learn/inc/blocks.php
@@ -206,20 +206,6 @@ function course_data_render_callback( $attributes, $content ) {
 		)
 	);
 
-	/**
-	// Get the total number of learners who have completed the course
-	$completions = Sensei_Utils::sensei_check_for_activity(
-		array(
-			'type'     => 'sensei_course_status',
-			'status'   => 'complete',
-			'post__in' => $course_id,
-		)
-	);
-
-	// Calculate the percentage of learners who have completed the course, rounded off to 1 decimal point
-	$percent_complete = round( ( $completions / $learners * 100 ), 1 );
-	**/
-
 	// Get the average grade scross all learners
 	$average_grade = $course_service->get_courses_average_grade( array( $course_id ) );
 
@@ -232,12 +218,6 @@ function course_data_render_callback( $attributes, $content ) {
 			'label' => __( 'Number of learners', 'wporg-learn' ),
 			'value' => $learners,
 		),
-		/**
-		'completions' => array(
-			'label' => __( 'Completion rate', 'wporg-learn' ),
-			'value' => $percent_complete . '%',
-		),
-		**/
 		'grade' => array(
 			'label' => __( 'Average grade', 'wporg-learn' ),
 			'value' => $average_grade,
--- a/wp-content/plugins/wporg-learn/inc/blocks.php
+++ b/wp-content/plugins/wporg-learn/inc/blocks.php
@@ -4,6 +4,8 @@ namespace WPOrg_Learn\Blocks;
 
 use Error;
 use Sensei_Lesson;
+use Sensei_Utils;
+use Sensei_Reports_Overview_Service_Courses;
 use function WordPressdotorg\Locales\get_locale_name_from_code;
 use function WPOrg_Learn\{get_build_path, get_build_url, get_views_path};
 use function WPOrg_Learn\Form\render_workshop_application_form;
@@ -26,6 +28,7 @@ add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_block_style_assets' 
 function register_types() {
 	register_lesson_plan_actions();
 	register_lesson_plan_details();
+	register_course_data();
 	register_workshop_details();
 	register_workshop_application_form();
 }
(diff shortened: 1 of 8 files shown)


adamwoodnz (commented):
Great start!

I think it would nice to reuse the styling from Workshop Details for this:

Screen Shot 2022-12-01 at 3 04 23 PM

This is a block in the Learn plugin, so you could replicate that and pass in the Course data. Or if there isn't value in having this as a block you could move the logic to `functions.php` and use similar markup and style in the theme. What do you think?

hlashbrooke:
Oh yeah - that styling would make a lot of sense - I'll update the markup for that.

> This is a block in the Learn plugin, so you could replicate that and pass in the Course data.

I like that - I wasn't sure how we create those blocks, but that makes a lot more sense and allows us to position things more dynamically, so I would much rather we do it that way. I'll work on changing things up there.

adamwoodnz:
Cool, just took another look and the Lesson Plan Details block is a little simpler, I'd use that as a starting point

wp-content/plugins/wporg-learn/inc/blocks.php
```diff
@@ -142,6 +145,111 @@ function lesson_plan_details_render_callback( $attributes, $content ) {
...
+	);
+
+	// Calculate the percentage of learners who have completed the course, rounded off to 1 decimal point
+	$percent_complete = round( ( $completions / $learners * 100 ), 1 );
+
+	// Get the average grade scross all learners
+	$average_grade = $course_service->get_courses_average_grade( array( $course_id ) );
+
+	// Get the average number of days it takes to complete a course
+	$avergage_days = $course_service->get_average_days_to_completion( array( $course_id ) );
```
adamwoodnz: Typo

```suggestion
	$average_days = $course_service->get_average_days_to_completion( array( $course_id ) );
```

hlashbrooke:
Add the new 'Course Data' block to the courses sidebar to test.

wp-content/plugins/wporg-learn/js/course-data/src/style.scss
```diff
@@ -0,0 +1,42 @@
...
+.wp-block-wporg-learn-course-data {
+	border-top: 1px solid #e2e4e7;
+	list-style: none;
+	margin: 0 0 2rem 0;
+
+	> li {
+		display: flex;
+		justify-content: space-between;
+		border-bottom: 1px solid #e2e4e7;
+		padding: 0.5rem 0 !important;
```
adamwoodnz: This `!important` breaks the print styling. I can see it's been added to override the `.lp-details` styling, so I think you can remove that wrapper from sidebar instead.

```suggestion
		padding: 0.5rem 0;
```

Screen Shot 2022-12-01 at 4 28 52 PM

adamwoodnz (changes requested):
One change, otherwise LGTM!

hlashbrooke:
Changes made - the CSS is no longer using `!important` and I removed the `lp-details` class from the sidebar markup. I confirmed that removing that class did not cause any layout issues with the existing sidebar blocks.

adamwoodnz (approved):
LGTM :shipit: 🚀

wp-content/plugins/wporg-learn/inc/blocks.php
```diff
@@ -142,6 +145,115 @@ function lesson_plan_details_render_callback( $attributes, $content ) {
...
+		array(
+			'type'     => 'sensei_course_status',
+			'status'   => 'complete',
+			'post__in' => $course_id,
+		)
+	);
+
+	// Calculate the percentage of learners who have completed the course, rounded off to 1 decimal point
+	$percent_complete = round( ( $completions / $learners * 100 ), 1 );
+	**/
```
adamwoodnz: Block comment like this

```suggestion
	/*
	// Get the total number of learners who have completed the course
	$completions = Sensei_Utils::sensei_check_for_activity(
		array(
			'type'     => 'sensei_course_status',
			'status'   => 'complete',
			'post__in' => $course_id,
		)
	);

	// Calculate the percentage of learners who have completed the course, rounded off to 1 decimal point
	$percent_complete = round( ( $completions / $learners * 100 ), 1 );
	*/
```

But I would lean on YAGNI here, just delete it, easy enough to add again later right?

wp-content/plugins/wporg-learn/inc/blocks.php
```diff
@@ -142,6 +145,115 @@ function lesson_plan_details_render_callback( $attributes, $content ) {
...
+		'learners' => array(
+			'label' => __( 'Number of learners', 'wporg-learn' ),
+			'value' => $learners,
+		),
+		/**
+		'completions' => array(
+			'label' => __( 'Completion rate', 'wporg-learn' ),
+			'value' => $percent_complete . '%',
+		),
+		**/
```
adamwoodnz: Again I would remove this if we don't want it, but it can be fixed with
```suggestion
		
		/*
		'completions' => array(
			'label' => __( 'Completion rate', 'wporg-learn' ),
			'value' => $percent_complete . '%',
		),
		*/
```

hlashbrooke:
Commented out code has been removed - PR is ready to merge now.

Issue #1008, opened 2022-10-06
Add filtering to the online workshops calendar


hlashbrooke:
Add filters to the online workshops calendar for language and topic.

This will require the language meta field and a topic taxonomy to be added to the `meeting` post type.

adamwoodnz:
The likely depends on https://github.com/WordPress/Learn/issues/1009 for adding the meta fields.

adamwoodnz:
Topic taxonomy and Language meta field have been added to `meeting` but there is currently no way to edit language on this post type only. Need to add a meta box on the edit meeting screen.

PR #1139, opened 2022-12-01
Adding language metabox to the meeting post type

hlashbrooke:
This PR adds the language metabox to the meeting post type - this completed the work done in #1009 and is necessary for the filtering in #1008 to be possible.


hlashbrooke:
I used the existing metabox files to add this new one - I hope that's the right place for it, but I'm happy to move things around if there's a better location.

hlashbrooke:
New 'Meeting Language' metabox shows up in the sidebar: 

Screenshot 2022-12-02 at 11 16 52 AM

The meta field registration and default value were already added in a previous PR, so this PR is simply rendering the metabox and saving the field when the post is updated.

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -384,6 +386,19 @@ function add_workshop_metaboxes() {
 	);
 }
 
+/**
+ * Add meta boxes to the Edit Meeting screen.
+ */
+function add_meeting_metaboxes( $post_type = '' ) {
+	add_meta_box(
+		'meeting-language',
+		__( 'Meeting Language', 'wporg_learn' ),
```
adamwoodnz: IMO this isn't necessary, and using just 'Language' makes it consistent with other meta boxes
```suggestion
		__( 'Language', 'wporg_learn' ),
```

hlashbrooke: tbh I had thought of that but just left it in the end, so good to have confirmation that you thought the same.

wp-content/plugins/wporg-learn/views/metabox-meeting-language.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * Template for Other Contributors metabox
```
adamwoodnz:
```suggestion
 * Template for Meeting Language metabox
```

wp-content/plugins/wporg-learn/views/metabox-meeting-language.php
```diff
@@ -0,0 +1,21 @@
...
+
+/** @var WP_Post $post */
+/** @var array $locales */
+/** @var string $language */
+?>
+
+<?php wp_nonce_field( 'meeting-metaboxes', 'meeting-metabox-nonce' ); ?>
+
+<p>
+	<select id="meeting-language" name="meeting-language" style="width: 100%;">
```
adamwoodnz: A11y issue with this id, it ends up being a duplicate:
Screen Shot 2022-12-05 at 3 14 03 PM

Suggest changing this to:
```suggestion
	<select id="meeting-language-selector" name="meeting-language" aria-label="Meeting Language" style="width: 100%;">
```

Aria label added as we don't want to have a visible label, it will just be a duplicate of the meta box title.

adamwoodnz (changes requested):
Works for me, just one a11y fix needed

hlashbrooke:
I have added all the changes you suggested, including the a11y fix - that was a good catch! This is ready to go now.

adamwoodnz (approved):
LGTM :shipit:

PR #1154, opened 2022-12-05
Add banner with survey link

### LINKED ISSUE #1153: Add persistent banner linking to individual learner survey
It would be great if we had a persistent link to the Individual learn survey: https://learn.wordpress.org/individual-learner-survey/

I'm thinking something like this (mocked up in the inspector):

Screenshot 2022-12-05 at 1 31 31 PM

That would be on every page of the site, including the home page.

hlashbrooke:
This PR adds a persistent banner below the primary site navigation that includes a link to the individual learner survey. Fixes #1153


hlashbrooke:
I added a new SCSS file for this and I think I built everything correctly from there, but let me know if not.

wp-content/themes/pub/wporg-learn-2020/css/components/_notice.scss
```diff
@@ -0,0 +1,48 @@
+.notice {
```
adamwoodnz, line 1: Most of this came from `_locale_notice.scss`, only the survey notice style is new

wp-content/plugins/wporg-learn/inc/locale.php
```diff
@@ -60,8 +60,8 @@ function update_locale_data() {
  * Register style and script assets for later enqueueing.
  */
 function register_assets() {
-	// Locale switcher script.
-	wp_register_script(
```
adamwoodnz, line 64: This meant the script was never loaded on the frontend

adamwoodnz:
Screen Shot 2022-12-05 at 4 55 52 PM

Screen Shot 2022-12-05 at 4 54 49 PM

Screen Shot 2022-12-05 at 4 55 02 PM

hlashbrooke:
This LGTM :shipit:

PR #1252, opened 2023-01-16
 Fix tutorials with default language not showing in archive when filtered

### LINKED ISSUE #1172: Language meta field not being saved correctly on tutorials
When saving a new tutorial, the default language is set to 'English', so we would naturally leave it unchanged for new tutorials in English. However, it doesn't seem to save the language properly to the post meta. This causes a problem where the tutorial doesn't show up in the archive after saving because the archive query checks that the language meta field has been set.

The fix is to change it to any other language, save the post, change it back to English, and then save the post again. It's easy to work around, but also easy to forget and quite an annoying issue.

adamwoodnz:
Fixes #1172

### Bug
Newly published tutorials with the default language (en_US) do not show in the [archive](https://learn.wordpress.org/tutorials/) when filtered by that language.

### Context
We have a [custom query for the Tutorial archive](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2020/functions.php#L477) view which prioritises Tutorials with either `language` or `video_caption_language` meta fields matching the user's language.

This universal `language` field has only recently been added, and the `video_language` field was deprecated, with the query switching to the new field. The crucial difference appears to be that `language` is using a `PluginDocumentSettingPanel` block, whereas the old `video_language` field was using a meta box.

The bug presents because publishing with the block does not save the default `language` value, resulting in the tutorial being filtered out of the query results.

### The fix
Force the default `language` post meta value to be saved when saving the tutorial if it is not set.

### How to test
1. Add a new tutorial, leaving the meta language field with the default value 'English'
2. Publish and navigate to http://localhost:8888/tutorials/
3. Your tutorial should be in the results towards the end. Note that searching does not use this query so you need to page through until you find it.
4. Filter the tutorials by language: English [en_US] and observe that the tutorial is still in the results

#### Edge testing
Try different combinations of tutorial language, frontend locale setting and frontend filtering

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -18,7 +18,7 @@ add_action( 'add_meta_boxes', __NAMESPACE__ . '\add_lesson_plan_metaboxes' );
 add_action( 'add_meta_boxes', __NAMESPACE__ . '\add_workshop_metaboxes' );
 add_action( 'add_meta_boxes', __NAMESPACE__ . '\add_meeting_metaboxes' );
 add_action( 'save_post_lesson-plan', __NAMESPACE__ . '\save_lesson_plan_metabox_fields' );
-add_action( 'save_post_wporg_workshop', __NAMESPACE__ . '\save_workshop_metabox_fields' );
+add_action( 'save_post_wporg_workshop', __NAMESPACE__ . '\save_workshop_meta_fields' );
 add_action( 'save_post_meeting', __NAMESPACE__ . '\save_meeting_metabox_fields' );
 add_action( 'admin_footer', __NAMESPACE__ . '\render_locales_list' );
 add_action( 'enqueue_block_editor_assets', __NAMESPACE__ . '\enqueue_editor_assets' );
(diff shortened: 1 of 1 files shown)


wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -181,6 +181,9 @@ function register_misc_meta() {
 				'default'           => 'en_US',
 				'sanitize_callback' => __NAMESPACE__ . '\sanitize_locale',
 				'show_in_rest'      => true,
+				'auth_callback'     => function() {
+					return current_user_can( 'edit_posts' );
+				},
```
adamwoodnz, line 186: This is unrelated, but I believe I should have added it when registering this new field, so improving now.

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -521,6 +521,16 @@ function save_workshop_metabox_fields( $post_id ) {
...
 		}
 	}
+
+	// This language meta field is rendered in the editor sidebar using a PluginDocumentSettingPanel block,
+	// which won't save the field on publish if it has the default value.
+	// Our custom workshops query for locale prioritized tutorials (see functions.php `wporg_archive_query_prioritize_locale`)
+	// depends on it being set, so we force it to be updated after saving:
+	$language         = get_post_meta( $post_id, 'language', true );
+	$language_default = 'en_US';
+	if ( ! isset( $language ) || $language_default === $language ) {
```
jonathanbossenger, line 531: Forgive me if I'm reading this wrong, but is the $language_default === $language check required? 

if $language is already en_US, do we still need to update it?

jonathanbossenger (commented):
@adamwoodnz everything else here looks good to me, I just have one question on the conditional check.

wp-content/plugins/wporg-learn/inc/post-meta.php
adamwoodnz: Yeah good spotting. When debugging I found that the meta field is read as the default value even if the value is not saved in the db. So in this case there is no post meta entry for `language` but still `$language = 'en_US'`, and so we still need to force save to actually add a post meta entry. It's definitely not ideal.

Issue #1003, opened 2022-10-06
New meta field for recorded video on online workshops


hlashbrooke:
Add a new meta field to the online workshops (`meeting`) to add the WPTV URL of recorded online workshops. This just needs to be a text field with URL validation.

outdoor2kode:
@hlashbrooke I'm wondering where is the best place to add the text field for WPTV URL?

#### Is it custom fields? 
<img width="1649" alt="image" src="https://user-images.githubusercontent.com/18050944/213251409-16ade622-bafa-491b-9ba6-844556c363d7.png">
<img width="1609" alt="image" src="https://user-images.githubusercontent.com/18050944/213252027-8fe65dc2-33f7-46c8-98e9-252e031e96e6.png">


#### Or adding a new one similar to the `language` on the right-hand side? Which would also show a new column on the `All meetings` page.
<img width="1913" alt="image" src="https://user-images.githubusercontent.com/18050944/213251842-5ad3d4e8-6e32-4fe4-8a94-9ce363408765.png">
<img width="1744" alt="image" src="https://user-images.githubusercontent.com/18050944/213252519-b81c2847-aa23-4af5-b3d3-4bb45bea4217.png">

And I'm also curious about where this metadata would be used?

Thanks.

adamwoodnz:
> Or adding a new one similar to the language on the right-hand side? Which would also show a new column on the All meetings page.

@renintw I believe this was the intention

And from memory on the front end there would be a new link on past meetings in the modal:

![Screen Shot 2023-01-19 at 4 01 23 PM](https://user-images.githubusercontent.com/1017872/213345528-43839576-36c5-48fe-aefe-a042fc7a5153.jpg)

outdoor2kode:
@adamwoodnz 
> And from memory on the front end there would be a new link on past meetings in the modal:

I'm not quite sure which page or format "past meetings" refers to, so I tried to search through the Slack history using keywords but couldn't find any relevant discussions. Do you remember if the link should go to a Make post that has a section for past meetings, or should all the past online workshops be gathered in one post (like an archive page)? If it's the latter, I'm curious if there were any discussions on the style for that archive page. 

Or, is it [this page](https://wordpress.tv/category/learn-wordpress-online-workshops/) that the `View recorded online workshops` button on this [page](https://learn.wordpress.org/online-workshops/) links to? If so, since we already have the button, do we still want to add an additional link in the modal? 

Thanks!

adamwoodnz:
[...]

PR #1448, opened 2023-03-30
Change the text copy to 'register for event'

outdoor2kode:
See https://github.com/WordPress/Learn/issues/1003#issuecomment-1482370137.

This PR

* Replace text copy as requested
* Add built CSS files path to `.gitignore`

## Screenshot

image


adamwoodnz (approved):
LGTM 👍 

It will be interesting to see what happens when the sync is run, in theory those files are still in SVN control so once built they'll be committed as usual 🤞

outdoor2kode:
Things work as expected 🤞 🎉 

Screen Shot 2023-03-30 at 9 39 21 AM

Screen Shot 2023-03-30 at 9 40 47 AM

PR #1500, opened 2023-04-18
Add filtering of lesson plans by language

### LINKED ISSUE #874: Add language filtering option for lesson plans
URL: https://learn.wordpress.org/lesson-plans/
It would be very helpful if we had a language filtering options (language category or tag). 
So the locales team could better promote the lesson-plans written in their own language, ex. Greek 

Bests, 
Lena

adamwoodnz:
Closes #874 

This PR adds admin and frontend filtering of Lesson Plans, expanding on the functionality previously added for Tutorials.

### Screenshots

| Admin | Frontend | Cards |
|-|-|-|
| Screen Shot 2023-04-20 at 1 41 41 PM | Screen Shot 2023-04-20 at 1 43 22 PM | Screen Shot 2023-04-20 at 1 44 31 PM |

#### IMPORTANT before you start
In the Lesson Plan list if the Language field displays `English [en_US]` it is most likely that the Lesson Plan actually has no meta language set and the column is displaying the default value. The easiest way to fix this is to bulk edit all these entries and set them to English. This has already been done in prod, so if you're testing in sandbox things should just work.

##### Admin

1. Ensure you have Lesson Plans with multiple languages in your environment
2. In the admin, navigate to the Lesson Plan list. Check that the filter language dropdown has options covering all the items listed. If you find there is no English option, ensure you have read the note above.
3. Filter the admin list. Changing post status should update the languages in the filter dropdown.
4. Add a new Lesson Plan and leave the language as the default
5. Navigate to the Tutorial list page. Smoke test that the filtering still works there, matching the behaviour for Lesson Plans.

##### Frontend
1. Open the frontend Lesson Plan archive at http://localhost:8888/lesson-plans/?_view=all
2. Note the new language filter at the top of the sidebar
3. Test filtering and check the results against the admin
4. Check that the cards in the left have the correct language displayed
6. Filter by `English [en_US]` and check that the Lesson Plan you created in step 4 above is displayed
7. Test the 'Clear all filters' action. You should stay on the archive page and not be redirected to the Lesson Plan landing page.
8. Navigate to the Tutorials archive and smoke test filtering using the language and subtitle dropdowns

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -269,28 +269,27 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
 function get_available_post_type_locales( $meta_key, $post_type, $label_language = 'english', $post_status ) {
 	global $wpdb;
 
-	$results = isset( $post_status )
-		? $wpdb->get_col( $wpdb->prepare(
-			"
-				SELECT DISTINCT postmeta.meta_value
-				FROM {$wpdb->postmeta} postmeta
-					JOIN {$wpdb->posts} posts ON posts.ID = postmeta.post_id AND posts.post_type = %s AND posts.post_status = %s 
-				WHERE postmeta.meta_key = %s
-			",
-			$post_type,
-			$post_status,
-			$meta_key
-		) )
-		: $wpdb->get_col( $wpdb->prepare(
-			"
-				SELECT DISTINCT postmeta.meta_value
-				FROM {$wpdb->postmeta} postmeta
-					JOIN {$wpdb->posts} posts ON posts.ID = postmeta.post_id AND posts.post_type = %s 
-				WHERE postmeta.meta_key = %s
-			",
-			$post_type,
-			$meta_key
-		) );
+	$and_post_status = '';
+	if ( in_array(
+		$post_status,
+		array( 'all', 'publish', 'draft', 'trash', 'private', 'needs-vetting', 'unlisted', 'declined' ),
+		true
+	) ) {
+		$and_post_status = "AND posts.post_status = '$post_status'";
+	}
+
+	$results = $wpdb->get_col( $wpdb->prepare(
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $and_post_status only includes $post_status if it matches an allowed string.
+		"
+		SELECT DISTINCT postmeta.meta_value
+		FROM {$wpdb->postmeta} postmeta
+			JOIN {$wpdb->posts} posts ON posts.ID = postmeta.post_id AND posts.post_type = %s $and_post_status
+		WHERE postmeta.meta_key = %s
+	",
+		$post_type,
+		$meta_key
+		// phpcs:enable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+	) );
 
 	if ( empty( $results ) ) {
 		return array();
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -261,12 +261,12 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
  *
  * @param string $meta_key
  * @param string $post_type
- * @param string $label_language
  * @param string $post_status
+ * @param string $label_language
  *
  * @return array
  */
-function get_available_post_type_locales( $meta_key, $post_type, $label_language = 'english', $post_status ) {
+function get_available_post_type_locales( $meta_key, $post_type, $post_status, $label_language = 'english' ) {
 	global $wpdb;
 
 	$and_post_status = '';
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -257,17 +257,18 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
 }
 
 /**
- * Get a list of locales that are associated with at least one workshop.
+ * Get a list of locales that are associated with at least one post of the specified type.
  *
- * Optionally only published workshops.
+ * Optionally only published posts.
  *
  * @param string $meta_key
+ * @param string $post_type
  * @param string $label_language
  * @param bool   $published_only
  *
  * @return array
  */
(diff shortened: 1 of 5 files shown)


wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -345,6 +353,15 @@ function save_lesson_plan_metabox_fields( $post_id ) {
...
 	update_post_meta( $post_id, 'slides_download_url', $download_url );
+
+	// This language meta field is rendered in the editor sidebar using a PluginDocumentSettingPanel block,
+	// which won't save the field on publish if it has the default value.
+	// Our filtering by locale depends on it being set, so we force it to be updated after saving:
+	$language         = get_post_meta( $post_id, 'language', true );
+	$language_default = 'en_US';
+	if ( ! isset( $language ) || $language_default === $language ) {
+		update_post_meta( $post_id, 'language', $language_default );
+	}
```
adamwoodnz, line 356: This follows the same approach that we had to take with Tutorials: https://github.com/WordPress/Learn/blob/trunk/wp-content/plugins/wporg-learn/inc/post-meta.php#L525

`$language` can be read as the default value of `en_US` even if it's not set, hence forcing save in this case as well, on line 362.

jonathanbossenger:
Nice work @adamwoodnz I've made a note to review this ASAP.

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -257,35 +257,40 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
...
+		: $wpdb->get_col( $wpdb->prepare(
+			"
+				SELECT DISTINCT postmeta.meta_value
+				FROM {$wpdb->postmeta} postmeta
+					JOIN {$wpdb->posts} posts ON posts.ID = postmeta.post_id AND posts.post_type = %s 
+				WHERE postmeta.meta_key = %s
+			",
+			$post_type,
+			$meta_key
+		) );
```
iandunn: I worry that duplicating the query here could lead to them accidentally getting out of sync, and it seems less readable to me.

What do you think about integrating `$post_status` into the query conditionally? That seems like the common approach to me. Here's a few examples:

https://github.com/WordPress/wordcamp.org/blob/e3638a232799b31827d4cb107ec503919b0b04ff/public_html/wp-content/plugins/wordcamp-payments-network/includes/payment-requests-list-table.php#L50-L109

https://github.com/WordPress/WordPress/blob/ead3a2ec81f88842c2cba8b8a693871d18522104/wp-includes/canonical.php#L962-L964

https://github.com/WordPress/WordPress/blob/ead3a2ec81f88842c2cba8b8a693871d18522104/wp-includes/class-wp-query.php#L2536-L2537

https://github.com/WordPress/WordPress/blob/ead3a2ec81f88842c2cba8b8a693871d18522104/wp-includes/canonical.php#L940-L950

iandunn (approved):
Looks great! I left a minor suggestion, but I don't feel strongly about it.

jonathanbossenger (approved):
The functionality works great, nice one @adamwoodnz

wp-content/plugins/wporg-learn/inc/post-meta.php
adamwoodnz: Yeah it's a fair point. I was trying to remove the need for `phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared`, but it seems to be common practice. Changed to something closer resembling the previous code, and updated the comment.

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -257,32 +257,32 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
...
  * @return array
  */
-function get_available_workshop_locales( $meta_key, $label_language = 'english', $published_only = true ) {
+function get_available_post_type_locales( $meta_key, $post_type, $label_language = 'english', $post_status ) {
 	global $wpdb;
 
 	$and_post_status = '';
-	if ( $published_only ) {
-		$and_post_status = "AND posts.post_status = 'publish'";
+	if ( in_array( $post_status, get_post_stati(), true ) ) {
```
adamwoodnz, line 273: @iandunn are you satisfied with the security of this approach? `$post_status` is passed in from `INPUT_GET` [here](https://github.com/WordPress/Learn/pull/1500/files#diff-2644427c2043c94e7dcc0a36501c5142deb7b72c046baa6210e2cd61cf16b8a3R238).

This line attempts to ensure that only safe strings matching `post_status` values in the system are included in the query.

adamwoodnz: `post_status` is also only sourced from `INPUT_GET` in the admin. On the frontend `post_status` is always hard coded to `publish`.

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -257,32 +257,32 @@ function get_workshop_duration( WP_Post $workshop, $format = 'raw' ) {
...
  * @param string $meta_key
+ * @param string $post_type
  * @param string $label_language
- * @param bool   $published_only
+ * @param string $post_status
  *
  * @return array
  */
-function get_available_workshop_locales( $meta_key, $label_language = 'english', $published_only = true ) {
+function get_available_post_type_locales( $meta_key, $post_type, $label_language = 'english', $post_status ) {
```
iandunn: I get this notice on PHP `8.0.28`. We don't run that on prod yet, but will in the near future, so it'd be good to be prepared now IMO.

`PHP Deprecated:  Required parameter $post_status follows optional parameter $label_language in wp-content/plugins/wporg-learn/inc/post-meta.php on line 269`

wp-content/plugins/wporg-learn/inc/post-meta.php
iandunn: Yeah, that looks like a great way to guarantee it's a valid value 👍🏻

iandunn (approved):
Looks good!

wp-content/plugins/wporg-learn/inc/post-meta.php
adamwoodnz: Fixed in https://github.com/WordPress/Learn/pull/1500/commits/4cb6519b767e474cbdbea77e3f7c8416f7909388

Issue #941, opened 2022-09-15
Header's search field overflows


vagelisp:
## Description
It looks like the search field in the header causes the horizontal scroll.

## Screenshots or screen recording (optional)
![IMG_0005](https://user-images.githubusercontent.com/9304640/190515776-d9ae0568-d91b-4c04-8257-ad7342e3ba1c.jpeg)

![IMG_0153](https://user-images.githubusercontent.com/9304640/190515770-94ee3c17-1f59-42e0-95c5-bb63bb2d3cfa.jpeg)

## Environment information
-  iPad Mini on Safari
-  iPad on Chrome

rjekic:
This issue is already fixed in this PR - https://github.com/WordPress/Learn/pull/995, so it will be closed. It can be reopened if needed.

cc @devmuhib009 @WordPress/learn-editors

PR #1621, opened 2023-06-08
Changed css to fix navigation issues in long menu names.

iamchetanp:
I have changed CSS to fix overflow issues in longer menu names. Specially in translation.

#941


speedyprem (approved):
Code looks good to me.

PR #1641, opened 2023-06-22
Breadcrumbs: Try to add a single space to avoid screen reader run-on sentences

### LINKED ISSUE #644: Breadcrumbs aren't read individually by NVDA
As explained by @alexstine 

> This is a Windows-specific bug. Voiceover and Windows screen readers work differently in the way that Voiceover traverses the content one item at a time where NVDA reads the links like a sentence due to the display: inline; set. I think that is okay for breadcrumbs, but to get the best verbosity results, it needs to be "Link 1, Link2, etc." Breadcrumbs can be tricky... I wonder what would happen if we tried rendering them as display: block;. Then there likely would not be a need for the screen reader text anymore.

We should try using `display: block` breadcrumbs within a flex row container if the attempted fix in WordPress/Learn#1641 doesn't work.

alexstine:
See WordPress/wporg-mu-plugins#644 

Not sure if this will work or not, but worth a try. Might need to use a non-breaking space here.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
+++ b/wp-content/themes/pub/wporg-learn-2020/template-parts/component-breadcrumbs.php
@@ -55,7 +55,7 @@ if ( is_post_type_archive() ) {
 		}
 
 		if ( $x < $crumb_length - 1 ) {
-			echo '<span class="screen-reader-text">, </span><span aria-hidden="true" class="bbp-breadcrumb-sep"> » </span>';
+			echo '<span class="screen-reader-text">,  </span><span aria-hidden="true" class="bbp-breadcrumb-sep"> » </span>';
 		}
 	}
 	?>


adamwoodnz:
> Not sure if this will work or not, but worth a try. Might need to use a non-breaking space here.

I think the browser will combine these spaces. Are you not able to test it?

For me Voiceover reads the links separately when I use tab, maybe I'm navigating the content differently?

alexstine:
@adamwoodnz This is a Windows-specific bug, sorry, should have spelled that out. Voiceover and Windows screen readers work differently in the way that Voiceover traverses the content one item at a time where NVDA reads the links like a sentence due to the `display: inline;` set. I think that is okay for breadcrumbs, but to get the best verbosity results, it needs to be "Link 1, Link2, etc." Breadcrumbs can be tricky... I wonder what would happen if we tried rendering them as `display: block;`. Then there likely would not be a need for the screen reader text anymore. That's above my styling paygrade though not having vision.

adamwoodnz:
> I wonder what would happen if we tried rendering them as `display: block;`. Then there likely would not be a need for the screen reader text anymore.

Yeah this might be worth a try. `display: block` children in a flex row would be visually the same. I've been meaning to get a windows environment set up on my mac so I can test with NVDA.

adamwoodnz (commented):
Requires testing on Windows with NVDA

adamwoodnz:
Closing as this PR affects the soon to be replaced 2020 theme

PR #1642, opened 2023-06-22
Search form label: Make it clear to screen readers that search occurs for Learn only

alexstine:
Currently, the screen reader label does not contain any information about what is being searched. This PR fixes that.


adamwoodnz (approved):
👍 Works for me

Using Voiceover I get very a verbose description when tabbing to the input, due to the label, placeholder and submit button: "Search resources on Learn. Search for a learning resource. Search."

Not sure if we can improve this, or if it's not an issue with better/different tools?

alexstine:
@adamwoodnz This is caused by screen readers trying to catch up. For years, every screen reader ignored placeholder text. Some of them are slowly adding support for it so here we are. No real good solution at this point, I think.

PR #1644, opened 2023-06-22
Single: Fix an accessibility bug in Google Chrome where sidebars are not recognized

alexstine:
This bug happens because the `<section>` tag gets a `role="generic"` in Google Chrome. Since this tag wraps the `<aside>`, the sidebar semantics are removed from the accessibility tree. Let's use `<div>` instead.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/template-parts/content-single.php
+++ b/wp-content/themes/pub/wporg-learn-2020/template-parts/content-single.php
@@ -11,7 +11,7 @@
 
 <article id="post-<?php the_ID(); ?>" <?php post_class(); ?>>
 
-	<section>
+	<div>
 		<header class="row align-middle between section-heading section-heading--with-space">
 			<h1 class="section-heading_title h2"><?php the_title(); ?></h1>
 		</header>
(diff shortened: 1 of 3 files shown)


adamwoodnz:
Thanks for this @alexstine 

Please help me understand this change, as the difference isn't obvious to me the when I inspect the tree in Chrome:

| Local | Prod |
|-|-|
| Screen Shot 2023-06-23 at 1 51 11 PM | Screen Shot 2023-06-23 at 1 44 58 PM |

Both show the aside as being `generic`, and I can't see any other changes, what am I missing?

alexstine:
@adamwoodnz I tried testing this in Google Chrome and found the `<section>` caused it. Google Chrome reads the `<aside>` fine but when I wrap it in a `<section>`, it never works. I tried it in just a test.html document separate from this PR. My local env isn't currently working. 😞 I think we should try it, no downside if styling is the same.

adamwoodnz:
> @adamwoodnz I tried testing this in Google Chrome and found the `<section>` caused it. Google Chrome reads the `<aside>` fine but when I wrap it in a `<section>`, it never works. I tried it in just a test.html document separate from this PR. My local env isn't currently working. 😞 I think we should try it, no downside if styling is the same.

Sorry just realised that providing screenshots of the tree isn't helpful at all 🤦 

Sure, I trust your testing, happy to try it 👍

adamwoodnz (approved):

PR #1652, opened 2023-06-26
Adding the interactive code block plugin

jonathanbossenger:
This PR adds the interactive code block plugin to Learn WordPress, which enables running PHP code directly in the browser thanks to WordPress Playground.

https://wordpress.org/plugins/interactive-code-block/

While this plugin is still largely experimental, the idea of this PR is to get the base plugin installed, so that when future updates are rolled out, we can get them more quickly.


jonathanbossenger:
@adamwoodnz I'm not sure if this code needs to be reviewed separately other than the review that would have taken place getting it approved for the plugin repo?

sumitsinghwp:
I will test this PR and let you know @jonathanbossenger

alexstine:
This also needs accessibility testing. The WP Playground is largely a mess ATM. Another example of how we failed to check before promoting and making Playground live. I'll try to have a look soon.

adamwoodnz (commented):
Is there a reason to add this to source? We don't usually; Sensei Pro is an exception because it's a paid plugin.

We can install this in prod at any time if it meets our requirements. I imagine plugin review is probably sufficient in terms of code, however it sounds like there are accessibility concerns.

For local dev I think it should be added to the list of plugins in composer.json `require-dev`.

jonathanbossenger:
> Is there a reason to add this to source? We don't usually; Sensei Pro is an exception because it's a paid plugin.
> 
> We can install this in prod at any time if it meets our requirements. I imagine plugin review is probably sufficient in terms of code, however it sounds like there are accessibility concerns.
> 
> For local dev I think it should be added to the list of plugins in composer.json `require-dev`.

Honestly, I wasn't aware it was possible to add additional plugins specifically to Learn any other way 😁

jonathanbossenger:
I'm going to close this PR for now, we can always revisit this later.

PR #1695, opened 2023-07-06
Revert "Single: Fix an accessibility bug in Google Chrome where sidebars are not recognized"

adamwoodnz:
Reverts WordPress/Learn#1644


alexstine (approved):
👍 

Here is the related bug report and it was a verified issue.

https://bugs.chromium.org/p/chromium/issues/detail?id=1459657

Hopefully they'll ship a fix for it in an upcoming release.

Issue #1813, opened 2023-08-20
UI - Search icon is broken on Lesson Plans page


ankitguptaindia:
## Description

Search icon is broken on the Lesson Plans page. 

## Step-by-step reproduction instructions

1. Goto the URL https://learn.wordpress.org/lesson-plans/?search=testing
2. Look at the search icon in search bar

## Expected behaviour
Icon should have proper background color

## Actual behaviour
Icon has missing background color 

## Screenshots or screen recording (optional)

Actual - 

![Screenshot 2023-08-20 at 4 09 29 PM](https://github.com/WordPress/Learn/assets/6384299/dad6fe0b-b9ce-4d9c-a3d4-9302158b4e5f)

Expected -

![Screenshot 2023-08-20 at 4 14 30 PM](https://github.com/WordPress/Learn/assets/6384299/ad3cc771-ff48-48eb-85fd-56467f23212e)

## Environment information

Browser -  

- Chrome Version - 115.0
- OS - macOS Monterey 12.3.1

jonathanbossenger:
Duplicate of #1808

Issue #1808, opened 2023-08-18
Long Topic Title in Learn WordPress overlapping with the searchbox


jhimross:
## Description
Long topic title will cause overlapping display with the searchbox

## Step-by-step reproduction instructions
1. Go to this page https://learn.wordpress.org/topic/extending-wordpress/page/2/
2. See the Title and Search bar. https://share.getcloudapp.com/E0uL1Gw6

## Expected behaviour
It should display like this:- https://share.getcloudapp.com/Apug0BAQ

## Actual behaviour
But if the title is longer, the issue will happen. https://share.getcloudapp.com/geuPNv82

## Environment information
-I can replicate the issue in any browser, any OS, and mostly when in desktop view

## Additional Info
Simple fix is by decreasing the size of the .section-heading_title.h2

```
.section-heading_title.h2 {
    font-size: 2.0rem;
}
```
See https://share.getcloudapp.com/d5uD7rPb

cynthianorman:
I am interested in working on this issue. For the solution so far, I'm leaning towards placing the search box within its own row placed just above the Topic's Title. The fix of reducing the font size of H2 will affect all pages so that may not be desirable.

cynthianorman:
Here's my proposed update:

- archive-lesson-plan.php
- lines 190 to 193
```
		<div class="row align-middle between section-heading section-heading--with-space">
			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
			<?php get_template_part( 'template-parts/component', 'archive-search' ); ?>
		</div>
```
- replace with lines 190 to 195
```
		<div class="row align-right">
			<?php get_template_part( 'template-parts/component', 'archive-search' ); ?>
		</div>
		<div class="row align-left">
			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
		</div>
```

Expected result:
https://snipboard.io/1ME5ue.jpg

jonathanbossenger:
@cynthianorman that sounds like a good solution.

Are you comfortable to go ahead and create a pull request with this change? This will allow someone to test your proposed change on their own local copy of the site, and provide feedback.

jonathanbossenger:
Hi @cynthianorman that looks like a solid solution. If you're happy to go ahead and create a pull request for this change, I can review it for you. Please feel free to ping me here once the PR is ready.

cynthianorman:
Thank you @jonathanbossenger the PR is ready

[...]

PR #1829, opened 2023-08-28
Update archive-lesson-plan.php

cynthianorman:
Splitting the row in 2 parts with the search field top right align and the heading title just below it left align.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
+++ b/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
@@ -187,11 +187,9 @@ if ( '' === get_query_var( 'search' ) && empty( $_GET ) && is_post_type_archive(
 
 <?php else : ?>
 	<section>
-		<div class="row align-right">
+		<div class="row gutters between section-heading section-heading--with-space">
+			<?php the_archive_title( '<h1 class="section-heading_title h2 col-8">', '</h1>' ); ?>
 			<?php get_template_part( 'template-parts/component', 'archive-search' ); ?>
-		</div>
-		<div class="row align-left">
-			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
 		</div>
 
 		<hr>
--- a/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
+++ b/wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
@@ -187,9 +187,11 @@ if ( '' === get_query_var( 'search' ) && empty( $_GET ) && is_post_type_archive(
 
 <?php else : ?>
 	<section>
-		<div class="row align-middle between section-heading section-heading--with-space">
-			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
+		<div class="row align-right">
 			<?php get_template_part( 'template-parts/component', 'archive-search' ); ?>
+		</div>
+		<div class="row align-left">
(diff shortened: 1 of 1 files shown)


jonathanbossenger:
Fixes #1808

jonathanbossenger:
It looks like this also fixes #1813

jonathanbossenger:
This looks good to me. I tested the lesson plan archive by topic, as well as the lesson plan archive with a search, and it all works as expected.

We might want to at some point in the future consider making that section of the lesson plan archive look more like the same section in the tutorial archive, but that's a future problem to tackle.

jonathanbossenger:
@adamwoodnz when you have a moment, would you mind merging and deploying this change?

jonathanbossenger (approved):

wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
```diff
@@ -187,10 +187,12 @@
 
 <?php else : ?>
 	<section>
-		<div class="row align-middle between section-heading section-heading--with-space">
```
adamwoodnz, line 190: I think we could change this layout to use a `col` class to limit the width of the title, something like:

```
		<div class="row gutters between section-heading section-heading--with-space">
			<?php the_archive_title( '<h1 class="section-heading_title h2 col-8">', '</h1>' ); ?>
			<?php get_template_part( 'template-parts/component', 'archive-search' ); ?>
		</div>
```

This is the approach taken with the section intro pattern:

Screenshot 2023-08-30 at 2 37 54 PM

Because this is a shared pattern there is more complexity to the change though unfortunately. This style rule is being applied to override the `align-middle` class, and is affecting the layout when changed to the above, when the class should actually just be removed:

```
@media only screen and (min-width: 768px) {
    .section-heading.row {
        align-items:baseline;
    }
}
``` 

In my tests that seems to be safe. I think we should try to remove this class from the template headings throughout the site, and then remove this style rule too.

adamwoodnz (commented):
Thanks for contributing this fix @cynthianorman 

This pattern of title or paragraph on the left with search on the right is very common throughout the site, so I'd like to see if we can maintain that if possible. I've left a comment inline regarding a solution using column classes, which is used in other places.

Let me know if you're comfortable trying that, or I'm happy to assist.

wp-content/themes/pub/wporg-learn-2020/archive-lesson-plan.php
```diff
@@ -187,10 +187,12 @@
...
-			<?php the_archive_title( '<h1 class="section-heading_title h2">', '</h1>' ); ?>
+		<div class="row align-right">
 			<?php get_template_part( 'template-parts/component', 'archive-search' ); ?>
```
adamwoodnz, line 192: Moving the search before the title changes the mobile view, currently search is second:

cynthianorman:
hi @adamwoodnz thanks so much for your guidance! I've made the suggested changes and pushed them already. Please let me know if I need to change anything else.
This is what we can expect now for long titles - https://prnt.sc/2BmZBg3_NfUR

jonathanbossenger:
Thank you for reviewing and picking these items up @adamwoodnz

adamwoodnz:
Nearly there! I've added another commit and opened a new [PR](https://github.com/WordPress/Learn/pull/1834), as I can't push to your branch. Please try it out.

jonathanbossenger:
Thank you for all the hard work so far @cynthianorman and @adamwoodnz.

I'm going to close this PR now in favor of #1834

PR #1871, opened 2023-09-19
Add function_exists check

pkevan:
During early loading of the template, the function `locale_notice` does not exists, so produces a fatal.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/header.php
+++ b/wp-content/themes/pub/wporg-learn-2020/header.php
@@ -84,4 +84,6 @@ $menu_items = array(
 			</div><!-- .site-branding -->
 		</header><!-- #masthead -->
 		
-		<?php if ( function_exists( 'locale_notice' ) ) { locale_notice(); } ?>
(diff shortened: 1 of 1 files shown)


jonathanbossenger:
Thanks for creating this PR @pkevan I'll make a point to test it on a local dev environment, but it looks like a simple enough fix for the problem.

jonathanbossenger:
@adamwoodnz I'm just not finding time to review this, would you be able to?

adamwoodnz (approved):
Code change LGTM. Interestingly I don't see the locale notice locally, in sandbox or prod before or after this change... probably worth more investigation?

pkevan:
> Code change LGTM. Interestingly I don't see the locale notice locally, in sandbox or prod before or after this change... probably worth more investigation?

@adamwoodnz it only happens in certain circumstances, see the other issue on the [Neso board](https://github.com/Automattic/neso/issues/75).

jonathanbossenger:
Thanks, folks. I'm going to go ahead and merge this, it seems a straightforward enough fix.

If someone can please deploy this at their next available opportunity, thanks.

PR #1944, opened 2023-10-15
Force resolution of eslint-plugin-jsdoc to support Node 18

### LINKED ISSUE #1943: Can't install JS dependencies on Node 18
See https://github.com/WordPress/gutenberg/issues/55374

Our builds use `@wordpress/scripts` which depends on `eslint-plugin-jsdoc`, and the version currently trying to be installed does not support Node 18:

```
🚚  Fetching packages...
error eslint-plugin-jsdoc@36.1.1: The engine "node" is incompatible with this module. Expected version "^12 || ^14 || ^16". Got "18.16.0"
error Found incompatible module.
```

## Description
The outcome of this is that the project cannot be built and deployed on infrastructure used by Meta engineers (sandboxes).

1. Checkout the project
2. Ensure you are using Node 18.*
3. Run `yarn`

## Expected behaviour
Dependencies are installed 

## Actual behaviour
Dependencies are not installed, see error message above

## Environment information
- Node 18 on MacOS and Linux

adamwoodnz:
Fixes #1943 

@wordpress/scripts depends on `eslint-plugin-jsdoc@36.1.1` which doesn't install on Node 18.
`eslint-plugin-jsdoc@39.9.1` is the first major version after that with support.

### Testing
1. Ensure you are using Node 18.*
2. Run `yarn`
3. Dependencies should install
4. Run `yarn workspaces run build`
5. Projects should compile
6. Run `wp-env start`
7. Smoke test the site

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,5 +33,8 @@
 	"main": "index.js",
 	"devDependencies": {
 		"@wordpress/scripts": "^19.2.2"
+	},
+	"resolutions": {
+		"eslint-plugin-jsdoc": "^39.9.1"
 	}
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1304,14 +1304,14 @@
   resolved "https://registry.yarnpkg.com/@emotion/weak-memoize/-/weak-memoize-0.3.0.tgz#ea89004119dc42db2e1dba0f97d553f7372f6fcb"
   integrity sha512-AHPmaAx+RYfZz0eYu6Gviiagpmiyw98ySSlQvCUhVGDRtDFe4DBS0x1bSjdF3gqUDYOczB+yYvBTtEylYSdRhg==
 
-"@es-joy/jsdoccomment@0.10.8":
-  version "0.10.8"
-  resolved "https://registry.yarnpkg.com/@es-joy/jsdoccomment/-/jsdoccomment-0.10.8.tgz#b3152887e25246410ed4ea569a55926ec13b2b05"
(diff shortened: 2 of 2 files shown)


jonathanbossenger:
@adamwoodnz if it helps I tested this out locally, and it all seems to work as expected.

package.json
```diff
@@ -33,5 +33,8 @@
 	"main": "index.js",
 	"devDependencies": {
 		"@wordpress/scripts": "^19.2.2"
+	},
+	"resolutions": {
```
outdoor2kode, line 37: 👍

outdoor2kode (approved):
LGTM. Thanks for sorting this out. Changing only the `eslint-plugin-jsdoc` version seems pretty safe, and I haven't noticed any issues while testing the sites. However, there's an error `Uncaught ReferenceError: Prism is not defined` popping up on every page, which isn't happening on the production. But I kind of remember this error has been there for quite a while..? something related to CORS or local env, I'm not entirely sure.

ps. the build workflow and and sync script changes look nice as well 👍

adamwoodnz:
Thanks for testing @renintw and @jonathanbossenger 

Since creating this I've had conversations with Gutenberg folks and found out that the latest @wordpress/scripts does support Node 18. I made the mistake of only updating the top level package.json and missed the workspace dependencies when I was testing. So we could fix this by upgrading that package across all the workspaces, however I think moving to building and pushing to a build branch using a GitHub action like we do for most projects is the better option. As Ren mentioned [I've added the workflow](https://github.com/WordPress/Learn/compare/a8cdd6c23ac1c85f146d78f592f4dcd9653ea144...trunk) and I have a patch ready to update the sandbox sync script. This has been reviewed so I'll look to migrate asap.

PR #2120, opened 2023-12-19
Adding //lesson automation

jonathanbossenger:
Add the //lesson automation, which allows contributors and faculty to automate the process of adding the Lesson Creation checklist to new Lesson issues

### DIFF
--- a/.github/workflows/comment-with-checklist.yml
+++ b/.github/workflows/comment-with-checklist.yml
@@ -40,7 +40,7 @@ jobs:
             - [ ] Request a review from the team
             - [ ] [Publish content](https://make.wordpress.org/training/handbook/courses/publishing/)
       - uses: peter-evans/create-or-update-comment@v3
-        if: contains(inputs.content-type, 'lesson-plan')  
+        if: contains(inputs.content-type, 'teach')  
         with:
           issue-number: ${{ github.event.issue.number }}
           body: |
--- a/.github/workflows/comment-with-checklist.yml
+++ b/.github/workflows/comment-with-checklist.yml
@@ -72,5 +72,19 @@ jobs:
             - [ ] Assign [a Co-host](https://make.wordpress.org/training/handbook/online-workshops/co-hosting-an-online-workshop/) to event (Remove the "Needs Co-host" tag from issue once assigned)
             - [ ] Hold Online Workshop
             - [ ] Complete [After an Online Workshop](https://make.wordpress.org/training/handbook/online-workshops/after-an-online-workshop/) next steps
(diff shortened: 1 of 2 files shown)


jonathanbossenger:
Note that we can only merge this once the new [Lesson Creation guidelines](https://docs.google.com/document/d/1twvDGmVZsyjBZbi9abiqoA_a2G7O_WNis_iGL2Sccg0/edit?usp=sharing) have been published, as we will have to add links to the relevant new handbook pages.

kaitohm:
Actually, in its current state, the `//lesson` command is going to trigger both the Lesson check list and the Lesson Plan check list... The last English (non-translated) Lesson Plan published on Learn seems to be in July 2022. As Faculty haven't created a new lesson plan in 18 months, and as that check list is most likely not going to be used, what do you think about removing the lesson plan trigger and check list all together?

Also noting https://github.com/WordPress/Learn/blob/trunk/.github/ISSUE_TEMPLATE/02-content-development-faculty.md will need to be updated as part of these changes.

jonathanbossenger:
Out of interest, when you set up the automations, did you try using [Operators](https://docs.github.com/en/actions/learn-github-actions/expressions#operators) instead of [functions](https://docs.github.com/en/actions/learn-github-actions/expressions#functions) to test the `if` condition?

Looking at the documentation

```
if: ${{ inputs.content-type == 'lesson' }}
```

and 

```
if: ${{ inputs.content-type == 'lesson-plan' }}
```

Should perform an "is equal to" check, and therefore only return true if it's an exact match.

jonathanbossenger:
> Also noting https://github.com/WordPress/Learn/blob/trunk/.github/ISSUE_TEMPLATE/02-content-development-faculty.md will need to be updated as part of these changes.

Good point, I will create a PR for those in the new year, and then we can review and merge all three together.

kaitohm (approved):
I've modified the lesson plan command so that it doesn't conflict with the new lesson command.

kaitohm:
> Out of interest, when you set up the automations, did you try using [Operators](https://docs.github.com/en/actions/learn-github-actions/expressions#operators) instead of [functions](https://docs.github.com/en/actions/learn-github-actions/expressions#functions) to test the if condition?

Nope, no particular reason. I'm brand new at this and mostly copying things I'm seeing other folks do 😅 

On that note, the `==` operator looks like it would work for `comment-with-checklist.yml` where we're comparing an input against the keywords.

However, for `content-checklist-from-issue.yml`, the function `contains(github.event.issue.body, '//lesson')` would still pickup `//lesson-plan` if it existed, so I feel like the two keywords should still be different.

jonathanbossenger:
@bsanevans checking in here, is this ready to be merged?

Issue #2111, opened 2023-12-16
Script to count unique learners


kaitohm:
## Goal
We're looking to track the increase in learners we expect will occur when we launch Learning Pathways next year. To do this, we need to see the number of unique course subscribers (called "students" in Sensei) on Learn.

## Problem
Theoretically, this should be shown on https://learn.wordpress.org/wp-admin/admin.php?page=sensei_reports . However, due to the way bbPress is integrated with make.wordpress.org, we are only seeing a small subset there. This can be confirmed by going to https://learn.wordpress.org/wp-admin/admin.php?page=sensei_reports&course_id=8657&view=user, which shows we have at least 8,900 learners subscribed to that course alone.

## Solution
@m1r0 has written us a script that will count the number of unique learners on Learn and display this on [the Students report screen](https://learn.wordpress.org/wp-admin/admin.php?page=sensei_reports). It also includes a date filter which will only count students who started a course in that time period.

```
add_action( 'sensei_reports_overview_before_top_filters', 'wporg_learn_add_student_count_to_reports' );
function wporg_learn_add_student_count_to_reports( $type ) {
    if ( 'users' !== $type ) {
        return; // Only show the count on the students report screen.
    }
 
    $from_date = DateTime::createFromFormat( 'Y-m-d', $_GET['from_date'] ?? '', new DateTimeZone( 'UTC' ) );
    $to_date   = DateTime::createFromFormat( 'Y-m-d', $_GET['to_date'] ?? '', new DateTimeZone( 'UTC' ) );
 
    global $wpdb;
    $query = "SELECT COUNT(DISTINCT user_id)
FROM $wpdb->comments
WHERE comment_type = 'sensei_course_status'";
 
    if ( $from_date ) {
        $from_date->setTime( 0, 0, 0 );
 
        $query .= " AND comment_date_gmt >= '" . $from_date->format( 'Y-m-d H:i:s' ) . "'";
    }
 
    if ( $to_date ) {
        $to_date->setTime( 23, 59, 59 );
 
        $query .= " AND comment_date_gmt <= '" . $to_date->format( 'Y-m-d H:i:s' ) . "'";
    }
 
    $student_count = $wpdb->get_var( $query );
 
    ?>
    <div class="actions bulkactions">
        <label><?php esc_html_e( 'Total number of students', 'wporg-learn' ); ?></label>
        <input
            class="sensei-date-picker"
            name="from_date"
            type="text"
            autocomplete="off"
[...]

PR #2172, opened 2024-01-23
Add script to count unique learnersin Sensei

kaitohm:
Adds the script provided by @m1r0 in #2111 to wporg-learn plugin

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/sensei.php
+++ b/wp-content/plugins/wporg-learn/inc/sensei.php
@@ -177,8 +177,8 @@ function wporg_learn_add_student_count_to_reports( $type ) {
 
 	global $wpdb;
 	$query = "SELECT COUNT(DISTINCT user_id)
-FROM $wpdb->comments
-WHERE comment_type = 'sensei_course_status'";
+	FROM $wpdb->comments
+	WHERE comment_type = 'sensei_course_status'";
 
 	if ( $from_date ) {
 		$from_date->setTime( 0, 0, 0 );
--- a/wp-content/plugins/wporg-learn/inc/sensei.php
+++ b/wp-content/plugins/wporg-learn/inc/sensei.php
@@ -182,13 +182,13 @@ WHERE comment_type = 'sensei_course_status'";
 
 	if ( $from_date ) {
 		$from_date->setTime( 0, 0, 0 );
- 
+
 		$query .= " AND comment_date_gmt >= '" . $from_date->format( 'Y-m-d H:i:s' ) . "'";
 	}
 
 	if ( $to_date ) {
 		$to_date->setTime( 23, 59, 59 );
- 
+
 		$query .= " AND comment_date_gmt <= '" . $to_date->format( 'Y-m-d H:i:s' ) . "'";
 	}
 
@@ -202,7 +202,7 @@ WHERE comment_type = 'sensei_course_status'";
 			name="from_date"
 			type="text"
 			autocomplete="off"
-			placeholder="<?php echo esc_attr( __( 'From Date', 'wporg-lms' ) ); ?>"
+			placeholder="<?php echo esc_attr( __( 'From Date', 'wporg-learn' ) ); ?>"
 			value="<?php echo esc_attr( $from_date ? $from_date->format( 'Y-m-d' ) : '' ); ?>"
 		/>
 		<input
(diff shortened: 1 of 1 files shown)


kaitohm:
@jonathanbossenger This pull request is ready for a review! 🥳 

I'll note, I suppressed a lint error in line 195:
```
$student_count = $wpdb->get_var( $query ); //phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
```
Jonathan and I discussed that this might be permissible in this case.

wp-content/plugins/wporg-learn/inc/sensei.php
```diff
@@ -163,3 +163,60 @@ function wporg_fix_learning_mode_header_space() {
...
+	if ( 'users' !== $type ) {
+		return; // Only show the count on the students report screen.
+	}
+
+	$from_date = \DateTime::createFromFormat( 'Y-m-d', $_GET['from_date'] ?? '', new \DateTimeZone( 'UTC' ) );
+	$to_date   = \DateTime::createFromFormat( 'Y-m-d', $_GET['to_date'] ?? '', new \DateTimeZone( 'UTC' ) );
+
+	global $wpdb;
+	$query = "SELECT COUNT(DISTINCT user_id)
+FROM $wpdb->comments
```
jonathanbossenger: @bsanevans could I ask you to indent the second and third lines of the SQL query under the first line, just to make it cleaner.

wp-content/plugins/wporg-learn/inc/sensei.php
```diff
@@ -163,3 +163,60 @@ function wporg_fix_learning_mode_header_space() {
...
+		$query .= " AND comment_date_gmt >= '" . $from_date->format( 'Y-m-d H:i:s' ) . "'";
+	}
+
+	if ( $to_date ) {
+		$to_date->setTime( 23, 59, 59 );
+
+		$query .= " AND comment_date_gmt <= '" . $to_date->format( 'Y-m-d H:i:s' ) . "'";
+	}
+
+	$student_count = $wpdb->get_var( $query ); //phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
```
jonathanbossenger, line 195: @pkevan Ben and I discussed this, and we feel ignoring the PHPCS rule here is ok, but we'd appreciate your opinion. I personally can't see how this could be attacked via SQL injection, but maybe I'm missing something.

jonathanbossenger (approved):
This looks good to me, just waiting on feedback on that SQL line

jonathanbossenger:
@pkevan when you have a moment, @bsanevans and I would appreciate a second pair of eyes on this PR, thanks.

wp-content/plugins/wporg-learn/inc/sensei.php
pkevan: While it's true that the current code isn't vulnerable, that doesn't mean in the future it wouldn't be, so it's preferable to built in elements which would force it to not be.

For example, the `$from_date` variable assignment logic could be changed to grab from the `$_GET` variable directly or the appended `$query` concatenation logic changed in a way which did make it.

By adding some prepare statements, it would help avoid those scenarios, particularly when dealing with user input.

wp-content/plugins/wporg-learn/inc/sensei.php
kaitohm:
```suggestion
	FROM $wpdb->comments
```

wp-content/plugins/wporg-learn/inc/sensei.php
```diff
@@ -163,3 +163,60 @@ function wporg_fix_learning_mode_header_space() {
...
+		return; // Only show the count on the students report screen.
+	}
+
+	$from_date = \DateTime::createFromFormat( 'Y-m-d', $_GET['from_date'] ?? '', new \DateTimeZone( 'UTC' ) );
+	$to_date   = \DateTime::createFromFormat( 'Y-m-d', $_GET['to_date'] ?? '', new \DateTimeZone( 'UTC' ) );
+
+	global $wpdb;
+	$query = "SELECT COUNT(DISTINCT user_id)
+FROM $wpdb->comments
+WHERE comment_type = 'sensei_course_status'";
```
kaitohm:
```suggestion
	WHERE comment_type = 'sensei_course_status'";
```

wp-content/plugins/wporg-learn/inc/sensei.php
jonathanbossenger: @pkevan Thanks, that's a fair point, I will work on adding sanitization to the date variables, and change the SQL concatenation to use a prepare statement.

jonathanbossenger:
@bsanevans I have merged #2178, so I am going to close this out.

PR #2178, opened 2024-01-26
Refactoring Add script to count unique learners in Sensei

### LINKED ISSUE #2111: Script to count unique learners

jonathanbossenger:
Update to #2172
Fixes #2111 
Add sanitization for query strings
Use wpdb->prepare for SQL query

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/sensei.php
+++ b/wp-content/plugins/wporg-learn/inc/sensei.php
@@ -183,7 +183,7 @@ function wporg_learn_get_date( $query_var ) {
  */
 function wporg_learn_get_student_count( $from_date, $to_date ) {
 
-	if ( ! $from_date && ! $to_date ) {
+	if ( ! $from_date || ! $to_date ) {
 		return 0;
 	}
 
--- a/wp-content/plugins/wporg-learn/inc/sensei.php
+++ b/wp-content/plugins/wporg-learn/inc/sensei.php
@@ -163,3 +163,81 @@ function wporg_fix_learning_mode_header_space() {
 	wp_add_inline_style( 'learning-mode-header-fix', $custom_styles );
 }
 add_action( 'sensei_course_learning_mode_load_theme', __NAMESPACE__ . '\wporg_fix_learning_mode_header_space' );
+
+/**
+ * Format a date query var into a DateTime object.
+ */
+function wporg_learn_get_date( $query_var ) {
+	$date = sanitize_text_field( $_GET[ $query_var ] ?? '' );
+
+	return \DateTime::createFromFormat( 'Y-m-d', $date ?? '', new \DateTimeZone( 'UTC' ) );
+}
+
+/**
+ * Get the number of unique learners between two dates.
+ *
+ * @param \DateTime $from_date
+ * @param \DateTime $to_date
+ *
+ * @return int
+ */
+function wporg_learn_get_student_count( $from_date, $to_date ) {
+
+	if ( ! $from_date && ! $to_date ) {
+		return 0;
+	}
+
+	global $wpdb;
+
+	return $wpdb->get_var(
+		$wpdb->prepare(
+			"SELECT COUNT(DISTINCT user_id) FROM $wpdb->comments 
+			WHERE comment_type = 'sensei_course_status'
+			AND comment_date_gmt >= %s 
+			AND comment_date_gmt <= %s",
+			array(
+				$from_date->format( 'Y-m-d H:i:s' ),
+				$to_date->format( 'Y-m-d H:i:s' ),
+			)
+		)
+	);
+}
+
+/**
+ * Add script to count unique learners
+ */
+function wporg_learn_add_student_count_to_reports( $type ) {
+	if ( 'users' !== $type ) {
+		return; // Only show the count on the students report screen.
+	}
+
+	$from_date = wporg_learn_get_date( 'from_date' );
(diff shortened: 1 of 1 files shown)


jonathanbossenger:
@bsanevans In attempting to refactor #2172 to include sanitizing the query variables and to use $wpdb->prepare for the query, I had to make quite a few changes to the code.

Could I ask you to get  @m1r0 to review this refactored version to confirm it would still work functionally the same as the original code, then I'll get someone else from the meta team to review it.

kaitohm:
I've reached out to @m1r0 .

wp-content/plugins/wporg-learn/inc/sensei.php
```diff
@@ -163,3 +163,81 @@ function wporg_fix_learning_mode_header_space() {
...
+ * Get the number of unique learners between two dates.
+ *
+ * @param \DateTime $from_date
+ * @param \DateTime $to_date
+ *
+ * @return int
+ */
+function wporg_learn_get_student_count( $from_date, $to_date ) {
+
+	if ( ! $from_date && ! $to_date ) {
```
m1r0: I'm getting an error when filling in only one of the dates.

```
Call to a member function format() on bool
```

Do you intend to make it that both dates are required? If that's the case, maybe the logic should be:

```suggestion
	if ( ! $from_date || ! $to_date ) {
```

WDYT?

m1r0 (commented):
Hello, hello! 👋 🙂 

> still work functionally the same as the original code

The only functional difference I'm noticing is that now both dates are needed. If that's intended, I've left some feedback.

Everything else looks solid. 🪨

kaitohm:
It would be great if we didn't need to set both dates. For example, if we could get all data from a specific date until today without needing to set today's date, that would be great.

@jonathanbossenger Is that a change you can implement?

wp-content/plugins/wporg-learn/inc/sensei.php
jonathanbossenger: @m1r0 thanks for the feedback. My focus was on the refactoring to get the php linter to pass, so I wasn't 100% sure myself, happy to make the change.

jonathanbossenger:
Thanks @m1r0 and @bsanevans I've updated the PR.

@pkevan This PR is an update to #2172, adding sanitization to the date variables, and changing the SQL concatenation to use a prepared statement.

Could we ask you to give this a review?

Thanks

kaitohm (approved):

pkevan (approved):
👍 Looks good

Issue #2221, opened 2024-02-15
Tutorial Presenter Application form check box sign not proper alignment.


vipulupadala-multidots:
## Type of feedback
Need to improve the check box sign design for The Tutorial Presenter Application form page. 

## Description
I have reviewed the meta site and during the Tutorial Presenter Application form check box design is not properly displayed. So think we need to improve some design.

## Step-by-step reproduction instructions (optional)
1. First Open the [URL](https://learn.wordpress.org/tutorial-presenter-application/)
2. Select the check box and see design not displayed properly

## Screenshots or screen recording (optional)
Screenshot: https://share.cleanshot.com/fK9BS8FH07DgbNPsMwKD

devmuhib009:
Thanks @vipulupadala-multidots for creating the issue.

unprintedch:
I did try something.

There is couple of question in the PR. I put them back here for easy discussion.

Open questions:
How could we make a good use of sensei css variables?
I use the SVG icon in the CSS, I find it bad... (and the linter too) can someone tell me how to do it properly?
And where to put this "checkbox.scss" it is an override of a vendor, any good practice?

unprintedch:
I just notice that there was already a PR?

ironnysh:
The redesign resolved this issue (tested [this page](https://learn.wordpress.org/online-workshops/apply-to-facilitate/) in Chrome, Safari, and Firefox).

We can close the two related PRs: #2222 and #2359.

cc @jonathanbossenger

PR #2222, opened 2024-02-15
#2221 number issues is resolved

vipulupadala-multidots:


ironnysh:
The redesign resolved #2221. We can close this PR.

cc @jonathanbossenger

PR #2391, opened 2024-04-23
Update dependencies to build on Node 20

### LINKED ISSUE #2390: Upgrade to Node 20
This enables:

1. Playground with blueprint files (see ##2343) 
2. Upgrading wp-scripts for experimental modules support, so that we can use the Interactivity API in the new theme

adamwoodnz:
Closes #2390


package.json
```diff
@@ -6,7 +6,9 @@
 	"license": "GPL-2.0-or-later",
 	"private": true,
 	"dependencies": {
-		"@wordpress/env": "4.1.3"
+		"@wordpress/env": "9.3.0",
+		"@wordpress/scripts": "27.2.0",
+		"@wordpress/stylelint-config": "21.34.0"
```
adamwoodnz, line 11: Dependencies updated to match the [Pattern Directory theme](https://github.com/WordPress/pattern-directory/blob/d895a721a9390f35d67982814e4a4030ad5a88f1/package.json#L8)

.github/workflows/linters.yml
```diff
@@ -13,15 +13,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
```
adamwoodnz, line 16: Resolves warnings about deprecated Node 16 actions

PR #2398, opened 2024-04-23
Update linting to use repo-tools configs

### LINKED ISSUE #2399: Use repo-tools configs for linting
We already depend on `wporg/wporg-repo-tools` and use it for builds. For consistency of code style with our others repos we should use the lint configs too, in particular prettier and eslint. As we add more blocks in the new block theme this will become more important.

The currently installed stylelint config is already consistent with the repo-tools version and there is a custom phpcs config which should not be changed.

adamwoodnz:
Closes #2399 

Adds the prettier and eslint configs via the setup:tools script from wporg-repo-tools, and fixes lint errors.


ryelle:
In other projects, [we've added the wporg-repo-tools configs to .gitignore](https://github.com/WordPress/wporg-main-2022/blob/618fd76e47510e15fa0184ccd03d52562c1814f0/.gitignore#L14-L18), and if any overrides are needed, add [files that inherit from them](https://github.com/WordPress/wporg-main-2022/blob/trunk/phpcs.xml). Would that work here?

.github/workflows/linters.yml
```diff
@@ -20,6 +20,14 @@ jobs:
         mkdir wp-content/themes/pub/wporg
         svn export https://meta.svn.wordpress.org/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/css wp-content/themes/pub/wporg/css
 
+    - name: composer install
```
adamwoodnz, line 23: I couldn't work out why moving this step up (so that repo-tools is available) would cause lint to fail on the locale-switcher. The failures weren't consistent with the tracked code. Then I discovered that locale-switcher is installed via composer from [wordpress-meta/pub](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/locale-switcher) and overwrites the tracked copy. I've now removed this code from this repo and stopped linting it.

adamwoodnz:
> In other projects, [we've added the wporg-repo-tools configs to .gitignore](https://github.com/WordPress/wporg-main-2022/blob/618fd76e47510e15fa0184ccd03d52562c1814f0/.gitignore#L14-L18), and if any overrides are needed, add [files that inherit from them](https://github.com/WordPress/wporg-main-2022/blob/trunk/phpcs.xml). Would that work here?

I've got the css and js configs working now, but the phpcs config seems pretty far removed from what's in repo-tools, as when I tried to extend it with these custom rules there were a lot of failures. So I've left it alone for now.

.github/workflows/linters.yml
adamwoodnz: Reverted this removal as this does seem to be the source, and I'm instead resetting git to remove the changes caused by this install.

PR #2504, opened 2024-06-05
Add query filters to Course and Lesson archives

### LINKED ISSUE #2396: Courses landing page
Complete the Learning Pathway taxonomy template:

- [x]  Reuse Courses Grid block with query inherited from page and page size of 12 with pagination
- [x]  Add search
- [ ]  Add query filters

### LINKED ISSUE #2417: Lessons landing page
- [x] Reuse [Lessons cards grid](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2024/patterns/front-page-content.php#L101) markup, with the query inherited from the page, and a page size of 12, with pagination.
- [ ] Add query filters
- [x] Add search

adamwoodnz:
Closes #2396 
Closes #2417 

Adds query filters to the Course and Lesson archives. The implementation is mostly copied from Showcase. The main difference here is that we need to filter Lessons by Language, which is post meta.

I've added this Language filter on the Course archive too, as I think it's more likely that it will be set there than on individual courses, and [noted this on Figma](https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp?node-id=3473-44005&m=dev#830155785).

I haven't added the Lesson Duration filter yet as I'm not sure how it should work, and I've created a new [issue](https://github.com/WordPress/Learn/issues/2510).

#### Courses

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_courses__language%5B%5D=en_NZ(Desktop) | localhost_8888_courses__language%5B%5D=en_NZ(iPad) | localhost_8888_courses__language%5B%5D=en_NZ(Samsung Galaxy S20 Ultra) |

#### Lessons

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_lessons__wporg_lesson_level%5B%5D=beginner(Desktop) | localhost_8888_lessons__wporg_lesson_level%5B%5D=beginner(iPad) | localhost_8888_lessons__wporg_lesson_level%5B%5D=beginner(Samsung Galaxy S20 Ultra) |

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -6,6 +6,7 @@
 namespace WordPressdotorg\Theme\Learn_2024\Block_Config;
 
 add_filter( 'wporg_query_filter_options_level', __NAMESPACE__ . '\get_course_level_options' );
+add_filter( 'wporg_query_filter_options_topic', __NAMESPACE__ . '\get_course_topic_options' );
 
 /**
  * Get the list of levels for the course filters.
@@ -47,3 +48,42 @@ function get_course_level_options( $options ) {
 	);
 }
 
+/**
+ * Get the list of topics for the course filters.
+ *
+ * @param array $options The options for this filter.
+ * @return array New list of topic options.
+ */
+function get_course_topic_options( $options ) {
+	global $wp_query;
+	// Get top 20 topics ordered by count, then sort them alphabetically.
+	$topics = get_terms(
+		array(
+			'taxonomy' => 'topic',
+			'orderby' => 'count',
+			'order' => 'DESC',
+			'number' => 20,
+		)
+	);
+	usort(
+		$topics,
+		function ( $a, $b ) {
+			return strcmp( strtolower( $a->name ), strtolower( $b->name ) );
+		}
+	);
+	$selected = isset( $wp_query->query['wporg_workshop_topic'] ) ? (array) $wp_query->query['wporg_workshop_topic'] : array();
+	$count = count( $selected );
+	$label = sprintf(
+		/* translators: The dropdown label for filtering, %s is the selected term count. */
+		_n( 'Topics <span>%s</span>', 'Topics <span>%s</span>', $count, 'wporg-learn' ),
+		$count
+	);
+	return array(
+		'label' => $label,
+		'title' => __( 'Topics', 'wporg-learn' ),
+		'key' => 'wporg_workshop_topic',
+		'action' => home_url( '/courses/' ),
+		'options' => array_combine( wp_list_pluck( $topics, 'slug' ), wp_list_pluck( $topics, 'name' ) ),
+		'selected' => $selected,
+	);
+}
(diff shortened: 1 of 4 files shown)


wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Modify the query by adding meta query for language if set.
+ *
+ * @param WP_Query $query The query object.
+ */
+function modify_query( $query ) {
+	// Ensure this code runs only for the main query on archive pages
+	if ( ! is_admin() && $query->is_main_query() && $query->is_archive() ) {
+		$language = $_GET['language'];
```
outdoor2kode: It shows this notice on my side, but doesn't when a language is selected.
Screenshot 2024-06-07 at 02 11 39
But after clearing the selection, this screen shows up.
image

I tried adding a check like this and both notices are gone.
```
if ( ! isset( $_GET['language'] ) ) {
    return;
}
```

outdoor2kode (commented):
The query filters added to these two pages work as expected 👍 
Just saw this notice on my side, it doesn't happen for you? (Feel that this might be related to the environment as well, as [It looks like](https://github.com/WordPress/Learn/pull/2504/files#diff-9213c818b0a0a9c6a66fa67f0b071520f602f0565d53c4be6d0499a8178522feR146) the language should have a default value)

ryelle:
Is it expected that there are only English things on production? I'm testing on my sandbox on https://learn.wordpress.org/lessons/ and this is what I see:

Screenshot 2024-06-06 at 2 36 34 PM

I also can't view any filtered lessons, and I just want to check whether that's a content issue, something you're still working on, or a bug. For example, on https://learn.wordpress.org/lessons/?wporg_workshop_topic%5B%5D=using-wordpress (I tried a few topics)

Screenshot 2024-06-06 at 2 45 31 PM

The Level dropdown has an "Any" option, but it's in the middle of the others. However I don't know if you need it, since this is a multi-select dropdown. or you could switch it to a single select with `<!-- wp:wporg/query-filter {"key":"level","multiple":false} /-->`

Screenshot 2024-06-06 at 2 46 37 PM

I have the same empty result set on this filter ^ too.

The filters do appear to work on courses.

adamwoodnz:
Thanks for testing! So meta fields are a bit odd in that [even if there is a default value set when the field is registered](https://github.com/WordPress/Learn/blob/trunk/wp-content/plugins/wporg-learn/inc/post-meta.php#L181), that doesn't mean the API returns that value if nothing is set, hence why I have to manually inject the option, and handle matching on an empty value. Seems like I shouldn't do this if there are no other languages defined, then the filter won't be shown.

There isn't a lot of production data for any of these filters so yes I think that's why you're seeing empty data sets. I think I need to fetch options based on the current post type so that a filter isn't shown if there isn't any data.

adamwoodnz:
> The Level dropdown has an "Any" option, but it's in the middle of the others. However I don't know if you need it, since this is a multi-select dropdown. or you could switch it to a single select with `<!-- wp:wporg/query-filter {"key":"level","multiple":false} /-->`

I'll have a look at the `Any` term, agree it doesn't seem necessary.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
adamwoodnz: Fixed in [39ca990](https://github.com/WordPress/Learn/pull/2504/commits/39ca99020d07f8c6e7c093859ba5398db7ee78c2)

adamwoodnz:
> The Level dropdown has an "Any" option, but it's in the middle of the others. However I don't know if you need it, since this is a multi-select dropdown. or you could switch it to a single select

I've taken a look and `Any` is an existing term so it needs to be handled for now. I've got it hoisted to the top of the list, but I think we can look at cleaning up the data as you're right, it doesn't make sense with the new filtering.

PR #2540, opened 2024-06-18
Add single Lesson Plan page

### LINKED ISSUE #2448: Single Lesson Plan page

outdoor2kode:
Closes #2448

Add a single lesson page template.
The design of "View Slides" is pending [discussion](https://github.com/WordPress/Learn/issues/2448#issuecomment-2185013762).

## Screenshots

Screenshot 2024-06-24 at 01 12 18
image

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/patterns/sidebar-meta-info.php
+++ b/wp-content/themes/pub/wporg-learn-2024/patterns/sidebar-meta-info.php
@@ -5,14 +5,16 @@
  * Inserter: no
  */
 
- use function WPOrg_Learn\Sensei\{get_my_courses_page_url};
+use function WPOrg_Learn\Sensei\{get_my_courses_page_url};
+$current_post = get_post();
+$current_post_type = get_post_type();
 
 ?>
 
 <!-- wp:group {"align":"full","className":"wporg-learn-sidebar-meta-info","layout":{"type":"constrained"}} -->
 <div class="wp-block-group alignfull wporg-learn-sidebar-meta-info">
 
-	<?php if ( 'course' === get_post_type() ) : ?>
+	<?php if ( 'course' === $current_post_type ) : ?>
 		<?php if ( Sensei_Course::is_user_enrolled( get_the_ID() ) ) : ?>
 		<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|blueberry-1"}}},"typography":{"fontStyle":"normal","fontWeight":"400","lineHeight":26px}},"textColor":"blueberry-1","fontSize":"normal","fontFamily":"inter","className":""wporg-learn-sidebar-all-courses"} -->
 		<p class="has-blueberry-1-color has-text-color has-link-color has-inter-font-family has-normal-font-size wporg-learn-sidebar-all-courses" style="font-style:normal;font-weight:400;line-height:26px">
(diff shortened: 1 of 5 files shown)


wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
adamwoodnz, line 1: This isn't only Sensei things anymore so we should rename it to be more generic, maybe `sidebar-meta`?

wp-content/themes/pub/wporg-learn-2024/patterns/sidebar-meta-info.php
```diff
@@ -5,37 +5,81 @@
...
-	<?php if ( Sensei_Course::is_user_enrolled( get_the_ID() ) ) : ?>
-	<!-- wp:paragraph {"style":{"elements":{"link":{"color":{"text":"var:preset|color|blueberry-1"}}},"typography":{"fontStyle":"normal","fontWeight":"400","lineHeight":26px}},"textColor":"blueberry-1","fontSize":"normal","fontFamily":"inter","className":""wporg-learn-sidebar-all-courses"} -->
-	<p class="has-blueberry-1-color has-text-color has-link-color has-inter-font-family has-normal-font-size wporg-learn-sidebar-all-courses" style="font-style:normal;font-weight:400;line-height:26px">
-		<a href="<?php echo esc_url( get_my_courses_page_url() ); ?>">
-			<?php esc_html_e( 'All My Courses', 'wporg-learn' ); ?>
-		</a>
-	</p>
-	<!-- /wp:paragraph -->
-	<?php endif; ?>
+	<?php if ( 'course' === $current_post_type ) : ?>
```
adamwoodnz, line 17: I think rather than branching we should create a new pattern for each page sidebar; `sidebar-course.php`, `sidebar-lesson-plan.php`. It will make it easier to find the code for each, and simplify the imports, remove logic, etc.

wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
```diff
@@ -90,6 +90,48 @@ function render( $attributes, $content, $block ) {
...
 			),
 		);
+	} else if ( 'lesson-plan' === $block->context['postType'] ) {
+		$lesson_plan_id = $block->context['postId'];
+
+		$duration = get_post_taxonomy_terms_links( $lesson_plan_id, 'duration' );
+		$audience = get_post_taxonomy_terms_links( $lesson_plan_id, 'audience' );
+		$level = get_post_taxonomy_terms_links( $lesson_plan_id, 'level' );
+		$instruction_type = get_post_taxonomy_terms_links( $lesson_plan_id, 'instruction_type' );
+		$wporg_wp_version = get_post_taxonomy_terms_links( $lesson_plan_id, 'wporg_wp_version' );
```
adamwoodnz, line 100: These links aren't the same as the existing links.

Current theme link (for version 5.8): https://learn.wordpress.org/lesson-plans/?wp_version[]=3524
New theme link (for same version): https://learn.wordpress.org/wporg_wp_version/5-8/

The new theme link takes me to the homepage; it doesn't seem to resolve. Ideally it would take you to the Lesson Plan archive with the version filter applied, but as we don't have this and version isn't something that's being used anywhere else, I think we should remove the link and possibly even the field itself.

wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
```diff
@@ -90,6 +90,48 @@ function render( $attributes, $content, $block ) {
 				'key'   => 'last-updated',
 			),
 		);
+	} else if ( 'lesson-plan' === $block->context['postType'] ) {
+		$lesson_plan_id = $block->context['postId'];
+
+		$duration = get_post_taxonomy_terms_links( $lesson_plan_id, 'duration' );
+		$audience = get_post_taxonomy_terms_links( $lesson_plan_id, 'audience' );
+		$level = get_post_taxonomy_terms_links( $lesson_plan_id, 'level' );
```
adamwoodnz, line 98: These links are in this format https://learn.wordpress.org/level/beginner/, ie. they are going to the archive of all content in that taxonomy. I think they should be https://learn.wordpress.org/lesson-plans/?wporg_lesson_level=beginner so that the user lands on the filtered Lesson Plan archive.

adamwoodnz: Same goes for duration, audience and type. My expectation would be that I'm filtering only Lesson Plans by that field, rather than all content types.

Thoughts @fcoveram ?

wp-content/themes/pub/wporg-learn-2024/patterns/sidebar-meta-info.php
```diff
@@ -5,37 +5,81 @@
...
+		<div class="wp-block-buttons">
+			<!-- wp:button {"textAlign":"center","width":100,"style":{"border":{"radius":"2px"},"spacing":{"padding":{"left":"13px","right":"13px","top":"16px","bottom":"16px"}},"typography":{"lineHeight":0,"fontStyle":"normal","fontWeight":"400"}},"className":"aligncenter is-style-fill","fontSize":"normal","fontFamily":"inter"} -->
+			<div class="wp-block-button has-custom-width wp-block-button__width-100 has-custom-font-size aligncenter is-style-fill has-inter-font-family has-normal-font-size" style="font-style:normal;font-weight:400;line-height:0">
+				<a class="wp-block-button__link has-text-align-center wp-element-button" onclick="window.print()" style="border-radius:2px;padding-top:16px;padding-right:13px;padding-bottom:16px;padding-left:13px" target="_blank" rel="noreferrer noopener">
+					<?php esc_html_e( 'Print Lesson Plan', 'wporg-learn' ); ?>
+				</a>				
+			</div>
+			<!-- /wp:button -->
+		</div>
+		<!-- /wp:buttons -->
```
adamwoodnz, line 76: Noting that we don't have a print stylesheet for the new theme yet, so printing these pages results in some pretty mangled header and footer elements. I think in the current theme we've set `display:none` on all of this. Out of scope for this PR obviously. How high priority is printing still @bsanevans @westnz ?

adamwoodnz (commented):
Nicely done! Mostly working for me apart from the sidebar links.

wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
fcoveram: +1 to your suggestion. I would expect the same.

> so that the user lands on the filtered Lesson Plan archive.

Just to double check. You mean [this page](https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp/Learn?node-id=3873-12813&m=dev) but on its "Lesson version" (aka, without the colored hero and pagination at the end), right? If so, then we're aligned.

adamwoodnz: Correct, looks like [this](https://learn.wordpress.org/lesson-plans/?wporg_lesson_level=beginner&new-theme=1)

Screenshot 2024-06-25 at 8 17 37 AM

fcoveram: That's perfect

wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
outdoor2kode: [b3f91e3](https://github.com/WordPress/Learn/pull/2645/commits/b3f91e3cced61f8c972262b4da5017e9552cfbf1)

wp-content/themes/pub/wporg-learn-2024/patterns/sidebar-meta-info.php
outdoor2kode: 65e27f2

wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
outdoor2kode: Yeah, exactly. I directed the link to the term archive page here because there is no filter function and apart from the "wp version", all other terms are accessible. It appears that "wp version" was initially set to not allow viewing the archive page: ```'public' => false.``` If we set it to true, then its archive page can be accessed. It seems fine to allow it should be fine. What do you think?

image

renintw: **Update**

Hold on, I see your comment below, yeah, this field should either be unlinked or just removed, as this term archives not only filtering Lesson Plans.

outdoor2kode: When I set:
```
'public'            => true,
'rewrite'           => false,
```
it achieves the following effect - you can see all lesson plan posts of version 6.1, but it won't reflect in the filter.

image

Do you think this setting is good to make the version archive accessible? Do we want to add a specific filter? 
If we add a filter for duration, it will only be used for lesson plans. The version will be additionally used for tutorials.

Or we just don't link any of them at all as you said and just present it as additional information to the user, especially since it's going to be deprecated. If we do decide to link, adding a filter seems to make more sense.

wp-content/themes/pub/wporg-learn-2024/src/sensei-meta-list/index.php
outdoor2kode: https://github.com/WordPress/Learn/pull/2645/commits/50d75baded222765879664ac0b432f7f55245a9f I currently only allow 'level' to maintain links. Potential further adjustments https://github.com/WordPress/Learn/pull/2540#discussion_r1669531408

image
image

outdoor2kode:
Design updated: https://github.com/WordPress/Learn/pull/2645/commits/4e2052b614882f889a442e0585b69dd055c9ae47

image

outdoor2kode:
This PR is closed in favor of https://github.com/WordPress/Learn/pull/2645

PR #2558, opened 2024-06-20
Add align-wide support

### LINKED ISSUE #2557: Add wide alignment support to the current theme
It would be great if the current theme supported the block editor 'align-wide' theme support. 

I want to be able to use the Playground block in the code editor mode for practical exercises in the developer learning pathway. 

However, I would like to create these exercises on separate pages, where the block can be set to take up the full page width, as this provides a better user experience when using the code editor. 

Adding it this way on a separate page also means I can link to it from the relevant lesson instead of forcing users to view/skip over the block in the lesson itself.

jonathanbossenger:
Fixes #2557

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2020/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2020/functions.php
@@ -17,6 +17,7 @@ function setup() {
 	add_theme_support( 'sensei' );
 
 	add_theme_support( 'responsive-embeds' );
+	add_theme_support( 'align-wide' );
 
 	global $woothemes_sensei;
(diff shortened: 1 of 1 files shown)


jonathanbossenger:
@adamwoodnz or @renintw to the best of my knowledge, adding this support won't have any negative effect on the current theme, but I'll defer to your judgement. 

I want to publish the developer learning pathway next Friday the 28th, so it will be great if this can be merged before then, so I can create and test setting the playground block to wide alignment.

jonathanbossenger:
Closing PR, see linked issue.

Issue #2573, opened 2024-06-26
Fix card content alignment


adamwoodnz:
The footer content is not being pushed to the bottom of the card. Need to fix the flex containers in all archive grid cards, like we did in the My Courses grid cards.

Some content is also overflowing

![Image](https://github.com/WordPress/Learn/assets/1017872/f9a0faa1-d294-47e4-a670-dd3adb49d885)

PR #2578, opened 2024-06-27
Move Course Grid query loop block variation to theme

adamwoodnz:
Follow up to https://github.com/WordPress/Learn/commit/8858ef40aa56cb005b3a24bee36a0d8663256b70, where this variation switched to using the Course Card template part from the theme.

This query loop variation is a presentation component which is tightly coupled to the theme, so it doesn't make sense for it to be in the plugin.

See https://github.com/WordPress/Learn/issues/2573

### Testing

There should be no change to the functionality.
If it's still working, the Featured Courses on the homepage should be showing only featured courses.
If you add the 'Learn Course Grid' block in the editor, it should have a 'Featured only' switch in the block settings in the sidebar.


outdoor2kode (approved):
Works as expected 👍

PR #2586, opened 2024-06-30
Course completed page

### LINKED ISSUE #2528: Course Completed page
Update the template to match the [design](https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp/Learn?node-id=4321-21492&m=dev).

The default success message is shown in the design 'Congratulations on completing this course!'. This message will be customized if the Course is a Learning Pathway, and can differ depending on Level, [see copy](https://docs.google.com/document/d/1eYcuNUjUdPm1CL3P1Sd0J8hZsn2chBzmuZq4hz7RAdw/edit#heading=h.bp9e84t1ujl2).

For example there will be copy for a Beginner level Developer Learning Pathway Course.

We have existing copy per Learning Pathway level stored [here](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2024/functions.php#L163). We can probably add a 'success' item to each array:

```
'developer' => array(
	'beginner' => array(
		'title' => '[TBD Beginner developer title]',
		'description' => '[TBD Beginner developer description]',
		'success' => '...'
	),
	...
),
```

Survey CTA link: https://docs.google.com/forms/d/e/1FAIpQLSf0QMflUedxjta0u5qS4_pl-9aY06BDBXgRn2PoZA1gRvD9jw/viewform

outdoor2kode:
Closes #2528

It seems the copy hasn’t been shared yet. I’ve asked in the docs. 
Let me summarize a bit here to ensure I understand correctly: we are planning to have different copies for different levels of each learning pathway. Is that correct?

## Screenshot

| **Desktop** | **Tablet** | **Mobile** |
|-|-|-|
| Screenshot 2024-07-01 at 05 39 37 | Screenshot 2024-07-01 at 05 39 48 | Screenshot 2024-07-01 at 05 40 13 |


adamwoodnz:
> we are planning to have different copies for different levels of each learning pathway. Is that correct?

Correct 🙂

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -136,6 +136,61 @@ body.sensei {
 			border: solid 1px var(--sensei-secondary-color);
 		}
 	}
+
+	section.wp-block-sensei-lms-course-results {
+		display: flex;
+		flex-direction: column;
+		margin-top: 40px;
+		margin-bottom: 40px;
```
adamwoodnz, line 172: How does `var(--wp--preset--spacing--30)` look? I generally use the closest theme var I can find unless it's a really specific size.

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -136,6 +136,61 @@ body.sensei {
...
+					margin: 0;
+				}
+			}
+		}
+
+		> ul.wp-block-sensei-lms-course-results__lessons {
+			margin: 0;
+
+			> li.wp-block-sensei-lms-course-results__lesson {
+				border: 1px solid #d9d9d9;
```
adamwoodnz:
```suggestion
				border: 1px solid var(--wp--custom--color--border);
```

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -136,6 +136,61 @@ body.sensei {
...
+				margin-bottom: var(--wp--preset--spacing--20);
+				border-radius: var(--wp--custom--button--border--radius);
```
adamwoodnz: This isn't a button so it really shouldn't use this var. The value is very commonly used though so I've considered adding a custom var for these boxes.
```suggestion
				border-radius: var(--wp--custom--box--border--radius);
```

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -136,6 +136,61 @@ body.sensei {
...
+			}
+		}
+
+		> .wp-block-sensei-lms-course-results__grade {
```
adamwoodnz, line 210: This has the 2px border radius too in the design

adamwoodnz:
Just discovered that the prod page slug is `course-complete`, so this template needs to be `page-course-complete.html` to match.

adamwoodnz:
There's a black bar when I view a [completed course](https://learn.wordpress.org/course-complete/?course_id=12041&new-theme=1) in prod

Screenshot 2024-07-01 at 9 07 24 PM

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
outdoor2kode: 8f35a7b

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
outdoor2kode: b172d41

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
outdoor2kode: b172d41

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
outdoor2kode: I personally prefer to align with the design spec. I think that sometimes what I find acceptable, thinking it's just a small difference, might not be viewed the same way by others. Or in some cases, as you mentioned, we can use predefined styles due to the small difference, but in other cases, we use exact pixel values. This already leads to inconsistent styles. I feel that both ways more or less cause some confusion to some extent.
Ultimately, it's best to consult with the design team from the beginning to see if we can use predefined values. Based on previous experience, it's usually acceptable. That said, as I understand it, the design spec should align with our predefined styles? There shouldn't be numbers like 40px because, from what I've seen, the color, font-size, etc., all match up.

outdoor2kode:
> Just discovered that the prod page slug is course-complete, so this template needs to be page-course-complete.html to match.

Updated in 0cbae18. 

We don't have a courses-overview page, right? Although I have it locally, the "Find more courses" link [currently](https://learn.wordpress.org/course-complete/?course_id=12041&new-theme=0) points to `/courses` in production. It should be the same I think.

outdoor2kode:
> There's a black bar when I view a [completed course](https://learn.wordpress.org/course-complete/?course_id=12041&new-theme=1) in prod

Great catch. I can reproduce it locally as well. It's rendered for courses that don't have any modules. ([ref](https://github.com/Automattic/sensei/blob/trunk/includes/blocks/class-sensei-course-results-block.php#L163-L165))

Fixed in b554679

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
adamwoodnz: > There shouldn't be numbers like 40px because, from what I've seen, the color, font-size, etc., all match up.

This is the key bit for me. That's why it has to be very exceptional for me to use a custom value for this kind of spacing. If the difference gets called out in a design review, that's when I'll start the discussion on changing the value in the system.

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
adamwoodnz: Thanks! I'll update the other usages.

adamwoodnz:
> It should be the same I think.

Agreed, should go to `/courses`

adamwoodnz (approved):
LGTM, could probably rebase on trunk and merge. The border radius var can ship later imo.

adamwoodnz:
Needs a rebase now that the list PR has shipped @renintw

outdoor2kode:
Close in favor of #2600

PR #2588, opened 2024-07-01
Featured Lessons only on front page

### LINKED ISSUE #2575: Featured Lessons on home page
Currently the 6 Lessons on the home page are simple the most recent Lessons published. This means there is very little control over what shows up there.

To match the behaviour of the Featured Courses section above, we can follow the same approach and add a `featured` post meta field to Lessons, enabling precise selection of the Lesson set.

The query loop for that section will then need a custom query loop block (following that created for featured courses) which enables filtering by the post meta field.

- [ ] Add featured Lesson post meta to plugin
- [ ] Add featured Lessons block variation to theme
- [ ] Update home page template to use featured Lessons

adamwoodnz:
This adds a new 'Featured' post meta field to Lessons (the same as Sensei adds to Courses), and a direct copy of the Courses Grid block variation, which enables a 'Featured Lessons only' toggle on a Lesson Query Loop block.

This enables the Lessons section of the home page to display Featured Lessons only.

Closes #2575 

### Screenshots

| Lesson control | Grid control | Front page |
|-|-|-|
| Screenshot 2024-07-02 at 3 02 19 PM | Screenshot 2024-07-02 at 3 03 15 PM | localhost_8888_(Desktop) (4) |

### Testing

1. Edit or create a Lesson and look for the 'Featured Lesson' setting in the sidebar
2. Test turning on and off and saving. Optionally check the db for the post meta field.
3. Leave the Lesson and open a page or post. In the editor there should be a Learn Lesson Grid block available.
4. Insert this block, and look for the 'Featured' setting in the sidebar
5. Test switching it in the editor. The posts should change and the Lesson you set to featured in step 2 should be displayed when 'Featured only' is set. Also check this page or post on the frontend.
6. Check the front page of the site, it should also only be showing featured Lessons.

### DIFF
--- a/wp-content/plugins/wporg-learn/js/lesson-featured-meta/index.js
+++ b/wp-content/plugins/wporg-learn/js/lesson-featured-meta/index.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { PanelRow, CheckboxControl } from '@wordpress/components';
+import { CheckboxControl, PanelRow } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
 import { useState } from '@wordpress/element';
--- a/wp-content/plugins/wporg-learn/js/lesson-featured-meta/index.js
+++ b/wp-content/plugins/wporg-learn/js/lesson-featured-meta/index.js
@@ -1,0 +1,42 @@
+/**
+ * WordPress dependencies
+ */
+import { PanelRow, CheckboxControl } from '@wordpress/components';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { registerPlugin } from '@wordpress/plugins';
+
+const FEATURED = 'featured';
+
+const LessonFeaturedMeta = () => {
+	const postMetaData = useSelect( ( select ) => select( 'core/editor' ).getEditedPostAttribute( 'meta' ) || {} );
+	const { editPost } = useDispatch( 'core/editor' );
+	const [ lessonFeatured, setLessonFeatured ] = useState( postMetaData?._lesson_featured === FEATURED );
+
+	return (
+		<PluginDocumentSettingPanel title={ __( 'Featured Lesson', 'wporg-learn' ) }>
+			<PanelRow>
+				<CheckboxControl
+					label={ __( 'Feature this lesson', 'wporg-learn' ) }
+					checked={ lessonFeatured }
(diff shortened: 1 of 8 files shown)


wp-content/plugins/wporg-learn/js/lesson-featured-meta/index.js
```diff
@@ -0,0 +1,40 @@
...
+				<CheckboxControl
+					label={ __( 'Feature this lesson', 'wporg-learn' ) }
+					checked={ lessonFeatured }
+					onChange={ ( newLessonFeatured ) => {
+						setLessonFeatured( newLessonFeatured );
+
+						editPost( {
+							meta: {
+								...postMetaData,
+								_lesson_featured: newLessonFeatured ? 'featured' : '',
```
adamwoodnz, line 28: Follows the same string based approach as Featured Courses in sensei-lms.

wp-content/plugins/wporg-learn/js/lesson-featured-meta/index.js
```diff
@@ -0,0 +1,42 @@
...
+ * WordPress dependencies
+ */
+import { PanelRow, CheckboxControl } from '@wordpress/components';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { PluginDocumentSettingPanel } from '@wordpress/edit-post';
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { registerPlugin } from '@wordpress/plugins';
+
+const FEATURED = 'featured';
```
adamwoodnz, line 11: Follows the same string based approach as Featured Courses in sensei-lms.

outdoor2kode:
https://github.com/WordPress/Learn/assets/18050944/4d1d2bec-8db3-4dc9-836f-c5e9fd5c5697

Toggling the option doesn't seem to reflect on the front end.

**Update**

No problems on the course grid, and so is the related code logic.
I also can see that both the editor and the front end have the query
```
array(6) {
  ["post_type"]=>
  string(6) "lesson"
  ["order"]=>
  string(4) "DESC"
  ["orderby"]=>
  string(4) "date"
  ["post__not_in"]=>
  array(0) {
  }
  ["offset"]=>
  int(0)
  ["posts_per_page"]=>
  int(6)
}
```

I think it's caused by data, as those lessons were published at the same time (Published
2024/07/01 at 10:11 pm), not sure why the order is different, though.

outdoor2kode (approved):
Both code and testing cases work look great!

Except for the inconsistency of lessons displayed between the editor and the front end mentioned above, but I think that's probably caused by the data I have locally.

adamwoodnz:
> Except for the inconsistency of lessons displayed between the editor and the front end mentioned above, but I think that's probably caused by the data I have locally.

I haven't tested in sandbox yet due to it mutating data, but I'll be sure to check consistency before deploying.

PR #2594, opened 2024-07-02
Add support for Learning Mode for standalone lessons

### LINKED ISSUE #2593: Single freestanding Lesson layout is broken
Lessons which are not part of a Course have a broken layout, seemingly because the Sensei learning-mode.css styles aren't being loaded. It seems the setting 'Enable Learning Mode for all Courses' doesn't apply, which makes sense.

Example: https://learn.wordpress.org/lesson/what-is-accessiblity-and-why-is-it-important/?new-theme=1

adamwoodnz:
Sensei doesn't seem to officially support the Learning Mode Course theme for Lessons which are not part of a Course.

This PR adds this support by enqueuing the Learning Mode styles if necessary, adding required body classes and custom properties for the styles.

It also hides the notice to take the course and the exit course button, as these don't make sense for a single Lesson.

I have reached out to the Sensei team to see if there is a better way to achieve this.

Fixes #2593 

| Before | After |
|--------|--------|
| learn wordpress org_lesson_what-is-accessiblity-and-why-is-it-important_(Desktop) | learn wordpress org_lesson_what-is-accessiblity-and-why-is-it-important_(Desktop) (1) |

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2024/functions.php
@@ -98,7 +98,7 @@ function enqueue_assets() {
  * Enqueue the styles and add the required body class if needed.
  */
 function maybe_enqueue_sensei_assets() {
-	if ( is_singular( 'lesson' ) && ! wp_style_is( 'sensei-course-theme-style', 'enqueued' ) ) {
+	if ( ( is_singular( 'lesson' ) || is_singular( 'quiz' ) ) && ! wp_style_is( 'sensei-course-theme-style', 'enqueued' ) ) {
 		wp_enqueue_style( 'sensei-learning-mode' );
 
 		add_filter( 'body_class', function( $classes ) {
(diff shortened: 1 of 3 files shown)


adamwoodnz:
Looking at this layout again, it doesn't make sense to have the left column. I'll update to have a centred column.

adamwoodnz:
> Looking at this layout again, it doesn't make sense to have the left column. I'll update to have a centred column.

Layout updated

localhost_8888_lesson_freestanding-featured-lesson_(Desktop) (1)

adamwoodnz:
I got a response back from Sensei, but they didn't have any better solutions unfortunately.

> Yeah, Unfortunately, we don't have anything prepared to have lessons out of a course. My only suggestion, if possible, would be a course for independent lessons. Maybe something like "general knowledge"?
There are also 2 features that could be useful in this case, depending on what exactly you need: Open Access or Lesson Preview.
Or, as another alternative, if it doesn't need any course feature, maybe it could be a normal post with a custom template to look like a lesson?

I think these suggestions are all going to have various effects on search and information architecture, and the features don't help here.

> And the other alternative would be your PR. I'm not sure what types of edge cases it could have, but if it solves your issues and works as the best alternative for you, I think it's fine.

No major red flags, but I'll note it as something to test in QA so we can hopefully find edge cases.

adamwoodnz:
Looks like this needs more work as I think quizzes for freestanding lessons might have the same layout issues

adamwoodnz:
> Looks like this needs more work as I think quizzes for freestanding lessons might have the same layout issues

I've fixed the Quiz layout issues, however Quizzes for standalone Lessons are blocked from rendering by Sensei, see https://github.com/WordPress/Learn/issues/2643

PR #2598, opened 2024-07-03
Align local nav and add breadcrumbs for pages

### LINKED ISSUE #2511: Align local nav & breadcrumbs of pages with design
See https://github.com/WordPress/Learn/pull/2491#issuecomment-2151818712 for design and discussion.

outdoor2kode:
Closes #2511

This PR ensures that the page has the correct breadcrumbs.
Additionally, it makes the local navigation behave according to the design shown in the image below.

image

| **Default** | **Scroll** |
|-|-|
| Screenshot 2024-07-04 at 03 53 35 | Screenshot 2024-07-04 at 03 53 44 |

Also, for the L2B part, it will need to have a breadcrumb level value to help with the determination. So I'm going to wait to see if the [current method](https://github.com/WordPress/Learn/pull/2598/commits/c92a147a66ce779eeed3fb916e4b6d6254ae8445#r1664655722) of retrieving that value makes sense.

## Screenshots

| | **Default** | **Scroll** |
|-|-|-|
| **L3** | Screenshot 2024-07-03 at 22 42 42 | Screenshot 2024-07-03 at 22 42 55 |
|| Screenshot 2024-07-04 at 03 14 19 | Screenshot 2024-07-04 at 03 14 28 |
|| Screenshot 2024-07-04 at 03 14 37 | Screenshot 2024-07-04 at 03 14 43 |
| **L2A** | Screenshot 2024-07-04 at 03 26 06 | Same |
|| Screenshot 2024-07-04 at 03 26 13 | Same |
|| Screenshot 2024-07-04 at 03 26 20 | Same. Note that the link isn't highlighted |
|| Screenshot 2024-07-04 at 03 25 49 | Same. Note that the link isn't highlighted |
|| Screenshot 2024-07-04 at 03 25 58 | Same |
| **L1B** | This site doesn't have L1B | - |
| **L1A** | image | image |


wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
```diff
@@ -0,0 +1,23 @@
...
+ * Slug: wporg-learn-2024/site-breadcrumbs
+ * Inserter: no
+ */
+
+// Ensure breadcrumbs are displayed only when there are at least 3 levels.
+$block_content = '<!-- wp:wporg/site-breadcrumbs {"fontSize":"small"} /-->';
+$parsed_content = do_blocks( $block_content );
+preg_match_all( '/<span\b[^>]*>/', $parsed_content, $matches );
+$breadcrumb_level = count( $matches[0] );
+$show_breadcrumbs = $breadcrumb_level >= 3;
```
outdoor2kode: I tried quite a few methods here, and currently, this one seems the most stable, although it looks a bit hacky.

Have tried using `do_action` and the global variable [here](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2024/functions.php#L205-L257) in functions.php to pass out the breadcrumb level, but both of them don't work in patterns, possibly due to the execution order.

Also tried `update_option` to save the level, but there’s an issue with fetching empty values before they are saved to the database at the beginning.

It’s likely to write a function in wporg-mu-plugins to fetch the value, but that seems equally tricky.

There might be a better way that I haven't thought of yet, please share if you have any ideas, thanks! cc @ryelle

wp-content/themes/pub/wporg-learn-2024/patterns/local-navigation-bar.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+?>
+
+<!-- wp:wporg/local-navigation-bar {"className":"has-display-contents","backgroundColor":"charcoal-2","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"},":hover":{"color":{"text":"var:preset|color|white"}}}}},"textColor":"white","fontSize":"small"} -->
+
+	<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"textColor":"light-grey-1","layout":{"type":"flex","flexWrap":"nowrap"}} -->
+	<div class="wp-block-group has-light-grey-1-color has-text-color">
+		<!-- wp:site-title {"level":0,"fontSize":"small","textColor":"white"} /-->
+		<?php if ( ! ( is_page( 'learning-pathways' ) || is_archive( 'course' ) || is_archive( 'lesson' ) || is_page( 'online-workshops' ) || is_page( 'my-courses' ) ) ) : ?>
```
adamwoodnz: If we used a more template based approach we could avoid having PHP logic here, and therefore a pattern.

Create 2 parts `local-nav.html`, `local-nav-with-title.html`, for the 2 states of this pattern.

Have 2 header parts which use them; existing `header.html` uses whatever the most common state is (looks like `local-nav-with-title.html`), and a second header part for these 4 exception templates (maybe `header-without-title.html`).

A bit tricky to explain, make sense?

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-header.php
```diff
@@ -14,11 +14,7 @@
 	<!-- wp:group {"style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space","top":"0px","bottom":"0px"}}},"className":"sensei-course-theme-header-content","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
 	<div class="wp-block-group sensei-course-theme-header-content" style="padding-top:0px;padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:0px;padding-left:var(--wp--preset--spacing--edge-space)">
 
-		<!-- wp:group {"className":"wporg-breadcrumbs","align":"full","style":{"spacing":{"padding":{"top":"23px","bottom":"23px","right":"var:preset|spacing|edge-space"}}},"backgroundColor":"white","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
-		<div class="wporg-breadcrumbs wp-block-group alignfull has-white-background-color has-background" style="padding-top:23px;padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:23px;">
-			<!-- wp:wporg/site-breadcrumbs {"fontSize":"small"} /-->
-		</div>
-		<!-- /wp:group -->
+		<!-- wp:pattern {"slug":"wporg-learn-2024/site-breadcrumbs"} /-->
```
adamwoodnz: Seem to have gained extra padding here

Screenshot 2024-07-04 at 4 08 39 PM

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
adamwoodnz: Maybe I'm misunderstanding or oversimplifying, but couldn't you check the length of the breadcrumbs in functions.php and return an empty array if < 3?

adamwoodnz (commented):
Looking pretty good to me. Couple of questions/suggestions inline.

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
outdoor2kode: I've tried this, but each `site-breadcrumbs` block is wrapped in a `group` block, so even when return empty, the outer layer still renders. I think I can try removing the outer `group` block, though, haven't tried this way yet.

wp-content/themes/pub/wporg-learn-2024/patterns/local-navigation-bar.php
outdoor2kode: yeah, this is clearer. thanks!

c344813, d33caa1

wp-content/themes/pub/wporg-learn-2024/parts/header.html
```diff
@@ -1,20 +1,6 @@
...
-		<!-- wp:query-title {"type":"filter","level":0,"fontSize":"small","fontFamily":"inter","className":"wporg-local-navigation-bar__fade-in-scroll"} /-->
-
-		<!-- wp:post-title {"level":0,"fontSize":"small","fontFamily":"inter","className":"wporg-local-navigation-bar__fade-in-scroll"} /-->
-	</div>
-	<!-- /wp:group -->
-
-	<!-- wp:navigation {"menuSlug":"learn","icon":"menu","overlayBackgroundColor":"charcoal-2","overlayTextColor":"white","layout":{"type":"flex","orientation":"horizontal"},"fontSize":"small"} /-->
-
-<!-- /wp:wporg/local-navigation-bar -->
+<!-- wp:template-part {"slug":"local-nav","className":"has-display-contents","tagName":"div"} /-->
```
ryelle: Now that you've split out the two local-nav partials, could they be merged back into the header partials?

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
```diff
@@ -0,0 +1,23 @@
...
+
+?>
+
+<?php if ( $show_breadcrumbs ) : ?>
+	<!-- wp:group {"className":"wporg-breadcrumbs","align":"full","style":{"spacing":{"padding":{"top":"23px","bottom":"23px","left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}},"backgroundColor":"white","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
+	<div class="wporg-breadcrumbs wp-block-group alignfull has-white-background-color has-background" style="padding-top:23px;padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:23px;padding-left:var(--wp--preset--spacing--edge-space)">
```
ryelle: I think this should be 18px top/bottom padding, to match the breadcrumbs on the other sites.

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
ryelle: I would reach for a separate template-part for the third-level header, like on the Developer site (https://github.com/WordPress/wporg-developer/pull/500). In this case, I have an additional header template, `header-third`, which is [dynamically switched out](https://github.com/WordPress/wporg-developer/blob/8b33d50a34a8a539a9f2980c4645d1b7fa5850d0/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php#L54-L81) on the pages where we need the breadcrumbs. In that case, I used WP-style conditionals like `is_single()` & `wporg_is_handbook()` to determine whether the breadcrumbs show up. I don't know enough about this site setup to know if you could do something similar, e.g., `is_singular('course')` to show on single courses, etc.

adamwoodnz:
Found another edge case @renintw, freestanding Lessons don't have a Course parent for the breadcrumb. Breadcrumb should probably be Home / Lessons / Lesson Title

Screenshot 2024-07-05 at 12 01 14 PM

wp-content/themes/pub/wporg-learn-2024/parts/header.html
outdoor2kode: Do you mean the change like this: 196baea, for reducing complexity?

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
renintw: Thanks. @fcoveram Could

outdoor2kode: Thanks 👍 0ee33ff  @fcoveram, do you mind also updating styles in the figma? The top/bottom padding for the breadcrumbs:
image

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
outdoor2kode: [dynamically switched out](https://github.com/WordPress/wporg-developer/blob/8b33d50a34a8a539a9f2980c4645d1b7fa5850d0/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php#L54-L81), this approach is awesome, thanks!!

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-header.php
outdoor2kode: Fixed in 0888452

outdoor2kode:
> Found another edge case @renintw, freestanding Lessons don't have a Course parent for the breadcrumb. Breadcrumb should probably be Home / Lessons / Lesson Title

Weird, I don't see this issue either in this PR or in #2594
image

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
outdoor2kode: Alright, I have implemented dynamic switching for the [search page](https://github.com/WordPress/Learn/pull/2598/commits/c7edbf9d69bc23779b5b247a9fc20953c7a6a7ae), and the same can be applied to other cases as well, although I have directly [written it into the template](https://github.com/WordPress/Learn/pull/2598/commits/9d9f84ded8c8d926900ad85aae3ab840e8e5f934) here, as I feel it’s similar.

I also have added quite a few partial templates in order to fully match the design, and while working on it, I noticed that [the L2B](https://developer.wordpress.org/block-editor) on the developer page is broken or it was initially overlooked. The L2B for the page is also the troublesome part I mentioned earlier: the "page" template might render both 2-level (L2B) and 3-level (L3) breadcrumbs simultaneously. (see child and grandchild screenshots)

Currently, I have decided to ensure that the more common 2-level pages conform to the design. As for the 3-level pages, it depends on whether we think it's necessary. If so, we will need to get the current page's level information (which is tricky). Personally, I prefer not to address it for now.

Paired PR: https://github.com/WordPress/wporg-mu-plugins/pull/637

**Updated Screenshots:**

| | **Default** | **Scroll** |
|-|-|-|
| **L3** <br> single course | Screenshot 2024-07-03 at 22 42 42 | Screenshot 2024-07-03 at 22 42 55 |
| grandchild (the only case that doesn't match the design) | Same, but the title should not display, just like above |image |
| **L2B** <br> child | Screenshot 2024-07-08 at 03 27 54 | Same |
course complete | image | Same |
| **L2A** <br> lesson archive | Screenshot 2024-07-04 at 03 26 06 | Same |
| online workshops | Screenshot 2024-07-04 at 03 26 13 | Same |
| my courses | Screenshot 2024-07-04 at 03 26 20 | Same. Note that the link isn't highlighted |
| learning pathways archive | Screenshot 2024-07-04 at 03 25 49 | Same. Note that the link isn't highlighted |
| coruse archive | Screenshot 2024-07-04 at 03 25 58 | Same |
| **L1B** | This site doesn't have L1B | - |
| **L1A** | image | image |
| **Others** <br> learning pathways term | image | image |
| tax search page | image | Same |
| course search page | image | Same |

outdoor2kode: A thought: given that this breadcrumbs + local nav design is used universally across all wporg sites, perhaps we could abstract some of the rules/elements into wporg-mu-plugins. Haven’t looked into it in detail yet, but it seems feasible? Might need quite a few changes for themes using it though.

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
```diff
@@ -1,23 +0,0 @@
...
-$parsed_content = do_blocks( $block_content );
-preg_match_all( '/<span\b[^>]*>/', $parsed_content, $matches );
-$breadcrumb_level = count( $matches[0] );
-$show_breadcrumbs = $breadcrumb_level >= 3;
-
-?>
-
-<?php if ( $show_breadcrumbs ) : ?>
-	<!-- wp:group {"className":"wporg-breadcrumbs","align":"full","style":{"spacing":{"padding":{"top":"18px","bottom":"18px","left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}},"backgroundColor":"white","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
-	<div class="wporg-breadcrumbs wp-block-group alignfull has-white-background-color has-background" style="padding-top:18px;padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:18px;padding-left:var(--wp--preset--spacing--edge-space)">
```
outdoor2kode: Not sure why the site-breadcrumbs needs to be wrapped in a group, as it appears that the style can be adjusted directly on the block itself. I noticed that other sites do this as well, so maybe I'm missing something?

adamwoodnz:
> > Found another edge case @renintw, freestanding Lessons don't have a Course parent for the breadcrumb. Breadcrumb should probably be Home / Lessons / Lesson Title
> 
> Weird, I don't see this issue either in this PR or in #2594 

Try this one https://learn.wordpress.org/lesson/publishing-with-the-block-editor/

wp-content/themes/pub/wporg-learn-2024/inc/block-hooks.php
```diff
@@ -0,0 +1,31 @@
...
+	if (
+		'core/template-part' === $parsed_block['blockName'] &&
+		! empty( $parsed_block['attrs']['slug'] ) &&
+		str_starts_with( $parsed_block['attrs']['slug'], 'header' )
+	) {
+		$template_slug = 'header-second-archive-title';
+		if ( is_search() ) {
+			$parsed_block['attrs']['slug'] = $template_slug;
+		}
+	}
```
adamwoodnz: Combine these conditions?
```suggestion
	if (
		'core/template-part' === $parsed_block['blockName'] &&
		! empty( $parsed_block['attrs']['slug'] ) &&
		str_starts_with( $parsed_block['attrs']['slug'], 'header' ) &&
		is_search()
	) {
		$parsed_block['attrs']['slug'] = 'header-second-archive-title';
	}
```

outdoor2kode: 740e91b

adamwoodnz (approved):
This is looking great, thanks for your perseverance!

Maybe let's get this merged and we can follow up with the standalone Lesson edge case?

outdoor2kode:
> > > Found another edge case @renintw, freestanding Lessons don't have a Course parent for the breadcrumb. Breadcrumb should probably be Home / Lessons / Lesson Title
> > 
> > 
> > Weird, I don't see this issue either in this PR or in #2594
> 
> Try this one https://learn.wordpress.org/lesson/publishing-with-the-block-editor/

I think that's the data issue, as other freestanding lessons look fine, though I didn't check all the lessons, there might be some out there. This can be fixed by saving the post again, it's likely that the lesson originally belonged to a course, and that course was deleted but the related data for the lesson is still there.

Open a ticket for it: https://github.com/WordPress/Learn/issues/2644#issue-2394564182

Note: this is the [logic](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2024/functions.php#L254-L256) where processing the freestanding lessons, for the example raised here, it has a value for `lesson_course_id`, but the course doesn't exist. That's why I think it originally belonged to a course.

wp-content/themes/pub/wporg-learn-2024/patterns/site-breadcrumbs.php
fcoveram: Sure. I will apply the change

PR #2651, opened 2024-07-09
Update icon, font size, color for Learning Duration

### LINKED ISSUE #2627: Adjust the font size and icon in course cards for the "time" part
There’s also something about the font sizes and colors of these items at the footer:

Image

The icon should be updated to use the new time icon from [the Figma](https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp/Learn?node-id=1705-79&t=VyCt6LYpASvg63gx-1), and the font size and color of the text (6 hours) should be the same as that of 59 lessons.

This has been fixed in the Figma file, but was previously wrong there.

ryelle:
Fixes #2627 — This updates the icon for the Learning Duration block to use the SVG from figma, instead of a dashicon. It also updates the style attributes on each instance of the block so that it matches the "Lesson Count" block, which inherits text color and uses `extra-small` font. This matches the figma, which is 12px default text color (charcoal-1).

| Before | After |
|---|---|
| Screenshot 2024-07-09 at 2 14 13 PM | Screenshot 2024-07-09 at 2 04 31 PM |

**To test**

- Add durations to some courses. This block is also used on lessons, though none on production seem to have duration; but the style update has been done for lessons as well.
- Everywhere the duration appears, it should be 12px font size, charcoal-1 color, and the icon should be an SVG (the SVG has a thinner circle border around it)


adamwoodnz (approved):
LGTM 🚢

PR #2652, opened 2024-07-09
Add copy & expand buttons to syntax-highlighted code blocks

### LINKED ISSUE #2572: Feedback - New theme code samples don't include copy/expand buttons.
## Description
When viewing a lesson plan that includes a code block, the previous theme has Copy/Expand buttons. The next code sample does not.

1. Go to https://learn.wordpress.org/lesson/accessing-wordpress-data/
2. Scroll down
3. Notice the code block has no copy/expand buttons

| Old | New |
|--------|--------|
| Screenshot 2024-06-26 at 10 52 32 AM | Screenshot 2024-06-26 at 10 52 25 AM |

ryelle:
Fixes #2572 — This adds the copy & expand buttons from Developer to the code blocks on Learn. I started from the [function-reference.js](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/js/function-reference.js) script, but removed the jQuery dependency & other developer-custom code.

This "block" is not actually a block, it's just using the folder structure for the build process. On init, it registers a script and style, and when the code block is rendered, the script and style are loaded. When the JS sees a code block, if it's using the syntax highlighting, it injects a copy button & maybe the expand button.

Initially I started this as a block that could live in wporg-mu-plugins (and it worked), but trying to get it to work with the existing developer blocks might be a bit more coordination than we want to do for Learn's launch. Moving this back into mu-plugins could be a future improvement though.

| Before | After |
|---|---|
| Screen Shot 2024-07-09 at 18 46 26 | Screen Shot 2024-07-09 at 18 40 30 |

Note: I restricted this to syntax-highlighted only because these blocks also have some styles that make it look better (padding, primarily). For example, if it's on all code blocks, these look awkward:

Screenshot 2024-07-09 at 6 58 33 PM

**To test**

- View a lesson with code syntax blocks, ex https://learn.wordpress.org/lesson/accessing-wordpress-data/
- The code blocks should have copy buttons, and if they're longer than 12 lines, they should be collapsed with an expand button
- View a lesson with code blocks without syntax highlighting, ex http://localhost:8888/lesson/javascript-concepts/
- No copy or expand buttons


adamwoodnz (approved):
Works for me, very nice!

:shipit:

PR #2657, opened 2024-07-11
Prioritize Courses, Lessons, Lesson Plans and Tutorials matching user's locale

### LINKED ISSUE #2607: New Theme: Lessons should only show my language first.
## Description
When I visit https://learn.wordpress.org/lessons/?new-theme=1, it shows me lessons that are not my language.

1. Go to https://learn.wordpress.org/lessons/?new-theme=1

learn wordpress org_lessons__new-theme=1

adamwoodnz:
Moves the existing prioritization for Tutorials to the plugin so that it is enabled on the new theme as well.

Copies the Tutorials query modification and applies it to Courses and Lessons, with one significant change: when the locale is English then Courses and Lessons with no language post meta set will also be prioritized over those with other languages. This matches the default post meta value for Language. Unless the Language has been manually set the value will actually be empty, even though the setting looks saved in the editor.

Closes #2607 

### Testing

Easiest in sandbox unless you have languages set up locally.

1. Regression test Tutorials prioritization with the old theme. Change your locale using the locale switcher and the order of the Tutorials archive should update accordingly, with those matching your locale first
2. Switch to the new theme and do the same for the Tutorials, Lessons and Courses archives
3. For Courses and Lessons there is very little non-english content so navigating to the last page will likely contain all that content
4. There are no non-english courses published so I have privately published a Japanese one. If logged in as admin you should see it.


wp-content/plugins/wporg-learn/inc/locale.php
```diff
@@ -156,3 +157,162 @@ function disable_caching( $headers ) {
...
+		if ( false === strpos( $clauses['groupby'], "{$wpdb->posts}.ID" ) ) {
+			$clauses['groupby'] = "{$wpdb->posts}.ID";
+		}
+	}
+
+	return $clauses;
+}
+
+/**
+ * Modify the a post type archive query to prioritize posts in the user's locale.
```
outdoor2kode:
```suggestion
 * Modify the post type archive query to prioritize posts in the user's locale.
```

wp-content/plugins/wporg-learn/inc/locale.php
```diff
@@ -156,3 +157,162 @@ function disable_caching( $headers ) {
...
+ * AND wp_posts.post_type = 'lesson'
+ * GROUP BY wp_posts.ID
+ * ORDER BY has_language DESC, wp_posts.post_date DESC
+ *
+ * @param array  $clauses
+ * @param string $locale
+ *
+ * @return array
+ */
+function wporg_query_prioritize_locale( $clauses, $locale ) {
```
outdoor2kode, line 303: It seems that these two functions have some similarity. Do you think it would be clearer to combine them into one function like: `wporg_query_prioritize_locale($clauses, $locale, $post_type = '')`?

The condition inside could be changed to something like this:
```
// Common field for both post types
$clauses['fields'] .= ",
    MAX(
        CASE
            WHEN pmeta.meta_key = 'language' AND pmeta.meta_value LIKE '{$locale_root}%' THEN 1
            WHEN (pmeta.meta_key = 'language' AND pmeta.meta_value = '') OR pmeta.meta_key IS NULL THEN " . ( $is_english ? '1' : '0' ) . '
            ELSE 0
        END
    ) AS has_language
';

// Additional fields and joins for 'workshop' post type
if ( $post_type === 'workshop' ) {
    $clauses['fields'] .= ",
        MAX( IF( pmeta.meta_key = 'video_caption_language' AND pmeta.meta_value LIKE '{$locale_root}%', 1, 0 ) ) AS has_caption
    ";
}

$clauses['join'] .= " LEFT JOIN {$wpdb->postmeta} pmeta ON ( {$wpdb->posts}.ID = pmeta.post_id AND pmeta.meta_key IN ('language'" . ( $post_type === 'workshop' ? ", 'video_caption_language'" : "" ) . ") )";

$clauses['orderby'] = 'has_language DESC' . ( $post_type === 'workshop' ? ', has_caption DESC' : '' ) . ', ' . $clauses['orderby'];
$clauses['groupby'] = "{$wpdb->posts}.ID";
```
And the examples in the comments for the workshop can be moved here as well.

I feel that this way having the two query instructions together makes it clearer what the differences are. It would be faster to maintain and understand the code. This is just my initial thought when first seeing this part.

outdoor2kode (approved):
Tested all the cases and LGTM! Left a couple of comments, nothing's a real blocker.

wp-content/plugins/wporg-learn/inc/locale.php
adamwoodnz: This is a fair point. The main difference between the two is that the Tutorials one does not account for empty or null post meta being treated as English. That's the way prod is at present, and given Tutorials are being deprecated I thought I'd leave that query alone.

But we could improve it, what do you think?

adamwoodnz: Maybe a deprecation note in the doc would make it clear that it has been superseded.

outdoor2kode: A deprecation note makes sense 👍

adamwoodnz: Done [6121cc3](https://github.com/WordPress/Learn/pull/2657/commits/6121cc338d9061bb7cc7716f4a310e155d5eaeff)

adamwoodnz:
I decided we may as well apply this prioritization to Lesson Plans as well. It affects production but seems to function as intended.

outdoor2kode:
> I decided we may as well apply this prioritization to Lesson Plans as well. It affects production but seems to function as intended.

Isn't lesson plan is going to be deprecated as well? Does it make sense to combine it with tutorial?

adamwoodnz:
> Isn't lesson plan is going to be deprecated as well?

Eventually, but not as soon. Lesson Plans will be around until we have a solution for Teacher's notes on Lessons

> Does it make sense to combine it with tutorial?

Lesson plans don't have captions so I'm not sure how compatible that function would be, but [I've now added a guard](https://github.com/WordPress/Learn/pull/2657/commits/f2d920fcbba2dc39ec640e09a954c9a8895075b5) so that the new function doesn't run with the old theme. No point in adding any extra testing or risk to this.

PR #2658, opened 2024-07-11
Add a post meta filter to exclude certain lessons from archive views

### LINKED ISSUE #2591: Filter quiz and completed lessons from Lessons archive
On the Lessons archive there are items included which arguably aren't valuable to be displayed, eg. a Lesson which only holds a quiz:

In order to not waste screen real estate displaying these, we should filter them out.

### Questions

Can we do this simply by filtering using the content of the title?

If so what terms should be filtered; 'quiz', anything else?

adamwoodnz:
Adds a post meta field  to Lessons, which enables hiding certain Lessons from archive views, eg. Quiz, Course Completed.

Adds a meta_query on archive queries to filter these posts out.

Closes #2591 

### Testing

1. Use the editor post meta to hide a lesson
2. Navigate the frontend and ensure the lesson doesn't appear in any of the lesson grids
3. Ensure it still appears in the Course it belongs to

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -51,6 +51,21 @@ function register_lesson_meta() {
 			},
 		),
 	);
+
+	register_post_meta(
+		'lesson',
+		'_lesson_archive_excluded',
+		array(
+			'description'       => __( 'Whether the lesson should be excluded from archive views.', 'wporg-learn' ),
+			'type'              => 'string',
+			'single'            => true,
+			'sanitize_callback' => 'sanitize_text_field',
+			'show_in_rest'      => true,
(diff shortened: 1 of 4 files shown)


wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -51,6 +51,21 @@ function register_lesson_meta() {
 			},
 		),
 	);
+
+	register_post_meta(
+		'lesson',
+		'_lesson_archive_excluded',
+		array(
+			'description'       => __( 'Whether the lesson should be excluded from archive views.', 'wporg-learn' ),
+			'type'              => 'string',
```
adamwoodnz, line 60: String based approach follows that of [featured courses](https://github.com/Automattic/sensei/blob/trunk/includes/class-sensei-course.php#L669), and has already been reused for featured lessons above.

PR #2659, opened 2024-07-11
Brand name typo

NekoJonez:
Hesitating if we shouldn't change "Twitter" to "X"?


jonathanbossenger:
Good catch.

It looks like the [Online Workshop page on the new theme doesn't have that form](https://learn.wordpress.org/online-workshops/?new-theme=1), and will direct users to [this form anyway](https://learn.wordpress.org/online-workshops/apply-to-facilitate/), which doesn't include any specific social media names

jonathanbossenger (approved):

Issue #2638, opened 2024-07-05
Single Lesson: Polish the design


jasmussen:
There’s some design work for the full-screen lessons that lives [in Figma here](https://href.li/?https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp/Learn?node-id=3473-34133&t=19mV6GnRSxbQzM21-1), that could be worth looking at. I realize it may be limited what you can do here, but there are some things about icons, positioning, sizes, etc, that could benefit from small tweaks.

![Image](https://github.com/WordPress/Learn/assets/1204802/7a26d5a3-867d-4323-a316-8c9b9f64e26f)

fcoveram:
Do we have room now to revisit the design of lessons @renintw @adamwoodnz?

I noticed a few inconsistencies and room for improvement in terms of spacing and text styles that also relate to #2809

outdoor2kode:
![image](https://github.com/user-attachments/assets/69604ca8-d101-4a3a-bb8a-eb8adcde4a63)

Hi @fcoveram 👋  regarding #2809, I noticed that there isn't a related design presented in Figma (or am I missing it?). #2809 refers to the "Lesson Completed" indicator, which appears after clicking "Complete Lesson." It was removed in #2806 due to the reasons @ryelle mentioned below. Could we consider redesigning it based on Kelly's or my ideas? Thanks!

**From Kelly** https://github.com/WordPress/Learn/pull/2806#issuecomment-2263389133

> I disagree with this. Part of the original feedback in the a11y issue was that "the second button is an interactive control that is not disabled, but does not do anything. This is confusing for all users. It should just be text, and not be an interactive element."
> 
> It is not really a disabled button, it's meant to convey information to the user. There's a suggestion that we could add the "completed" label to the main lesson title, and I think that would be better - maybe something like this

**Mine** https://github.com/WordPress/Learn/pull/2806#issuecomment-2263672121

> I was originally thinking about adding a green label somewhere to make it more obvious, for example like adding an icon like you did and making the entire title green. Also, the "completed" status icon in the sidebar should probably be green as well, to make it more consistent with other places.

outdoor2kode:
I also noticed that the progress bar in the design is still blue, are we aiming to change it back to blue as it's green at the moment in the production. I feel green is more consistent.
| Design | Prod |
|-|-|
[...]

Issue #2601, opened 2024-07-04
Lesson template - Stray pipe


kathrynwp:
On lesson pages that are part of a course, there seems to be a stray pipe` | ` displaying to the left of the module name, near the top left:

### Examples

<img width="1496" alt="Log_In_With_WP-Admin___Learn_WordPress_and_Inbox__817__-_kathryn_a8c_com_-_Automattic_Mail" src="https://github.com/WordPress/Learn/assets/7662449/f73e0448-cb9b-482f-a7d4-2accdcda18bb">

from: https://learn.wordpress.org/lesson/how-to-log-in-with-wp-admin/

<img width="1498" alt="Introduction_to_WordPress___Learn_WordPress_and_Inbox__817__-_kathryn_a8c_com_-_Automattic_Mail" src="https://github.com/WordPress/Learn/assets/7662449/138bff95-3383-4171-be0c-ad485a2a74d0">

from: https://learn.wordpress.org/lesson/introduction-to-wordpress-2/

-----

Should this be removed, or is there some additional information intended to go to the left of the pipe? I don't think we need the course name there since it's already in the breadcrumbs at the top left. Couldn't find this anywhere in the Figma mockups.

adamwoodnz:
Will be fixed as part of #2638

PR #2669, opened 2024-07-11
Add RTL style overrides, ensure RTL CSS is loaded on RTL sites

### LINKED ISSUE #2483: RTL Audit
- [x] Change to a right to left language and browse the site
- [x] Log issues as comments below
- [x] Fix issues

ryelle:
Fixes #2483. This updates the style.css registration to create an RTL replacement on RTL sites — this means the `style-index-rtl.css` file will be used instead. This PR also includes a few overrides specific to RTL sites, in the `_rtl.scss` file.

The issues were broadly with the Meeting Calendar plugin, Sensei, and hardcoded styles on blocks.

- The meeting calendar could probably be fixed by updating the wp-scripts package there, which would create RTL styles (but the 15 -> 27 version jump would probably cause other issues).
- Sensei's RTL support is limited, see https://github.com/Automattic/sensei/issues/1836, so there are overrides in the `_rtl.scss` file.
- The Learning Pathway Card sets the padding on the group block directly, and since it's loaded via a template, the `convert_inline_style_to_rtl` never runs to flip those values (hopefully the `the_block_template_html` filter will be added in 6.7, which will be a better place to run that filter) — in any case, I've added some CSS to `_rtl.scss` for that case as well.

**Screenshots**

| | Before | After |
|---|---|---|
| [Lesson page](https://learn.wordpress.org/lesson/how-to-find-help-with-wordpress-2/) |  | Screen Shot 2024-07-11 at 16 50 11 |
| [Learning Pathway homepage](https://learn.wordpress.org/learning-pathways/) |  | Screen Shot 2024-07-11 at 16 57 09 |
| [Online workshops](https://learn.wordpress.org/online-workshops/) |  | Screen Shot 2024-07-11 at 16 58 56 |
| [Course homepage](https://learn.wordpress.org/course/beginner-wordpress-developer/) | Screen Shot 2024-07-11 at 17 03 53 | Screen Shot 2024-07-11 at 16 59 43 |
| [Quiz](https://learn.wordpress.org/quiz/wordpress-taxonomies-2/) |  | Screen Shot 2024-07-11 at 17 03 01 |

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
+++ b/wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
@@ -1,0 +1,44 @@
+/**
+ * Styles for RTL sites.
+ * Note, these styles are set here using the LTR layout keywords, so that when
+ * `rtlcss` runs it flips the right/left to the correct values.
+ */
+
+body.rtl {
+
+	// Override the inline styles on Learning Pathway Card content.
+	.wporg-learn-learning-pathway-card-header-content {
+		padding-right: 0 !important;
+		padding-left: 30px !important;
+	}
+
+	// Override the alignment on Sensei buttons.
+	.wp-block-sensei-lms-button-take-course.has-text-align-left {
+		text-align: unset;
+	}
+
+	// Flip the chevrons on Sensei lesson listings, "back to lesson".
+	.wp-block-sensei-lms-course-outline-lesson__chevron,
+	.sensei-lms-quiz-back-to-lesson svg {
+		transform: rotate(180deg);
+	}
+
+	// Override the space between quiz inputs and question labels.
+	&.quiz form #sensei-quiz-list li ul li input {
+		margin: 0 10px 0 0;
+	}
+
+	// Fix the border on the Sensei lesson page.
+	.wp-block-sensei-lms-course-theme-lesson-module {
+		border-left: 2px solid;
+		border-right: 0;
+		padding-left: 0.5em;
+		padding-right: 0;
+	}
+
+	// Fix the spacing of the locked lesson icon.
+	.sensei-course-theme-locked-lesson-notice__icon {
+		padding-right: 14px;
+		padding-left: 0;
+	}
+}
--- a/wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
+++ b/wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
@@ -36,14 +36,6 @@ body.rtl {
 		}
 	}
 
-	// Fix the border on the Sensei lesson page.
-	.wp-block-sensei-lms-course-theme-lesson-module {
-		border-left: 2px solid;
-		border-right: 0;
-		padding-left: 0.5em;
-		padding-right: 0;
-	}
-
 	// Fix the spacing of the locked lesson icon.
 	.sensei-course-theme-locked-lesson-notice__icon {
 		padding-right: 14px;
--- a/wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
+++ b/wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
@@ -28,6 +28,14 @@ body.rtl {
 		margin: 0 10px 0 0;
 	}
 
+	&.quiz form #sensei-quiz-list .sensei-multiple-choice-answer-option-checkbox + label {
+		margin-left: 0;
+
+		&::before {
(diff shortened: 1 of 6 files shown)


wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
```diff
@@ -0,0 +1,44 @@
...
+	.sensei-lms-quiz-back-to-lesson svg {
+		transform: rotate(180deg);
+	}
+
+	// Override the space between quiz inputs and question labels.
+	&.quiz form #sensei-quiz-list li ul li input {
+		margin: 0 10px 0 0;
+	}
+
+	// Fix the border on the Sensei lesson page.
```
adamwoodnz: This border isn't actually supposed to be displayed; it's a Sensei thing and the design doesn't show it. An [issue](https://github.com/WordPress/Learn/issues/2601) was opened and closed to be addressed in https://github.com/WordPress/Learn/issues/2638. So I think we can remove this and unset the border and padding for all modes.

adamwoodnz:
The My Courses filter still need to be switched to the opposite side

Screenshot 2024-07-12 at 11 38 54 AM

adamwoodnz:
Quiz radio inputs look good but checkboxes are misaligned, see https://learn.wordpress.org/quiz/how-to-use-headings-for-accessibility-and-seo/

Screenshot 2024-07-12 at 11 43 57 AM

adamwoodnz:
The images in the Learning Pathway cards get cropped from the wrong side on smaller screens (screenshots are at iPad landscape size)

| LTR | RTL |
|--------|--------|
| learn wordpress org_learning-pathways__locale=art_xpirate(iPad) | learn wordpress org_learning-pathways_(iPad) | 

And on the homepage some borders need switching

learn wordpress org__locale=ar(iPad)

Screenshot 2024-07-12 at 11 54 39 AM

adamwoodnz (commented):
Looking much better, but I found a few more things sorry

wp-content/themes/pub/wporg-learn-2024/src/style/_rtl.scss
ryelle: Removed it in https://github.com/WordPress/Learn/pull/2669/commits/62adb0cf2817e868147bd0be0e9feb24975478dc

| LTR | RTL |
|---|---|
| Screenshot 2024-07-15 at 1 39 27 PM | Screenshot 2024-07-15 at 1 39 16 PM |

ryelle:
> The My Courses filter still need to be switched to the opposite side

This looks correct for me:
Screenshot 2024-07-15 at 1 09 31 PM

> And on the homepage some borders need switching

I can't replicate this, borders look fine to me: 
Screenshot 2024-07-15 at 1 27 15 PM

Do you have an out of date wporg-parent on your sandbox, maybe? These are both things that [the RTL filters](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/inc/gutenberg-tweaks.php) should fix.

Quizzes with checkboxes have been fixed (https://github.com/WordPress/Learn/pull/2669/commits/6d12c9efba2f29ccf43586efe19ec068602b0041)

Screen Shot 2024-07-15 at 13 40 52

And I've flipped the image alignment on learning pathways cards for RTL sites (https://github.com/WordPress/Learn/pull/2669/commits/c53f67c0926c21791d6b7f6c8778b6426fd3f33d)

Screen Shot 2024-07-15 at 13 41 20

adamwoodnz (approved):
LGTM now except for the border styles. I've updated all on my sandbox and still seeing them. The culprit seems to be https://learn.wordpress.org/wp-content/plugins/gutenberg/build/block-library/common-rtl.css?ver=18.8.0

Screenshot 2024-07-18 at 11 46 24 AM

If you're not seeing it then maybe ignore.

ryelle:
Aha, I managed to trigger it by preventing inline styles — I was seeing this as an inline style, but it's using the unflipped LTR version here (that seems like a different problem 😬 )

Screenshot 2024-07-18 at 6 03 58 PM

This style should not be flipped in the RTL stylesheet, and is [a bug in gutenberg](https://github.com/WordPress/gutenberg/pull/44170)- by setting the `border-right-style` it inherits default a width & color; and since the left style is not set, it's not visible.

This can be worked around by setting the right style (to make the inline-end border visible) and the left width 0 (to hide the inline-start border). Left/right are flipped in the RTL function in the parent theme, so it's easier to talk in logical property direction 🙂

PR #2671, opened 2024-07-14
Fix My Courses returns all courses when 'All' is selected

### LINKED ISSUE #2609: New Theme: My Course returns all courses.
## Description
When I visit [my courses](https://learn.wordpress.org/my-courses/?new-theme=1), I see a lot of courses, many that I haven't interacted with.

learn wordpress org_my-courses_

outdoor2kode:
Fixes #2609

Because the `get_course_ids_to_be_excluded` function from sensei-lms directly [returns an empty array](https://github.com/Automattic/sensei/blob/trunk/includes/blocks/course-list/class-sensei-course-list-student-course-filter.php#L95-L97) when encountering "all" and unexpected options, it causes the filter to return all courses when "all" is selected, instead of the courses that a student has been enrolled in. Since there is no native hook to adjust this, it has been ported from Sensei. 
The [filter_course_list](https://github.com/Automattic/sensei/blob/trunk/includes/blocks/course-list/class-sensei-course-list-filter-block.php#L30) function from `class-sensei-course-list-filter-block.php` is ported over, and unnecessary conditions and commands are removed.

The old theme used the learner-courses to directly [render](https://github.com/Automattic/sensei/blob/trunk/includes/blocks/class-sensei-learner-courses-block.php#L61) the [Sensei_Shortcode_User_Courses](https://github.com/Automattic/sensei/blob/trunk/includes/shortcodes/class-sensei-shortcode-user-courses.php) shortcode, which retrieves the expected course list. You can see from [here](https://github.com/Automattic/sensei/blob/trunk/includes/shortcodes/class-sensei-shortcode-user-courses.php#L235) that it specifically fetches enrolled courses for "all".

**Testing:**

1. The filters on the My Courses page should now work as expected.
2. The newly added filter should only affect the My Courses page.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -568,10 +568,7 @@ function inject_other_filters( $key ) {
  * @return array The modified course query.
  */
 function modify_course_query( $query ) {
-	if (
-		'course' === $query['post_type'] &&
-		get_the_ID() === Sensei()->settings->get_my_courses_page_id()
-	) {
+	if ( get_the_ID() === Sensei()->settings->get_my_courses_page_id() ) {
 		$key             = get_student_course_filter_query_var_name();
 		$selected_option = isset( $_GET[ $key ] ) ? sanitize_text_field( wp_unslash( $_GET[ $key ] ) ) : '';
 
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -570,8 +570,7 @@ function filter_student_course_list( $parsed_block ) {
 	if (
 		'core/query' !== $parsed_block['blockName'] ||
 		'course' !== ( $parsed_block['attrs']['query']['postType'] ?? '' ) ||
-		! isset( $parsed_block['attrs']['queryId'] ) ||
-		! get_the_ID() === Sensei()->settings->get_my_courses_page_id()
+		get_the_ID() !== Sensei()->settings->get_my_courses_page_id()
 	) {
 		return $parsed_block;
 	}
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -561,7 +561,7 @@ function inject_other_filters( $key ) {
 }
 
 /**
- * Filter the block data for the student course list.
+ * Modifies the courses query for the My Courses page, by adding courses to be excluded.
  *
  * @param array $parsed_block The parsed block data.
  * @return array The updated block data.
@@ -591,9 +591,9 @@ function filter_student_course_list( $parsed_block ) {
 }
 
 /**
- * Get the course IDs to be excluded from the query.
+ * Get the course IDs to be excluded from the query, using the query filter value.
  */
-function get_course_ids_to_be_excluded( $query_id ): array {
+function get_course_ids_to_be_excluded(): array {
 	$user_id = get_current_user_id();
 	if ( empty( $user_id ) ) {
 		return array();
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -564,7 +564,7 @@ function inject_other_filters( $key ) {
  * Modifies the courses query for the My Courses page, by adding courses to be excluded.
  *
  * @param array $parsed_block The parsed block data.
- * @return array The updated block data.
+ * @return array The modified block data.
  */
 function filter_student_course_list( $parsed_block ) {
 	if (
@@ -580,7 +580,16 @@ function filter_student_course_list( $parsed_block ) {
 		$parsed_block['attrs']['query']['exclude'] = array();
 	}
 
-	$exclude_ids = get_course_ids_to_be_excluded( $parsed_block['attrs']['queryId'] );
+	$filter_param_key = get_student_course_filter_query_var_name();
+	$selected_option  = isset( $_GET[ $filter_param_key ] ) ? sanitize_text_field( wp_unslash( $_GET[ $filter_param_key ] ) ) : '';
+
+	// The courses query with active and completed statuses have already been filtered in Sensei LMS, and can correctly display the course lists.
+	// See https://github.com/Automattic/sensei/blob/trunk/includes/blocks/course-list/class-sensei-course-list-student-course-filter.php#L114-L123.
+	if ( ! empty( $selected_option ) && ( 'active' === $selected_option || 'completed' === $selected_option ) ) {
+		return $parsed_block;
+	}
+
+	$exclude_ids = get_course_ids_to_be_excluded();
 
 	$parsed_block['attrs']['query']['exclude'] = array_merge(
 		$parsed_block['attrs']['query']['exclude'],
@@ -591,16 +600,26 @@ function filter_student_course_list( $parsed_block ) {
 }
 
 /**
- * Get the course IDs to be excluded from the query, using the query filter value.
+ * Get a list of course IDs to be excluded from the course lists filtered by the 'all' status
+ * or those that do not belong to the student course options.
+ *
+ * Corresponds to https://github.com/Automattic/sensei/blob/trunk/includes/blocks/course-list/class-sensei-course-list-student-course-filter.php#L95
  */
-function get_course_ids_to_be_excluded(): array {
+function get_course_ids_to_be_excluded() {
 	$user_id = get_current_user_id();
+
 	if ( empty( $user_id ) ) {
 		return array();
 	}
 
-	$filter_param_key = get_student_course_filter_query_var_name();
-	$selected_option  = isset( $_GET[ $filter_param_key ] ) ? sanitize_text_field( wp_unslash( $_GET[ $filter_param_key ] ) ) : '';
+	$args = array(
+		'posts_per_page' => -1,
+		'fields'         => 'ids',
+	);
+
+	$learner_manager     = Sensei_Learner::instance();
+	$courses_query       = $learner_manager->get_enrolled_courses_query( $user_id, $args );
+	$included_course_ids = $courses_query->posts;
 
 	$args           = array(
 		'post_type'      => 'course',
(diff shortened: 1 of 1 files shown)


wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,78 @@ function inject_other_filters( $key ) {
...
+ *
+ * @param array $parsed_block The parsed block data.
+ * @return array The updated block data.
+ */
+function filter_student_course_list( $parsed_block ) {
+	if (
+		'core/query' !== $parsed_block['blockName'] ||
+		'course' !== ( $parsed_block['attrs']['query']['postType'] ?? '' ) ||
+		! isset( $parsed_block['attrs']['queryId'] ) ||
+		! is_page( 'my-courses' )
```
adamwoodnz: The My Courses page can be changed via Sensei settings, which is why we have this function in the plugin: https://github.com/WordPress/Learn/blob/e851a022edd129f7bb8d1c8780729e64ca31bf89/wp-content/plugins/wporg-learn/inc/sensei.php#L262

Maybe you could check if the current page id matches `Sensei()->settings->get_my_courses_page_id()`?

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,78 @@ function inject_other_filters( $key ) {
...
 	}
 }
+
+/**
+ * Filter the block data for the student course list.
+ *
+ * @param array $parsed_block The parsed block data.
+ * @return array The updated block data.
+ */
+function filter_student_course_list( $parsed_block ) {
```
adamwoodnz: I think we could improve the docs here
```suggestion
/**
 * Modifies the courses query for the My Courses page, by adding courses to be excluded.
 *
 * @param array $parsed_block The parsed block data.
 * @return array The updated block data.
 */
function filter_student_course_list( $parsed_block ) {
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,78 @@ function inject_other_filters( $key ) {
...
+		$parsed_block['attrs']['query']['exclude'],
+		$exclude_ids
+	);
+
+	return $parsed_block;
+}
+
+/**
+ * Get the course IDs to be excluded from the query.
+ */
```
adamwoodnz:
```suggestion
/**
 * Get the course IDs to be excluded from the query, using the query filter value.
 */
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,78 @@ function inject_other_filters( $key ) {
...
+function get_course_ids_to_be_excluded( $query_id ): array {
```
adamwoodnz: $query_id isn't used
```suggestion
function get_course_ids_to_be_excluded(): array {
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,78 @@ function inject_other_filters( $key ) {
...
+		! is_page( 'my-courses' )
+	) {
+		return $parsed_block;
+	}
+
+	if ( ! array_key_exists( 'exclude', $parsed_block['attrs']['query'] ) || ! is_array( $parsed_block['attrs']['query']['exclude'] ) ) {
+		$parsed_block['attrs']['query']['exclude'] = array();
+	}
+
+	$exclude_ids = get_course_ids_to_be_excluded( $parsed_block['attrs']['queryId'] );
```
adamwoodnz: Thinking about this further, since `active` and `completed` were working before, do we need to run this exclusion logic for them? It seems quite expensive as there are several queries and merges/diffs involved.

adamwoodnz (commented):
Works for me, nice one.

I do wonder if we're doing more work than we need to though, see comments inline.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
adamwoodnz: Just tried this and it works for me locally. We could then simplify `get_course_ids_to_be_excluded` to only handle the `all` case.

```suggestion
        $filter_param_key = get_student_course_filter_query_var_name();
	$selected_option  = isset( $_GET[ $filter_param_key ] ) ? sanitize_text_field( wp_unslash( $_GET[ $filter_param_key ] ) ) : '';

	if ( ! empty( $selected_option ) && ( 'active' === $selected_option || 'completed' === $selected_option ) ) {
		return $parsed_block;
	}
	
	$exclude_ids = get_course_ids_to_be_excluded( $parsed_block['attrs']['queryId'] );
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: Thanks! This makes sense. 6a80b53

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: bec4e5c

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: bec4e5c

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: bec4e5c

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: Yeah, you are right. Although 'active' and 'completed' were working before, I am writing them here because I'd like to figure out their order first, so I haven't dealt with them. I also thought to first see if this method is a good approach before making further modifications.

7fb7b62
In the commit, I'm not just targeting 'all', but all statuses other than 'completed' and 'active'. This corresponds to the logic [here](https://github.com/Automattic/sensei/blob/trunk/includes/blocks/course-list/class-sensei-course-list-student-course-filter.php#L95), which means if it's not 'active' or 'completed', it shows all enrolled courses.

I'm not sure if the comment added is simple and clear enough, though..

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,74 @@ function inject_other_filters( $key ) {
...
+ *
+ * @param array $parsed_block The parsed block data.
+ * @return array The modified block data.
+ */
+function filter_student_course_list( $parsed_block ) {
+	if (
+		'core/query' !== $parsed_block['blockName'] ||
+		'course' !== ( $parsed_block['attrs']['query']['postType'] ?? '' ) ||
+		! isset( $parsed_block['attrs']['queryId'] ) ||
+		! get_the_ID() === Sensei()->settings->get_my_courses_page_id()
```
adamwoodnz: Without this change the code is running for me on the courses archive:

```suggestion
		get_the_ID() !== Sensei()->settings->get_my_courses_page_id()
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,74 @@ function inject_other_filters( $key ) {
...
+ * Modifies the courses query for the My Courses page, by adding courses to be excluded.
+ *
+ * @param array $parsed_block The parsed block data.
+ * @return array The modified block data.
+ */
+function filter_student_course_list( $parsed_block ) {
+	if (
+		'core/query' !== $parsed_block['blockName'] ||
+		'course' !== ( $parsed_block['attrs']['query']['postType'] ?? '' ) ||
+		! isset( $parsed_block['attrs']['queryId'] ) ||
```
adamwoodnz: This condition isn't relevant anymore, as we don't use it

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,74 @@ function inject_other_filters( $key ) {
...
+	}
+
+	$args = array(
+		'posts_per_page' => -1,
+		'fields'         => 'ids',
+	);
+
+	$learner_manager     = Sensei_Learner::instance();
+	$courses_query       = $learner_manager->get_enrolled_courses_query( $user_id, $args );
+	$included_course_ids = $courses_query->posts;
```
adamwoodnz: We actually know the ids for the list of Courses here, and don't need to do another query and diff to work out what should be excluded from the query. I tried setting the `post__in` property of the block query to this array, rather than setting the `excludes` property but it didn't work. Could be worth looking further into this though.

adamwoodnz (changes requested):
Works for me with the conditions fixed, see comments inline.

I think the 'excludes' approach does more work than it needs to, but that is an optimization we could look at later if you like.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: d9376a9

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: d9376a9

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: 483e865
Ah, wasn't aware that there's `post__in`. This makes the logic much simpler. Big thanks!

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -561,71 +561,39 @@ function inject_other_filters( $key ) {
...
+ * @return array The modified course query.
  */
-function filter_student_course_list( $parsed_block ) {
+function modify_course_query( $query ) {
 	if (
-		'core/query' !== $parsed_block['blockName'] ||
-		'course' !== ( $parsed_block['attrs']['query']['postType'] ?? '' ) ||
-		get_the_ID() !== Sensei()->settings->get_my_courses_page_id()
+		'course' === $query['post_type'] &&
+		get_the_ID() === Sensei()->settings->get_my_courses_page_id()
```
outdoor2kode: Not sure if you have any idea about this: I originally didn't include this, and on the course archive page, it didn't enter the condition (actually it didn't even enter the function), which I think it should have, as it uses the query loop block as well.

outdoor2kode: This is still needed though because the front page will be affected.

outdoor2kode: And upon further thought, `'course' === $query['post_type']` can probably be removed because the query loop on this page only queries courses, what do you think? (comment also needs to change accordingly)

adamwoodnz: Yeah agreed, I think we can just check the page id

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -557,3 +559,41 @@ function inject_other_filters( $key ) {
...
 }
+
+/**
+ * Modify the course query on the 'My Courses' page to display courses according to the filter status.
+ * Corresponds to https://github.com/Automattic/sensei/blob/trunk/includes/blocks/course-list/class-sensei-course-list-student-course-filter.php#L95
+ *
+ * @param array $query The course query.
+ * @return array The modified course query.
+ */
+function modify_course_query( $query ) {
```
adamwoodnz, line 570: Now that this is just a query modification I think it would be better in `inc/query.php`

adamwoodnz: Actually ignore that, I can see this is closely related to the other filters in here.

adamwoodnz (approved):
LGTM as is, amend the post_type condition if you like

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: fdad165

PR #2740, opened 2024-07-22
Allow selecting a Module for a Lesson without assigning to a Course

### LINKED ISSUE #2681: Feedback - Exit Course link on lesson leads to 404 when parent course is unpublished
## Type of feedback
//dev

## Description
"Exit Course" link on main Lessons pages sometimes works and sometimes 404s. I didn't click through all links, but it looks like "Exit Course" lessons for the Beginner WordPress Developer work and "Exit Course" lessons for Advanced WordPress Developer do not.

Examples where the Exit Course link doesn't work:

- https://learn.wordpress.org/lesson/adding-custom-user-roles-using-a-wordpress-plugin/
- https://learn.wordpress.org/lesson/using-caching-to-improve-website-performance-w3-total-cache-plugin/
- https://learn.wordpress.org/lesson/wordpress-taxonomies-2/
- https://learn.wordpress.org/lesson/creating-custom-post-types-and-taxonomies/

Examples where it does:
- https://learn.wordpress.org/lesson/website-optimization/
- https://learn.wordpress.org/lesson/wordpress-hooks/
- https://learn.wordpress.org/lesson/time-to-test-your-skills/

1. From learn.wordpress.org, click on [Lessons](https://learn.wordpress.org/lessons/) in the Learn navigation (logged in state doesn't matter)
2. Click on the [Adding custom user roles using a WordPress plugin](https://learn.wordpress.org/lesson/adding-custom-user-roles-using-a-wordpress-plugin/
) lesson
3. From top right screen (at least that's the position on desktop), click the [Exit Course](https://learn.wordpress.org/?post_type=course&p=257841) Link
4. You're redirected to a "Sorry, we couldn't find this page" page. Note redirect URL structure: https://learn.wordpress.org/?post_type=course&p=257841

Screenshot 2024-07-16 at 8 14 17 AM

Testing on latest MacOS + Chrome

adamwoodnz:
Fixes #2681 

The idea here is to allow us to build out modules and publish the contained Lessons without a course, so that the parent course does not show up in the UI for the Lesson (course outline, exit course button, breadcrumbs). When all the modules are complete they can be arranged into a course and published.

Screenshot 2024-07-23 at 3 00 06 PM

### Tasks

- [x] Enable selecting any module when there is no course set
- [x] Enable saving the Lesson with a module set and no course set
- [ ] Enable inserting a pre-existing module into a course

### Observations

When a course is selected the selected module is wiped in the UI without a page reload, if it doesn't belong to the course (good). A save is required to persist that state though. If the course is cleared the module selector does not update in reverse without a page reload (bad). More work required if we want that.

The Lesson uses the pre-built standalone lesson styles on the frontend, and looks pretty good; note that it shows the module name, but this isn't a link.

localhost_8888_lesson_lesson-for-unpublished-course_(2 1366x768)

### Conclusion

This approach does require a lot of code, likely more than if we were to detect whether the Lesson's parent Course was unpublished. We are fundamentally changing the intended relationship between modules and courses, so it's hard to say what kind of bugs we could run into. With that said we can do this all with actions and filters rather than implementing logic in php patterns, which is possibly a cleaner approach.

I'm on the fence, but my gut says this is not simple, and will only be worth it if it significantly improves your workflow.

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/sensei.php
+++ b/wp-content/plugins/wporg-learn/inc/sensei.php
@@ -27,6 +27,7 @@ add_action( 'sensei_login_form_before', __NAMESPACE__ . '\sensei_login_form_befo
 add_action( 'sensei_register_form_start', __NAMESPACE__ . '\sensei_register_form_start' );
 // Disable Sensei user login & creation.
 add_filter( 'init', __NAMESPACE__ . '\block_login_register_actions', 1 );
+add_action( 'add_meta_boxes', __NAMESPACE__ . '\maybe_customize_sensei_module_metabox', 20 );
 
 /**
  * Slugs in Sensei are translatable, which won't work for our site and the language switcher.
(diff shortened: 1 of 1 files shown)


adamwoodnz:
Closing in favor of https://github.com/WordPress/Learn/pull/2741

PR #2768, opened 2024-07-24
Fix sticky header position incorrect when on a smaller screen for lesson page

### LINKED ISSUE #2663: Feedback - Sticky header position incorrect on iOS safari
## Type of feedback
//dev

## Description
For the new Learn WordPress site. Breadcrumb header on mobile for individual lesson page (ex https://learn.wordpress.org/lesson/website-optimization/) is sticky at an incorrect position (below top of page)

1. Visit https://learn.wordpress.org/lesson/website-optimization/ on the new Learn WordPress site
2. Scroll down the page

https://github.com/WordPress/Learn/assets/6667837/dc11eaad-d63b-4902-8651-f95a45421d1f

iOS version 17.5.1
Safari for iOS 17.5

outdoor2kode:
Fixes #2663

This issue can also be reproduced on Chrome and Safari.

The root cause is that we fixed the `--sensei-wpadminbar-offset` to `var(--wp-admin--admin-bar--height)`. However, during scrolling, `--sensei-wpadminbar-offset` should actually be 0.

I first tried removing the fixed value here, and it appears that `--sensei-wpadminbar-offset` already can adjust the admin bar height according to different screen sizes. Was there any specific consideration for adding this fixed value initially? It seems safe to remove it now.

I also noticed that the progress bar on the inner lesson page is still blue. I believe we should align the progress bar here with the others? (See screenshots) @fcoveram 

## Screenshots

**Progress Bar**
image
image

**Sticky Header - Safari**
safari

**Sticky Header - Chrome**
chrome

**Sticky Header - Firefox**
ff

**Sticky Header - iPhone**
ios

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
+++ b/wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
@@ -9,7 +9,7 @@ body.sensei {
 	--sensei-button-text-color: var(--wp--preset--color--white);
 
 	--sensei-course-progress-bar-color: var(--wp--custom--color--border);
-	--sensei-course-progress-bar-inner-color: var(--wp--preset--color--blueberry-1);
+	--sensei-course-progress-bar-inner-color: var(--wp--custom--color--green-50);
 	--sensei-lesson-meta-color: var(--wp--preset--color--charcoal-4);
 	--sensei-module-lesson-color: var(--wp--preset--color--charcoal-1);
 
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
Sticky header fix LGTM

Could merge that while we wait for feedback on progress bar color.

adamwoodnz:
I agree though, looking at this:

Screenshot 2024-07-25 at 10 59 45 AM

outdoor2kode:
Yeah, I thought the same. I'll merge it without the color change and let @fcoveram decide.

fcoveram:
Thanks for addressing this. The implementation is consistent with its desktop view, but I'd like to apply some tweaks to add more context to the learning progress message.

The current might confuse visitors with the scroll-down reading progress bar that has become more common in pages with extended content.

PR #2776, opened 2024-07-25
Course outline block header enhance

### LINKED ISSUE #2614: New Theme: Content headers should collapse/open when clicked anywhere.
## Description
We hide content in course _categories_ ([example](https://learn.wordpress.org/course/intermediate-wordpress-user/)) and reveal it on click. However, the click target is very small. I should be able to open the bucket by clicking anywhere in the title.

- Red is current.
- Yellow is suggested.
Screenshot 2024-07-05 at 9 18 36 AM

### LINKED ISSUE #2778: A11y / Sensei / Learn - Section toggle button too small
## Type of feedback
//dev

## Description

The course outline toggle button is 32x8px, which is a very small clickable surface; difficult for users with low motor control. Minimum target size is 24x24 per WCAG 2.5.8, so this is about half the size it needs to be.

image

outdoor2kode:
Resolves #2614 and #2778

**2614**
Originally, the header was directly given a class to allow toggling, but this causes an accessibility issue since the header does not support the `aria-expanded="true"` attribute and also leads to the problem of "interactive controls being nested". So here triggers the button inside the header without allowing it to bubble up.

**2778**

~Doesn't seem to be a problem anymore, couldn't replicate it either on production or locally. I've left a comment on the issue to ask for help to check if the problem still exists.~

The issue can be replicated on Firefox.

## Screenshot

**2614**
scroll

**2778**

image

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
+++ b/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
@@ -8,7 +8,7 @@ document.addEventListener( 'DOMContentLoaded', () => {
 	 * Allow the entire header to toggle, giving users a larger area to interact with.
 	 */
 	document
-		.querySelectorAll( 'section.wp-block-sensei-lms-course-outline-module-bordered >  header' )
+		.querySelectorAll( 'section.wp-block-sensei-lms-course-outline-module-bordered > header' )
 		.forEach( ( header ) => {
 			const button = header.querySelector( 'button' );
 			header.addEventListener( 'click', () => {
--- a/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
+++ b/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
@@ -4,10 +4,20 @@ import { Icon, drafts, lockOutline } from '@wordpress/icons';
 import { renderToString } from '@wordpress/element';
 
 document.addEventListener( 'DOMContentLoaded', () => {
+	/**
+	 * Allow the entire header to toggle, giving users a larger area to interact with.
+	 */
 	document
 		.querySelectorAll( 'section.wp-block-sensei-lms-course-outline-module-bordered >  header' )
 		.forEach( ( header ) => {
-			header.classList.add( 'sensei-collapsible__toggle' );
+			const button = header.querySelector( 'button' );
+			header.addEventListener( 'click', () => {
+				button.click();
+			} );
+
+			button.addEventListener( 'click', ( event ) => {
+				event.stopPropagation();
+			} );
 		} );
 
 	/**
--- a/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
+++ b/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
@@ -27,6 +27,15 @@ document.addEventListener( 'DOMContentLoaded', () => {
 				const heading = link.parentElement;
 				heading.innerHTML = link.innerHTML;
 			}
+			// Also remove the link within the span with class "screen-reader-text" if it exists.
+			const span = header.querySelector( 'span.screen-reader-text' );
+			if ( span ) {
+				const spanLink = span.querySelector( 'a' );
+				if ( spanLink ) {
+					const spanContent = spanLink.innerHTML;
+					span.innerHTML = spanContent;
+				}
+			}
 		} );
 
 	/**
--- a/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
+++ b/wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
@@ -4,6 +4,12 @@ import { Icon, drafts, lockOutline } from '@wordpress/icons';
 import { renderToString } from '@wordpress/element';
 
 document.addEventListener( 'DOMContentLoaded', () => {
+	document
+		.querySelectorAll( 'section.wp-block-sensei-lms-course-outline-module-bordered >  header' )
+		.forEach( ( header ) => {
+			header.classList.add( 'sensei-collapsible__toggle' );
+		} );
+
 	/**
 	 * Find all in progress lessons, and replace the status icon with the Gutenberg-style `drafts` icon.
 	 */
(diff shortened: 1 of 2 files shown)


wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
```diff
@@ -4,10 +4,30 @@ import { Icon, drafts, lockOutline } from '@wordpress/icons';
 import { renderToString } from '@wordpress/element';
 
 document.addEventListener( 'DOMContentLoaded', () => {
+	/**
+	 * Allow the entire header to toggle, giving users a larger area to interact with.
+	 */
+	document
+		.querySelectorAll( 'section.wp-block-sensei-lms-course-outline-module-bordered >  header' )
```
adamwoodnz:
```suggestion
		.querySelectorAll( 'section.wp-block-sensei-lms-course-outline-module-bordered > header' )
```

adamwoodnz (commented):
LGTM, one suggestion.

I did notice while testing that the focus ring on lessons within modules doesn't show properly; you can just see it in the corners:

Screenshot 2024-07-26 at 11 17 11 AM

wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
outdoor2kode: 9a24c88

adamwoodnz:
Can we also fix https://github.com/WordPress/Learn/issues/2778 in this PR?

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -147,17 +148,19 @@ body.sensei {
 				}
 			}
 
-			.wp-block-sensei-lms-course-outline-module__lessons-title {
-				display: none;
-			}
-
 			.wp-block-sensei-lms-collapsible {
+				overflow: unset;
```
outdoor2kode, line 173: > I did notice while testing that the focus ring on lessons within modules doesn't show properly; you can just see it in the corners:

Fixed here.

**Note**

Changing `overflow: hidden` will only impact one scenario: when the window is resized on the desktop, causing text to overflow. However, in practical usage, when users initially access the page with different screen sizes, the height of the lesson items is automatically calculated, preventing any overflow. 

https://github.com/user-attachments/assets/0dd53e35-4feb-4e6f-8436-a74671f7d87d

If someone does this, in the case of an iPhone 5e, most titles will not overflow. Also, the originally hidden text (when `overflow: hidden`) cannot be seen through scrolling anyway, as this aspect is not specifically handled by Sensei.

focus

ryelle:
It looks like some course headers can be links, for example on [the beginner dev course](https://learn.wordpress.org/course/beginner-wordpress-developer/) "[A brief overview of how WordPress works](https://learn.wordpress.org/modules/a-brief-overview-of-how-wordpress-works/?course_id=56187)" is a link, while "The WordPress development stack" is not (seems to be the description, when that exists, it adds a module link to the title). Since there can be links here, the header can't be clickable like this, it conflicts with the link.

Do we need those links? Maybe they could be removed. If they can't, #2614 shouldn't be done, and just address #2778.

outdoor2kode:
> It looks like some course headers can be links, for example on [the beginner dev course](https://learn.wordpress.org/course/beginner-wordpress-developer/) "[A brief overview of how WordPress works](https://learn.wordpress.org/modules/a-brief-overview-of-how-wordpress-works/?course_id=56187)" is a link, while "The WordPress development stack" is not (seems to be the description, when that exists, it adds a module link to the title). Since there can be links here, the header can't be clickable like this, it conflicts with the link.
>
> Do we need those links? Maybe they could be removed. If they can't, https://github.com/WordPress/Learn/issues/2614 shouldn't be done, and just address https://github.com/WordPress/Learn/issues/2778.

Nice catch! Yes, it looks like after adding a description to a module, the header title becomes a link. IMO, I think this link can be removed, as the link redirects to the first lesson, having the same behavior as the first lesson item below it. @kathrynwp @jonathanbossenger @westnz what do you think? Could the link on the header title be removed so that this PR proposal can work correctly?

Note: This PR mainly aims to make the whole header area clickable to collapse/expand the course outline, as shown in the screenshots in the PR description above.

kathrynwp:
> Could the link on the header title be removed so that this PR proposal can work correctly?

As far as I'm concerned, the links on module headings on the course overview page can indeed be removed. I can't think of a good reason for them to stay, especially since they are not consistently present on every module.

@jonathanbossenger @westnz – do you agree? Thanks!

outdoor2kode:
The Firefox focus ring too small issue is fixed in 1deb4e8. Its screenshot is also updated in the PR description.

 efd1847 I have also removed the module heading link for now. If it is not needed, we can revert the commit. The approach should prevent screen reader users from being confused as well.

jonathanbossenger:
> Could the link on the header title be removed so that this PR proposal can work correctly?

I wasn't even aware they were links. Remove away.

wp-content/themes/pub/wporg-learn-2024/src/course-outline/index.js
```diff
@@ -4,10 +4,39 @@ import { Icon, drafts, lockOutline } from '@wordpress/icons';
...
+				event.stopPropagation();
+			} );
+
+			// To enable the entire header to be clickable for toggling without conflicts, remove the link.
+			// In fact, this link duplicates the first lesson link in the course outline.
+			// See https://github.com/WordPress/Learn/pull/2776#issuecomment-2258308422
+			const link = header.querySelector( 'h2 > a' );
+			if ( link ) {
+				const heading = link.parentElement;
+				heading.innerHTML = link.innerHTML;
```
ryelle, line 28: The title used in the screen reader text is also a link, can you replace that as well?

outdoor2kode: Which title you're referring to? I don't seem able to locate it..

image

ryelle: Inside the button, it's invisible text for screen readers.

Screenshot 2024-07-30 at 8 20 50 PM

outdoor2kode: ah.. didn't notice that! 0e8aa4b

outdoor2kode: | Before | After |
|-|-|
| image | image |

ryelle (approved):
Looks good, I've tested with mouse, keyboard-only, and screen reader 👍🏻

PR #2777, opened 2024-07-25
Fix course cards links are duplicated

### LINKED ISSUE #2769: In the course cards, all links are duplicated.
In the course cards, all links are duplicated. 

image

While this doesn't cause any immediate problems, it does mean that the travel path is significantly longer for keyboard and screen reader users, and creates a lot of redundancy in navigating course archives. It would be better if these two elements were combined into a single link & the alt text was empty on the image.

_Originally posted by @joedolson in https://github.com/WordPress/Learn/issues/2482#issuecomment-2249050547_

outdoor2kode:
Fixes #2769

This PR directly adds 'aria-hidden: true' and 'tabindex: -1' to the feature images, allowing screen readers and the tab key to skip the feature image while keeping it clickable with the mouse.

**Testing:**
1. Use a screen reader to test that it does not stop on the feature image on pages where the card course template part or my course page content pattern is used.
2. Also, test that the tab key does not navigate to the feature image.
3. Test if the image is still clickable with the mouse.


wp-content/themes/pub/wporg-learn-2024/src/card-course/index.php
```diff
@@ -0,0 +1,55 @@
...
+	$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
+	$metadata['file'] = $metadata_file;
+
+	$script_handle = register_block_script_handle( $metadata, 'viewScript', 0 );
+
+	// Enqueue the assets when the card course template is on the page.
+	add_action(
+		'render_block_core/template-part',
+		function( $block_content, $block ) use ( $script_handle ) {
+			$slugs = array( 'card-course-h3', 'card-course', 'card-lesson-h3', 'card-lesson', 'card' );
```
adamwoodnz: I think we need `card-lesson-plan` too

wp-content/themes/pub/wporg-learn-2024/src/card-course/view.js
```diff
@@ -0,0 +1,26 @@
...
+	 *
+	 * This function adds an `aria-hidden="true"` and `tabindex="-1"`
+	 * to the post featured image block in the card course template.
+	 */
+	function enhanceCardCourseAccessibility() {
+		document
+			.querySelectorAll( '.wporg-learn-card-grid figure.wp-block-post-featured-image' )
+			.forEach( ( featuredImg ) => {
+				featuredImg.setAttribute( 'aria-hidden', 'true' );
+				featuredImg.setAttribute( 'tabindex', '-1' );
```
adamwoodnz: I think we should only be doing this to the anchor element

wp-content/themes/pub/wporg-learn-2024/src/card-course/block.json
```diff
@@ -0,0 +1,4 @@
+{
+	"name": "wporg-learn/card-course",
```
adamwoodnz: Let's rename all the things to
```suggestion
	"name": "wporg-learn/card-featured-image-a11y",
```

adamwoodnz (commented):
Nice work! Couple of comments inline.

I was going to approach this either by just removing the links from images (would have affected user experience IMO), or using our LinkWrapper block (would have required updates to make it work within a query loop), but I think this is a reasonable solution 👍

wp-content/themes/pub/wporg-learn-2024/src/card-course/view.js
adamwoodnz: We could consider setting `role="presentation"` on the image, as the alt text is the same as the title and therefore it won't be adding any value.

wp-content/themes/pub/wporg-learn-2024/src/card-course/block.json
outdoor2kode: caaab48

wp-content/themes/pub/wporg-learn-2024/src/card-course/index.php
outdoor2kode: I don't see the card-lesson-plan, and I can see that the lesson plan archive is already working now. It doesn't work on your side?

wp-content/themes/pub/wporg-learn-2024/src/card-course/view.js
outdoor2kode: >  think we should only be doing this to the anchor element

Ah, yeah, thanks! 18a01a2

> We could consider setting role="presentation" on the image, as the alt text is the same as the title and therefore it won't be adding any value.

| Before | After |
|-|-|
| Screenshot 2024-07-26 at 08 38 59 | Screenshot 2024-07-26 at 08 46 03 |

wp-content/themes/pub/wporg-learn-2024/src/card-course/index.php
adamwoodnz: Oh yeah I was looking at sandbox, I must have synced a file in there at some point, ignore

wp-content/themes/pub/wporg-learn-2024/src/card-featured-image-a11y/index.php
```diff
@@ -0,0 +1,55 @@
...
+ * `wp_json_file_decode` & `register_block_script_handle.
+ */
+function init() {
+	$metadata_file = dirname( dirname( __DIR__ ) ) . '/build/card-featured-image-a11y/block.json';
+	$metadata = wp_json_file_decode( $metadata_file, array( 'associative' => true ) );
+	$metadata['file'] = $metadata_file;
+
+	$script_handle = register_block_script_handle( $metadata, 'viewScript', 0 );
+
+	// Enqueue the assets when the card course template is on the page.
```
adamwoodnz:
```suggestion
	// Enqueue the assets when one of the card templates is on the page.
```

outdoor2kode: e805a9a

adamwoodnz (approved):
LGTM :shipit:

Issue #2787, opened 2024-07-26
Update search archive page filter labels


Piyopiyo-Kitsune:
At the moment we have two filters which are not immediately obvious what they filter for:

- All -- used to distinguish skill level
- Type -- used to distinguish the content type

I recommend we update these labels to the following for clarity:

- All --> Skill Level
- Type --> Content Type

outdoor2kode:
Related: https://github.com/WordPress/Learn/issues/2784

ryelle:
These have been updated to

- Type: Any
- Level: All

<img width="312" alt="Screenshot 2024-08-01 at 4 08 53 PM" src="https://github.com/user-attachments/assets/147e4316-d879-400e-a778-9aa277521f72">

I went with single words to save space, but if they need the full label, we can reopen this issue.

Issue #2767, opened 2024-07-24
Lessons archive pages - Better distinguish between lessons with the same name


kathrynwp:
## Issue

At the moment, there is no quick way for learners to distinguish between lessons with the same name. There may be lessons with the same name that are part of different courses, or standalone lessons with the same name.

## Details

1. Visit the lessons archive page - https://learn.wordpress.org/lessons/
2. Do a search for a term like `block` - https://learn.wordpress.org/?s=block&post_type=lesson

## Result

In the search results, observe two courses called "Block attributes"

<img width="346" alt="Lessons___Learn_WordPress" src="https://github.com/user-attachments/assets/2abcf92d-fc21-426f-8814-0d063668b2f9">

There's no way for a learner to distinguish between the two – without clicking into it and seeing the course that it's part of.

## Improvement

Could we display the name of the course to which the lesson belongs, or think of another way to distinguish among lessons with the same name? In the example above, even if we added a level to the second course, it would also be Beginner, so that wouldn't be enough.

Example of what it could look like:

<img width="348" alt="Lessons___Learn_WordPress" src="https://github.com/user-attachments/assets/bb7c1b0e-dd74-47a7-8d79-9a63e86d8b90">

adamwoodnz:
@WordPress/meta-design your input please

jasmussen:
It seems useful to include the course taxonomy, in this case "Beginner WordPress Developer", to differentiate, for contexts like these.

I'd prefer to defer to Francisco/Marko on this one, they have some context for these card designs that I do not. Such as, perhaps in most cases, these cards are shown in context of a course, and therefore it's only ever on _search results_ that you'd see them atomically split out. If that's the case, then I'd think we should only include this taxonomy on search results, so as to not make the cards uniformly more verbose. 

If we do that, perhaps the permalink could have this format: "[Course]: [Lesson]". I.e. "Beginner WordPress Developer: Block Attributes".

Another path to explore is to segment the search results into subheadings. Pseudo code:

```
<h3>Beginner WordPress Developer course</h3>

[Lesson cards]

<h3>Developing your first WordPress block</h3>

[Lesson cards]
```

---

On a higher level, there's an aspect to naming and curation here, where it would be useful for lesson authors to be mindful of this particular SEO, and perhaps try to search and ambiguate when a match is found, before they name lessons the same. 

Additionally, we should be uniform in how text-case is applied. Is it "Block attributes" or "Block Attributes"?

[...]

Issue #2690, opened 2024-07-17
Feedback - Consider adding the Audience tag


westnz:
## Type of feedback

//dev

## Description

Adding the Audience tag to lesson cards

## Step-by-step reproduction instructions (optional)

1. Go to https://learn.wordpress.org/?new-theme=1
2. Scroll down to featured lessons
3.

<img width="1646" alt="Screenshot 2024-07-18 at 9 17 58 AM" src="https://github.com/user-attachments/assets/d82a05c3-9c18-464b-9ab5-cf7dcfe80121">

## Other information (optional)

When you look at a card, it only states the level.  Would it not be more helpful to learners to add the audience tag?  Then a user, for example, could immediately see if a lesson is meant for a developer or contributor, etc. I realize we need to ensure that the audience is set. Is there a way to bulk select this on courses? . <img width="1853" alt="Screenshot 2024-07-18 at 9 20 58 AM" src="https://github.com/user-attachments/assets/4f133d79-739c-41d3-843b-e44bfd4f98a2">

adamwoodnz:
@WordPress/meta-design seeking your input on this please.

One thing to consider is that Audience is a multi select field, so we may need to display multiple audiences on a card.

Level is too, but it seems less likely we'd select multiple values for that.

westnz:
Thanks Adam. In general, we don't select multiple audiences for our content.

marko-srb:
Hey hey @adamwoodnz ! 

As I see there are limitations to what we can do here. 

1st point. I see old design (left card below) had duration of a course in the mockup. I couldn't follow decisions for its removal, but hope that removal was mutually agreed by parties involved. 

2snd point. If my assumption under 1st point is correct then we can add only competency/experience levels as tags (see right card below). Adding audience tag would mess things up (see second screenshot).

—

Based on those observations, I'd leave only experience levels for the launch, while leaving this issue open for discussing 'audience' tag after the launch...


<img width="1039" alt="Screen Shot 2024-07-25 at 23 29 48" src="https://github.com/user-attachments/assets/3a948474-a4e7-44fa-8e33-c867d1c5a412">

—

<img width="461" alt="Screen Shot 2024-07-25 at 23 32 11" src="https://github.com/user-attachments/assets/bb1527dc-edfc-4d76-8000-338ed070d817">

kathrynwp:
[...]

PR #2801, opened 2024-07-30
Add Search bar to the homepage

### LINKED ISSUE #2744: Add Search box to Learn homepage
### Issue

When a learner visits the homepage at https://learn.wordpress.org there's currently nowhere to do a search

### Improvement

As a learner, I want to be able to search for a term like "block theme" or "plugin development" to see if there are any relevant resources. 

I want a quick and simple way to do a find the content I'm looking for – without needing to drill down manually into courses or lessons, because there's no way for me to know where the content I'm looking for might live.

Put simply: I need a search box that will search Learn.WordPress.org. 

### Other

I'm not sure what others think, but this seems like a clear omission. I noticed that in version 1 of the mockups, there was a prominent search on the homepage, so it must have been removed at some point.

I set the milestone to post-launch, but I do think it warrants another look and strong consideration.

outdoor2kode:
Resolves #2744

As the screencast shows, when searching on the homepage, it goes to the newly added template - search-front-page. When switching to another content type, "course" or "lesson", the behavior remains as is - stay in the archive-course, or archive-lesson templates, displaying their own dedicated UI and filters. And it switches back to the search-front-page template when the content type "Any" is chosen.

The unclear and confusing wording of the filter mentioned in #2787 will be addressed in another PR.

Please also feel free to provide feedback on the flow or design. For example like including a label somewhere on the card to allow users to know whether they are looking at a Course or a Lesson, or do we still want the big text on top of the search page when there's already a text functioning similarly under the search bar. @ndiego @kathrynwp @WordPress/meta-design 

| Big text | Text under the search bar |
|-|-|
| image | image |

## Testing

1. Search on the homepage.
2. Play around with various filters.
3. Search on the Course / Lesson / Learning pathway archive page, and the results should be on the same page (ie, the header, hero section as well as the filters should still be there)

## Screencast

https://github.com/user-attachments/assets/9b88c254-1017-4a1e-859f-289373bb4388

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2024/functions.php
@@ -65,7 +65,7 @@ function modify_single_template( $templates ) {
 }
 
 /**
- * Modify the search template hierarchy to use search-front-page templates.
+ * Modify the search template hierarchy to use search-all templates.
  *
  * @param array $templates Array of template files.
  * @return array
@@ -79,7 +79,7 @@ function modify_search_template( $templates ) {
 		! ( is_post_type_archive( 'course' ) || is_post_type_archive( 'lesson' ) ) &&
 		! is_tax( 'learning-pathway' )
 		) {
-			array_unshift( $templates, 'search-front-page' );
+			array_unshift( $templates, 'search-all' );
 	}
 
 	return $templates;
--- a/wp-content/themes/pub/wporg-learn-2024/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2024/functions.php
@@ -39,6 +39,7 @@ add_filter( 'sensei_register_post_type_lesson', function( $args ) {
 	return $args;
 } );
 add_filter( 'single_template_hierarchy', __NAMESPACE__ . '\modify_single_template' );
+add_filter( 'search_template_hierarchy', __NAMESPACE__ . '\modify_search_template' );
 add_filter( 'wporg_block_navigation_menus', __NAMESPACE__ . '\add_site_navigation_menus' );
 add_filter( 'wporg_block_site_breadcrumbs', __NAMESPACE__ . '\set_site_breadcrumbs' );
 add_filter( 'taxonomy_template_hierarchy', __NAMESPACE__ . '\modify_taxonomy_template_hierarchy' );
@@ -58,6 +59,27 @@ function modify_single_template( $templates ) {
 		array_unshift( $templates, 'single-lesson.html' );
 	} elseif ( is_singular( 'quiz' ) ) {
 		array_unshift( $templates, 'single-quiz.html' );
+	}
+
+	return $templates;
+}
+
+/**
+ * Modify the search template hierarchy to use search-front-page templates.
+ *
+ * @param array $templates Array of template files.
+ * @return array
+ */
+function modify_search_template( $templates ) {
+	// Should not change the search result template of course, lesson, and learning-pathway.
+	// Currently, they each use their specific templates: archive-course, archive-lesson, and taxonomy-learning-pathway,
+	// which have their own dedicated UI and filters.
+	if (
+		is_search() &&
+		! ( is_post_type_archive( 'course' ) || is_post_type_archive( 'lesson' ) ) &&
+		! is_tax( 'learning-pathway' )
+		) {
+			array_unshift( $templates, 'search-front-page' );
 	}
 
 	return $templates;
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -55,7 +55,7 @@ function get_content_type_options( $options ) {
 	);
 
 	$selected_slug = $wp_query->get( 'post_type' ) ? $wp_query->get( 'post_type' ) : 'any';
-	$label = $options[ $selected_slug ];
+	$label = $options[ $selected_slug ] ?? $options['any'];
 
 	return array(
 		'label' => $label,
@@ -531,7 +531,7 @@ function get_student_course_options( $options ) {
 	);
 
 	$selected_slug = $wp_query->get( $key ) ? $wp_query->get( $key ) : 'all';
-	$label = $options[ $selected_slug ];
+	$label = $options[ $selected_slug ] ?? $options['all'];
 
 	return array(
 		'label' => $label,
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -27,21 +27,14 @@ add_action( 'wporg_query_filter_in_form', __NAMESPACE__ . '\inject_other_filters
 add_filter( 'query_loop_block_query_vars', __NAMESPACE__ . '\modify_course_query' );
 
 /**
- * Get the current URL.
+ * Get the filtered URL by removing the page query var.
+ * If the path retains "page," the filtered result might be a 404 page.
  *
- * @return string The current URL.
+ * @return string The filtered URL.
  */
 function get_filtered_url() {
 	global $wp;
-
-	$current_url = home_url( add_query_arg( array(), $wp->request ) );
-
-	// Remove the page query var.
-	// If the path retains "page," the filtered result might be a 404 page.
-	$parsed_url = wp_parse_url( $current_url );
-	$path = isset( $parsed_url['path'] ) ? $parsed_url['path'] : '';
-	$filtered_path = preg_replace( '#/page/\d+/?$#', '', $path );
-
+	$filtered_path = preg_replace( '#page/\d+/?$#', '', $wp->request );
 	return home_url( $filtered_path );
 }
 
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -27,6 +27,25 @@ add_action( 'wporg_query_filter_in_form', __NAMESPACE__ . '\inject_other_filters
 add_filter( 'query_loop_block_query_vars', __NAMESPACE__ . '\modify_course_query' );
 
 /**
+ * Get the current URL.
+ *
+ * @return string The current URL.
+ */
+function get_filtered_url() {
+	global $wp;
+
+	$current_url = home_url( add_query_arg( array(), $wp->request ) );
+
+	// Remove the page query var.
+	// If the path retains "page," the filtered result might be a 404 page.
+	$parsed_url = wp_parse_url( $current_url );
+	$path = isset( $parsed_url['path'] ) ? $parsed_url['path'] : '';
+	$filtered_path = preg_replace( '#/page/\d+/?$#', '', $path );
+
+	return home_url() . $filtered_path;
+}
+
+/**
  * Get the content type options.
  * Used for the search filters and the archive filters.
  *
@@ -68,7 +87,7 @@ function create_content_type_options( $content_types ) {
 		'label' => $label,
 		'title' => __( 'Content Type', 'wporg-learn' ),
 		'key' => 'post_type',
-		'action' => home_url(),
+		'action' => get_filtered_url(),
 		'options' => $options,
 		'selected' => array( $selected_content_type ),
 	);
@@ -131,7 +150,7 @@ function create_level_options( $levels ) {
 		'label' => $label,
 		'title' => __( 'Level', 'wporg-learn' ),
 		'key' => 'wporg_lesson_level',
-		'action' => home_url(),
+		'action' => get_filtered_url(),
 		'options' => array_combine( wp_list_pluck( $levels, 'slug' ), wp_list_pluck( $levels, 'name' ) ),
 		'selected' => array( $selected_slug ),
 	);
@@ -273,7 +292,7 @@ function create_topic_options( $topics ) {
 		'label' => $label,
 		'title' => __( 'Filter', 'wporg-learn' ),
 		'key' => 'wporg_workshop_topic',
-		'action' => home_url(),
+		'action' => get_filtered_url(),
 		'options' => array_combine( wp_list_pluck( $topics, 'slug' ), wp_list_pluck( $topics, 'name' ) ),
 		'selected' => $selected,
 	);
@@ -454,7 +473,7 @@ function create_language_options( $languages ) {
 		'label' => $label,
 		'title' => __( 'Filter', 'wporg-learn' ),
 		'key' => 'language',
-		'action' => home_url(),
+		'action' => get_filtered_url(),
 		'options' => $languages,
 		'selected' => $selected,
 	);
@@ -559,7 +578,7 @@ function get_student_course_options( $options ) {
 		'label' => $label,
 		'title' => __( 'Completion status', 'wporg-learn' ),
 		'key' => $key,
-		'action' => home_url(),
+		'action' => get_filtered_url(),
 		'options' => $options,
 		'selected' => array( $selected_slug ),
 	);
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -8,6 +8,8 @@ namespace WordPressdotorg\Theme\Learn_2024\Block_Config;
 use function WPOrg_Learn\Post_Meta\{get_available_post_type_locales};
 use Sensei_Learner;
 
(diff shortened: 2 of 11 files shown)


wp-content/themes/pub/wporg-learn-2024/patterns/front-page-content.php
```diff
@@ -7,6 +7,16 @@
 
 ?>
 
+<!-- wp:group {"layout":{"type":"flex","flexWrap":"wrap"}} -->
+<div class="wp-block-group">
+	<!-- wp:search {"showLabel":false,"placeholder":"<?php esc_html_e( 'Search for resources', 'wporg-learn' ); ?>","width":100,"widthUnit":"%","buttonText":"<?php esc_html_e( 'Search', 'wporg-learn' ); ?>","buttonPosition":"button-inside","buttonUseIcon":true} /-->
```
outdoor2kode, line 12: Is "Search for resources" the language we are expecting finally? cc @ndiego

wp-content/themes/pub/wporg-learn-2024/patterns/search-content.php
```diff
@@ -0,0 +1,53 @@
...
+ * Slug: wporg-learn-2024/search-content
+ * Inserter: no
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"},"style":{"spacing":{"margin":{"top":"0","bottom":"var:preset|spacing|50"}}}} -->
+<div class="wp-block-group alignwide" style="margin-top:0;margin-bottom:var(--wp--preset--spacing--50)">
+
+	<!-- wp:search {"label":"<?php esc_attr_e( 'Search', 'wporg-learn' ); ?>","showLabel":false,"placeholder":"<?php esc_attr_e( 'Search for resources', 'wporg-learn' ); ?>","width":290,"widthUnit":"px","buttonText":"<?php esc_attr_e( 'Search', 'wporg-learn' ); ?>","buttonPosition":"button-inside","buttonUseIcon":true} /-->
```
outdoor2kode: "Search for resources"?

wp-content/themes/pub/wporg-learn-2024/functions.php
```diff
@@ -60,6 +61,27 @@ function modify_single_template( $templates ) {
...
+function modify_search_template( $templates ) {
+	// Should not change the search result template of course, lesson, and learning-pathway.
+	// Currently, they each use their specific templates: archive-course, archive-lesson, and taxonomy-learning-pathway,
+	// which have their own dedicated UI and filters.
+	if (
+		is_search() &&
+		! ( is_post_type_archive( 'course' ) || is_post_type_archive( 'lesson' ) ) &&
+		! is_tax( 'learning-pathway' )
+		) {
+			array_unshift( $templates, 'search-front-page' );
```
outdoor2kode: Is there a better name for [this template](https://github.com/WordPress/Learn/blob/c66d3e6da5704ac016c8a0de4682995aa0506ce5/wp-content/themes/pub/wporg-learn-2024/templates/search-front-page.html)?

ndiego: What about `search-all`?

ndiego:
Thanks @renintw. Regarding the placement of the Search on the Homepage, can we mirror the placement on Forums exactly? This will provide consistency.  In the video, it looks like there is a bit of extra padding/margin at the top. 

image

outdoor2kode:
> can we mirror the placement on Forums exactly

Oh, no problem, I was referring to [the Showcase](https://wordpress.org/showcase/), will change that. And it appears that [Plugins](https://wordpress.org/plugins/) also needs an update.

ndiego:
> Oh, no problem, I was referring to [the Showcase](https://wordpress.org/showcase/), will change that. And it appears that [Plugins](https://wordpress.org/plugins/) also needs an update.

Hmm, you are right. It looks like the search field on pages in the Learn category on WordPress.org is in a slightly different spot than those under Extend and Showcase. My recommendation is we update it so it's similar to Forums and then address post launch if we want to make everything consistent. 

> do we still want the big text on top of the search page when there's already a text functioning similarly under the search bar.

I like the text underneath the search field, but I do think the page needs a title. What if we went with "Search Learn WordPress" as a static title?

wp-content/themes/pub/wporg-learn-2024/patterns/front-page-content.php
ndiego: I like "Search for resources", what do you think @kathrynwp?

kathrynwp:
> I like "Search for resources", what do you think @kathrynwp?

Agreed, I like it!

kathrynwp:
> What if we went with "Search Learn WordPress" as a static title?

I like this for a static title, too.

ndiego:
The only thing I am unsure about is whether we need some sort of label distinguishing between courses and lessons. Visually, I think it's relatively clear which one is a course, but less clear that a lesson is a lesson based on the visuals, if that makes sense. I personally would be fine leaving it the way it is, but I could easily be swayed in the other direction. 😅

| Course | Lesson |
|-|-|
|image|image|

wp-content/themes/pub/wporg-learn-2024/functions.php
outdoor2kode: Thanks! 935d259

kathrynwp:
@ndiego Definitely see your point. Personally I think this should wait til after launch, as there are a few related issues that relate to distinguishing between:

- different audiences - see: #2690
- lessons with the same name / what course a lesson belongs to - see: #2767

...and they should probably all be dealt with holistically.

outdoor2kode:
> can we mirror the placement on Forums exactly

39b64fa

| Learn | Forum |
|-|-|
| Screenshot 2024-07-31 at 05 33 36 | Screenshot 2024-07-31 at 05 34 02 |

> > What if we went with "Search Learn WordPress" as a static title?
>
> I like this for a static title, too.

3a60523 

Version A or B? Version A has the same gap as the homepage. Version B has the same gap as the course archive. The current implementation is version B.

| Version A | Version B | Course Archive |
|-|-|-|
| Screenshot 2024-07-31 at 05 38 01 | Screenshot 2024-07-31 at 05 36 59 | image |

ndiego:
> Version A or B? Version A has the same gap as the homepage. Version B has the same gap as the course archive.

The homepage looks good, and I would go with Version B.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -25,13 +27,51 @@
...
- * Get the current URL.
+ * Get the content type options.
+ * Used for the search filters and the archive filters.
+ *
+ * @param array $options The options for this filter.
+ * @return array New list of custom content type options.
+ */
+function get_content_type_options( $options ) {
+	// Define your custom post types
+	$content_types = array( 'any', 'course', 'lesson' );
```
ryelle: Should these be translatable?

wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
```diff
@@ -28,6 +28,7 @@
 
 	<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"flex","flexWrap":"nowrap"},"className":"wporg-query-filters"} -->
 	<div class="wp-block-group wporg-query-filters">
+		<!-- wp:wporg/query-filter {"key":"content_type","multiple":false} /-->
```
ryelle: Do we really need the content type dropdown on Lesson & Course archives?

Screenshot 2024-07-30 at 5 13 19 PM

wp-content/themes/pub/wporg-learn-2024/patterns/search-all-content.php
```diff
@@ -0,0 +1,63 @@
...
+ * Inserter: no
+ */
+
+?>
+
+<!-- wp:group {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained","justifyContent":"left","contentSize":"750px"}} -->
+<div class="wp-block-group" style="margin-bottom:var(--wp--preset--spacing--50)">
+
+	<!-- wp:heading {"level":1} -->
+	<h1 class="wp-block-heading"><?php esc_html_e( 'Search Learn WordPress', 'wporg-learn' ); ?></h1>
```
ryelle: I know @ndiego and @kathrynwp agreed on "Search Learn WordPress" but that sounds like an action to me, like it's the heading on a search form. Most other pages use "Search results" (at least as a hidden `h1`) for the results page.

ryelle (changes requested):
The search is working, but I had a few other comments about some details.

wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
outdoor2kode: Good question. I added it because if selecting a course or lesson from the search all page, without this filter, you can't go back from the course or lesson archive. Seems to be somewhat inconsistent in the searching process starting from homepage.

ryelle: Ah, I see, if you do the search first and then filter by type, it moves over to the archive template for those search results.

IMO, people can use the back button to get back to all search results if they filter the post type— it's more confusing to me just landing on `/lessons/` and seeing the content type dropdown, since it brings you to a totally different page (also with different filtering options) if you use it.

Open to other opinions though, cc @ndiego @kathrynwp (not sure who else has input here)

wp-content/themes/pub/wporg-learn-2024/patterns/search-all-content.php
outdoor2kode: Should we just change to "Search Results"?

ndiego: That works for me. Mainly I just wanted some sort of title

outdoor2kode: 529fd7b
image

wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
outdoor2kode: I think both options are fine with me. What might ultimately influence my decision is which path the users use most frequently. If most users explore the site from the homepage search, then having a content type filter on the `/lessons/` or `/courses/` pages would provide a better experience IMO.

Alternatively, we could consider showing the filter on the `/lessons/` or `/courses/` pages only when users use the content type filter from the all search results page and are redirected to those pages. If they visit those pages directly, the filter would not be displayed.

I've removed the filter from the lesson and course archive for now. Let me know if we want it back.
b6be6ea

wp-content/themes/pub/wporg-learn-2024/patterns/search-all-content.php
ndiego: Looks good to me.

wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
ndiego: > IMO, people can use the back button to get back to all search results if they filter the post type— it's more confusing to me just landing on /lessons/ and seeing the content type dropdown, since it brings you to a totally different page (also with different filtering options) if you use it.

That's a good point. I have no strong opinions though. I found it nice that I could switch over to Courses from Lessons, but I'm not a typical user.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: yeah, missed that. Thanks! 040de3e
This commit also extracts duplicate parts into a function.

kathrynwp:
> Do we really need the content type dropdown on Lesson & Course archives?

I think this could be a useful toggle to have, although not strictly necessary.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -29,9 +31,47 @@
...
+	global $wp_query;
+
+	$options = array(
+		'any' => __( 'Any', 'wporg-learn' ),
+		'course' => __( 'Course', 'wporg-learn' ),
+		'lesson' => __( 'Lesson', 'wporg-learn' ),
+	);
+
+	$selected_slug = $wp_query->get( 'post_type' ) ? $wp_query->get( 'post_type' ) : 'any';
+	$label = get_label_by_slug( $selected_slug, $options );
```
ryelle: Since these are keyed by the slug, do you need an extra function? I think this should work for both here and `get_student_course_options` (though the fallback there should be `all`).

```suggestion
	$label = $options[ $selected_slug ] ?? $options['any'];
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -29,9 +31,47 @@
...
+
+	$current_url = home_url( add_query_arg( array(), $wp->request ) );
+
+	// Remove the page query var.
+	// If the path retains "page," the filtered result might be a 404 page.
+	$parsed_url = wp_parse_url( $current_url );
+	$path = isset( $parsed_url['path'] ) ? $parsed_url['path'] : '';
+	$filtered_path = preg_replace( '#/page/\d+/?$#', '', $path );
+
+	return home_url() . $filtered_path;
```
ryelle: You can pass the filtered_path to home_url as a parameter, it [accepts a path parameter](https://developer.wordpress.org/reference/functions/home_url/).

```suggestion
	return home_url( $filtered_path );
```

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -29,9 +31,47 @@
...
+function get_filtered_url() {
 	global $wp;
-	return home_url( add_query_arg( array(), $wp->request ) );
+
+	$current_url = home_url( add_query_arg( array(), $wp->request ) );
+
+	// Remove the page query var.
+	// If the path retains "page," the filtered result might be a 404 page.
+	$parsed_url = wp_parse_url( $current_url );
+	$path = isset( $parsed_url['path'] ) ? $parsed_url['path'] : '';
```
ryelle: I believe `$wp->request` will be just the path, no query args, so I don't think you need to do this extra parsing.

wp-content/themes/pub/wporg-learn-2024/inc/block-hooks.php
```diff
@@ -15,6 +15,8 @@
 
 /**
  * Update header template based on current query.
+ * Since the search results for courses and lessons still use their respective archive templates,
+ * we need to update the header template part to display the correct title.
```
ryelle, line 19: Documentation improvements like this could be committed right to trunk IMO, to avoid extra changes in PRs like this. No change requested for this PR, just a suggestion for the future 🙂

ryelle (commented):
Looking good! I still have a few minor code improvement suggestions around the filter code.

wp-content/themes/pub/wporg-learn-2024/inc/block-hooks.php
outdoor2kode: Noted, thanks!

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: bccee01

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: Thanks! 999b6c2

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
outdoor2kode: yeah, you're exactly right, those logic are unnecessary..
I removed the `?? $options['any'];` as `$options[ $selected_slug ]` will always have a value.
18c28a6

ryelle: It will, but it could be something not in the list — if someone manually types in a post type, for example. Adding the fallback works as a guard, just in case.

outdoor2kode: ah, yes. Missed the fact that it fetches the value from the URL. ae09e8e

ryelle (approved):
Great, I think this is good to merge now 👍🏻

PR #2802, opened 2024-07-30
Sensei Quizzes: Accessibility fixes for questions

### LINKED ISSUE #2779: A11y / Sensei / Learn - Quiz questions not associated with answer options
## Type of feedback
//dev

## Description

The input type for multiple choice questions are a radio group, but there is no semantic grouping to associate the question with the answers. For a screen reader user navigating using tab, this means that extra exploration is required to find out what question you're answering. 

The structure should be either a `fieldset` where the question is the `legend` value, or a group with role `radiogroup` and an `aria-labelledby` attribute pointing to the ID of the question.

image

ryelle:
Fixes #2779 — This adds a `fieldset` wrapper for all questions, and fixes the styles on multiple choice checkboxes to enable keyboard selection. The `fieldset` uses the question title text for a screen reader only label. This is also applied to all questions, regardless of type, but it seems like the vast majority of quizzes use multiple choice or true/false.

**Screenshots**

The multiple-choice (checkbox) field now has keyboard-nav support, and a new focus style (matching the button styles). There are no changes from the fieldset wrapper.

| Before | After |
|---|---|
| Screenshot 2024-07-30 at 4 53 08 PM | Screenshot 2024-07-30 at 4 56 19 PM |

The single-choice (radio) field also has the new focus style. There are no changes from the fieldset wrapper.

| Before | After |
|---|---|
| Screenshot 2024-07-30 at 4 53 46 PM | Screenshot 2024-07-30 at 4 56 01 PM |

<details>
<summary>Example of the new markup</summary>

```html
<li class="sensei-quiz-question wp-block-sensei-lms-quiz-question wp-block-sensei-lms-quiz-question boolean" value="2">
	<div class="sensei-lms-question-block__header">
		<h2 class="question question-title">2. Some security plugins include two-factor authentication.</h2>
		<span class="grade">1</span>
	</div>
	<div class="wp-block-sensei-lms-question-description"></div>
	<fieldset>
		<legend class="screen-reader-text">Some security plugins include two-factor authentication.</legend>
		<div class="wp-block-sensei-lms-question-answers">
			<ul class="answers options">
				<li class="">
					<input type="radio" id="question_23820-option-true" name="sensei_question[23820]" value="true" class="sensei-multiple-choice-answer-option-radio">
					<label for="question_23820-option-true">True</label>
				</li>	
				<li class="">
					<input type="radio" id="question_23820-option-false" name="sensei_question[23820]" value="false" class="sensei-multiple-choice-answer-option-radio">
					<label for="question_23820-option-false">False</label>
				</li>
			</ul>
		</div>
	</fieldset>
</li>
```

</details>

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2024/functions.php
@@ -44,6 +44,9 @@ add_filter( 'wporg_block_navigation_menus', __NAMESPACE__ . '\add_site_navigatio
 add_filter( 'wporg_block_site_breadcrumbs', __NAMESPACE__ . '\set_site_breadcrumbs' );
 add_filter( 'taxonomy_template_hierarchy', __NAMESPACE__ . '\modify_taxonomy_template_hierarchy' );
 
+// Attached at 50 to inject after title, description, etc, so that only answers are in the fieldset.
+add_action( 'sensei_quiz_question_inside_before', __NAMESPACE__ . '\sensei_question_add_opening_fieldset', 50 );
+add_action( 'sensei_quiz_question_inside_after', __NAMESPACE__ . '\sensei_question_add_closing_fieldset' );
 add_filter( 'sensei_learning_mode_lesson_status_icon', __NAMESPACE__ . '\modify_lesson_status_icon_add_aria', 10, 2 );
 
 remove_filter( 'template_include', array( 'Sensei_Templates', 'template_loader' ), 10, 1 );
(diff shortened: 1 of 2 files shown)


outdoor2kode (approved):
Looks good! Works as described 👍 

Only one thing, does it make sense to add text to the point on the right side for the screen reader? Currently it only reads the number, which seems a bit ambiguous.
image

ryelle:
> Only one thing, does it make sense to add text to the point on the right side for the screen reader? Currently it only reads the number, which seems a bit ambiguous.

Oh I forgot to mention that— showing points & the format of them is a UI setting. It looks like production doesn't show them now, maybe that was disabled? If they are shown, they should use one of the "with text" settings. @kathrynwp, do you know if you all want to have points visible on the quizzes?

Screenshot 2024-08-01 at 11 23 46 AM

kathrynwp:
Hi @ryelle - I'm not sure what the points signify and whether they are meant to be displayed, but I think we can leave them hidden for now. When @jonathanbossenger and @westnz are available, one of them can confirm that this is correct. Thank you!

ryelle:
In that case, I'll rebase and merge this, and if we need to address the points differently later, we can do that post-launch.

jonathanbossenger:
I'm hopping in to add that our quizzes are mostly like an episode of Whose Line, and the points don't really matter 😁. Usually, our quiz lessons are only worth one point, and we don't really keep track of quiz pass rates anyway.

PR #2804, opened 2024-07-31
Filters: Add prefix label to the single query filters

### LINKED ISSUE #2784: A11y / Course Filter / Learn - Toggle to open filters does not have a useful accessible name
## Type of feedback
//dev

## Description

The toggle that opens the course status filters is labeled using the currently selected status. Buttons to perform an action should be labeled using information about what the control does. Having the current status visible is OK; but it shouldn't be the only content.

This would be better labeled using something like "Filter Courses: All" or even "Filter: All". This would inform the user what they can do with the control. 

At https://learn.wordpress.org/my-courses/

image

### LINKED ISSUE #2787: Update search archive page filter labels

ryelle:
Fixes #2784 - this prefixes the single select filters with the filter name, to give it a more accessible name.

- Level filter is prefixed with "Level: ", ex "Level: Advanced"
- Content type filter is prefixed with "Type: ", ex "Type: Course"
- Completion status is prefixed with "Status: ", ex "Status: Completed"

cc @ndiego @kathrynwp Do those prefixes make sense? Using the full title for the last two would be too long, IMO, so I went with a short version.

**Screenshots:**

[Learning pathway](https://learn.wordpress.org/learning-pathway/developer/)

| Before | After |
|---|---|
| Screen Shot 2024-07-31 at 13 48 08 | Screen Shot 2024-07-31 at 13 42 41 |

[Lesson archive (with "Advanced" filter)](https://learn.wordpress.org/lessons/?post_type=lesson&wporg_lesson_level=advanced)

| Before | After |
|---|---|
| Screen Shot 2024-07-31 at 13 48 26 | Screen Shot 2024-07-31 at 13 43 48 |

[Search results](https://learn.wordpress.org/?s=wordpress)

| Before | After |
|---|---|
| Screen Shot 2024-07-31 at 13 48 45 | Screen Shot 2024-07-31 at 13 45 02 |

[My courses](https://learn.wordpress.org/my-courses/)

| Before | After |
|---|---|
| Screen Shot 2024-07-31 at 13 52 41 | Screen Shot 2024-07-31 at 13 52 54 |

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -59,7 +59,7 @@ function get_content_type_options( $options ) {
 
 	return array(
 		'label' => sprintf( __( 'Type: %s', 'wporg-learn' ), $label ),
-		'title' => __( 'Content Type', 'wporg-learn' ),
+		'title' => __( 'Content type', 'wporg-learn' ),
 		'key' => 'post_type',
 		'action' => get_filtered_url(),
 		'options' => $options,
@@ -100,24 +100,18 @@ function create_level_options( $levels ) {
 		$levels,
 	);
 
-	$label = __( 'Level', 'wporg-learn' );
+	$label = __( 'All', 'wporg-learn' );
 
 	$selected_slug = $wp_query->get( 'wporg_lesson_level' );
 	if ( $selected_slug ) {
 		// Find the selected level from $levels by slug and then get the name.
-		$selected_level = array_filter(
-			$levels,
-			function ( $level ) use ( $selected_slug ) {
-				return $level->slug === $selected_slug;
-			}
-		);
+		$selected_level = wp_list_filter( $levels, array( 'slug' => $selected_slug ) );
 		if ( ! empty( $selected_level ) ) {
 			$selected_level = array_shift( $selected_level );
 			$label = $selected_level->name;
 		}
 	} else {
 		$selected_slug = 'all';
-		$label = __( 'All', 'wporg-learn' );
 	}
 
 	return array(
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -58,7 +58,7 @@ function get_content_type_options( $options ) {
 	$label = $options[ $selected_slug ] ?? $options['any'];
 
 	return array(
-		'label' => $label,
+		'label' => sprintf( __( 'Type: %s', 'wporg-learn' ), $label ),
 		'title' => __( 'Content Type', 'wporg-learn' ),
 		'key' => 'post_type',
 		'action' => get_filtered_url(),
@@ -121,7 +121,7 @@ function create_level_options( $levels ) {
 	}
 
 	return array(
-		'label' => $label,
+		'label' => sprintf( __( 'Level: %s', 'wporg-learn' ), $label ),
 		'title' => __( 'Level', 'wporg-learn' ),
 		'key' => 'wporg_lesson_level',
 		'action' => get_filtered_url(),
@@ -534,7 +534,7 @@ function get_student_course_options( $options ) {
 	$label = $options[ $selected_slug ] ?? $options['all'];
 
 	return array(
-		'label' => $label,
+		'label' => sprintf( __( 'Status: %s', 'wporg-learn' ), $label ),
 		'title' => __( 'Completion status', 'wporg-learn' ),
 		'key' => $key,
 		'action' => get_filtered_url(),


ndiego:
Visually, it looks great. Thanks @ryelle 🚢

outdoor2kode:
Ah, I was also working on the same stuff but a different [ticket](https://github.com/WordPress/Learn/issues/2787).

This is the suggested label update mentioned there:

```
I recommend we update these labels to the following for clarity:

All --> Skill Level
Type --> Content Type
```

[Option 1 - Fix the label](https://github.com/WordPress/Learn/commit/2a8d3cea0ec07f3ac7655148c4c15119f609cd9b)

| Content Type | Skill Level |
|-|-|
| Screenshot 2024-08-01 at 02 33 52 | Screenshot 2024-08-01 at 02 34 48 |


[Option 2 - Same as this PR](https://github.com/WordPress/Learn/commit/23f1d20d93b4b5e3edb5df01909b21f880e74c30)
| Content Type | Skill Level |
|-|-|
| Screenshot 2024-08-01 at 02 47 31 | Screenshot 2024-08-01 at 02 52 03 |

Option 2 looks better to me.

**Q: Do we consider removing the label on the dropdown? Maybe it's fine for now and can leave it to post-launch.**
Markup on 2024-08-01 at 02:34:16

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -534,7 +534,7 @@ function get_student_course_options( $options ) {
 	$label = $options[ $selected_slug ] ?? $options['all'];
 
 	return array(
-		'label' => $label,
+		'label' => sprintf( __( 'Status: %s', 'wporg-learn' ), $label ),
 		'title' => __( 'Completion status', 'wporg-learn' ),
```
outdoor2kode, line 532: Can you change the first letter of "status" to uppercase to make it consistent?

ryelle: We use sentence case across the site, so leaving "status" lowercase in "Completion status" is correct. "Content Type" and "Skill Level" should probably be updated to "Content type" and "Skill level" instead. Unless I'm using that rule too broadly — maybe a check from @thetinyl would help.

thetinyl: Yes, let's use sentence case. Thanks for checking!

ryelle:
Ah sorry @renintw, I wasn't sure what that ticket was referencing, now I understand. If we went with Option 1, there is no feedback about which item is currently selected, and that info is not exposed anywhere else. So I also think Option 2 is better.

> Q: Do we consider removing the label on the dropdown? Maybe it's fine for now and can leave it to post-launch.

That's a wider question, as the change would happen in the query-filter block itself. I would prefer to decouple that change from Learn. I'm making a companion issue in wporg-mu-plugins right now to track updating this on other sites, I'll add a note about this there.

outdoor2kode:
> Ah sorry @renintw, I wasn't sure what that ticket was referencing, now I understand. If we went with Option 1, there is no feedback about which item is currently selected, and that info is not exposed anywhere else. So I also think Option 2 is better.

Not at all 🙂 Yeah, that's also the reason I prefer Option 2.

> > Q: Do we consider removing the label on the dropdown? Maybe it's fine for now and can leave it to post-launch.
>
> That's a wider question, as the change would happen in the query-filter block itself. I would prefer to decouple that change > from Learn. I'm making a companion issue in wporg-mu-plugins right now to track updating this on other sites, I'll add a note about this there.

I just looked around a bit at how other sites handle this and found an approach that might be good- adding an icon as a prefix to the button label, similar to how we switch GH branches here. The title inside the dropdown doesn't need to be removed in this way.

ryelle:
> I just looked around a bit at how other sites handle this and found an approach that might be good- adding an icon as a prefix to the button label, similar to how we switch GH branches here. The title inside the dropdown doesn't need to be removed in this way.

Hm, I don't know what icons that would understandably communicate "level" and "type", and icons instead of text can also be an accessibility issue.

I've added a note here to address whether the dropdown title should be hidden: https://github.com/WordPress/wporg-mu-plugins/issues/645

kathrynwp:
> Do those prefixes make sense? Using the full title for the last two would be too long, IMO, so I went with a short version.

@ryelle Yes these work for me!

outdoor2kode:
> Hm, I don't know what icons that would understandably communicate "level" and "type", and icons instead of text can also be an accessibility issue.

This might need some feedback from the design team. I was thinking that using an appropriate icon along with text inside the dropdown is also likely to help users get familiar with the filter quickly. And I think the icon can have an aria-label or aria-labelledby to handle the accessibility issue?

This isn't a blocker or something that must be done, just thought it might be worth having a preliminary discussion of the possibilities.

> I've added a note here to address whether the dropdown title should be hidden: https://github.com/WordPress/wporg-mu-plugins/issues/645

Thanks!

ryelle:
> This might need some feedback from the design team. I was thinking that using an appropriate icon along with text inside the dropdown is also likely to help users get familiar with the filter quickly. And I think the icon can have an aria-label or aria-labelledby to handle the accessibility issue?

The accessibility issue is more about understanding what the button does, not for screen reader users. For example, a sighted user might see an icon and without knowing what it means, might not click on it. This article has a good overview of [accessible icon design & use](https://www.a11y-collective.com/blog/icon-usability-and-accessibility/) - "An icon’s design should communicate its function clearly and intuitively, reducing the cognitive load on users." I don't know that there is something universally recognizable for these filters.

The github button works a) because we're technical users, and we've seen that icon many times, and b) it's labelled immediately next to the button:

Screenshot 2024-08-01 at 1 14 59 PM

From https://github.com/WordPress/wporg-mu-plugins/issues/645, it sounds like this approach + removing the title from the dropdown works, so I'm going to merge this. We can always iterate on it more.

outdoor2kode (approved):
LGTM 👍 And thanks for sharing that a11y article!

Have we decided to go with the short version? Maybe we should tag Destiny to let her know about the change to her proposal.

ryelle:
> Have we decided to go with the short version? Maybe we should tag Destiny to let her know about the change to her proposal.

I think so, in the interest of space. I'll comment on the other issue so she can reopen it if that exact text was required.

wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
```diff
@@ -100,24 +100,18 @@ function ( $a, $b ) {
...
 	$selected_slug = $wp_query->get( 'wporg_lesson_level' );
 	if ( $selected_slug ) {
 		// Find the selected level from $levels by slug and then get the name.
-		$selected_level = array_filter(
-			$levels,
-			function ( $level ) use ( $selected_slug ) {
-				return $level->slug === $selected_slug;
-			}
-		);
+		$selected_level = wp_list_filter( $levels, array( 'slug' => $selected_slug ) );
```
outdoor2kode, line 108: 💯

PR #2806, opened 2024-07-31
Fix duplicate navigation for lesson nav

### LINKED ISSUE #2782: A11y / Sensei / Learn - Duplicate navigation for lesson nav
## Type of feedback
//dev

## Description

There is a screen-reader-hidden region immediately above the visible lesson navigation buttons that are mostly duplicates of the visible buttons. They are inside a `nav` region, so they are available as a navigation landmark, but the navigation landmark has no accessible name. Additionally, the screen reader text hidden area contains a duplicate and *disabled* 'Completed' button.

Both the link and the button are still available to screen reader users, so screen reader users have up to two additional links and a disabled button in addition to the visible equivalents. 

It'll be even more confusing, since the first 'Completed' button will also be marked as disabled, but the second one is not. 

Furthermore, the second button is an interactive control that is not disabled, but does not do anything. This is confusing for all users. It should just be text, and not be an interactive element. Most likely, this information should be in the H1 heading on the page, so that a user gets that information immediately, rather than needing to find it at the end of the lesson.

Screenshot shows the container as it would appear with the `screen-reader-text` class removed. 

image

outdoor2kode:
Fixes #2782

This PR removes the entire screen-reader-text div - ie. the nav & the first "completed" button are removed.
Also added a visible "Previous Lesson" button to sensei lesson actions.
And since the style of the Sensei next lesson button cannot align well with the previous lesson button on smaller screens (either it doesn't automatically become two rows or it becomes unclickable), a new "Next Lesson" button is also added.

I haven't figured out how to indicate that the current lesson is already completed. Initially it was a button-like div, but it caused confusion, so I removed it. Any thoughts on this? @WordPress/meta-design 
image

## Screencasts

**Before**

https://github.com/user-attachments/assets/ef3570f2-9acf-4509-a5ee-b2c90ef9cc27

**After**

https://github.com/user-attachments/assets/62706564-2568-4667-84c6-a584ce9efd3a

**Smaller screen**
different screen size


jasmussen:
The progressbar is the main piece.

Changing the text of the button, and making it disabled, is a good idea. Could look like this:
light gray

Noting here that the button is disabled, and disabled elements do not need to meet contrast requirements, and therefore it is intentionally very light. We can also darken that text to charcoal 5:


charcoal 5

outdoor2kode:
Thanks @jasmussen. Does this match your expectations of the result?

https://github.com/user-attachments/assets/dfb4abeb-130f-43de-8900-0829e857e3fc

jasmussen:
Yes it does. There may still be optimizations to do with regards to the next/previous lesson buttons and where they sit, perhaps they are reduced to just "Next" and "Previous", and are arrows instead of the bigger buttons. But we're getting into a detail territory that isn't necessarily a blocker for this particular PR.

ryelle:
> Changing the text of the button, and making it disabled, is a good idea.
> Noting here that the button is disabled, and disabled elements do not need to meet contrast requirements, and therefore it is intentionally very light.

I disagree with this. Part of the original feedback in the a11y issue was that "the second button is an interactive control that is not disabled, but does not do anything. This is confusing for all users. It should just be text, and not be an interactive element."

It is not _really_ a disabled button, it's meant to convey information to the user. There's a suggestion that we could add the "completed" label to the main lesson title, and I think that would be better - maybe something like this

Screenshot 2024-08-01 at 11 41 10 AM

Then we could remove the "Completed" faux-button from the end of the lesson, so it only has the next/previous buttons.

wp-content/themes/pub/wporg-learn-2024/src/sensei-lesson-actions-a11y/view.js
```diff
@@ -0,0 +1,14 @@
+function init() {
```
ryelle: This could be done with CSS — `display:none` hides content from everyone, including screen readers.

```css
.sensei-lesson-footer .screen-reader-text {
    display: none;
}
```

outdoor2kode:
> It is not really a disabled button, it's meant to convey information to the user. There's a suggestion that we could add the "completed" label to the main lesson title, and I think that would be better - maybe something like this

Yeah, actually this is what I had in mind originally, but I also feel that using a disabled button could be acceptable, maybe just treating it as a button-like div, as with that commit, the screen reader wouldn't read it as a button, it would just read the text "lesson completed." But it might be a bit confusing to sighted users and adding a label is definitely better.

wp-content/themes/pub/wporg-learn-2024/src/sensei-lesson-actions-a11y/view.js
outdoor2kode: Ah, yeah, I was initially thinking about completely removing it, so I went ahead and did it that way. Thanks!
2899198

outdoor2kode:
> maybe something like this
> image

I'll try this first. However, I was originally thinking about adding a green label somewhere to make it more obvious, for example like adding an icon like you did and making the entire title green. Also, the "completed" status icon in the sidebar should probably be green as well, to make it more consistent with other places. This can be done post-launch, though.

wp-content/themes/pub/wporg-learn-2024/src/style/style.scss
```diff
@@ -51,3 +51,9 @@ body {
 		padding-top: unset !important;
 	}
 }
+
+.disabled {
+	color: var(--wp--preset--color--charcoal-5) !important;
+	border-color: var(--wp--preset--color--charcoal-5) !important;
+	pointer-events: none;
+}
```
ryelle: I think you can remove this style, since the disabled button has been removed.

ryelle (approved):
Looks good! I tested with VoiceOver and no extra buttons/links exist. I think the title update can be handled in a future PR. Good to merge pending the unused CSS removal.

ryelle:
> However, I was originally thinking about adding a green label somewhere to make it more obvious, for example like adding an icon like you did and making the entire title green. Also, the "completed" status icon in the sidebar should probably be green as well, to make it more consistent with other places. This can be done post-launch, though.

This was definitely thrown together in dev tools for a suggestion. I think this can wait until post-launch so we can get some proper design input for it.

wp-content/themes/pub/wporg-learn-2024/src/style/style.scss
outdoor2kode: 21f09df

outdoor2kode:
> This was definitely thrown together in dev tools for a suggestion. I think this can wait until post-launch so we can get some proper design input for it.

Opened [an issue](https://github.com/WordPress/Learn/issues/2809) for this.

PR #2808, opened 2024-08-01
Index lesson CPT

### LINKED ISSUE #2446: SEO
### Requirements

- Single Lessons are indexed 
- Single Lesson social card assets (see [comment](https://github.com/WordPress/Learn/issues/2446#issuecomment-2116942010))
- Try to enable Lesson preview by default (see [comment](https://github.com/WordPress/Learn/issues/2446#issuecomment-2122057499))

outdoor2kode:
Resolves #2446

I didn't index the lesson [here](https://github.com/WordPress/Learn/blob/trunk/wp-content/plugins/wporg-learn/inc/post-type.php#L252-L257) because I think this is the feature belonging to the theme, ie, requiring the lesson to be indexed.

I tried to test it, but it seems that Jetpack cannot be used locally. And on [the test site](https://learn.wordpress.org/test), the sitemap also doesn't seem to work correctly.  So at most, through the error log, I can see that the final returned post_types here include the lesson. And I think the hooks here should be executed later than the plugin's hooks.

Is there a better way to verify it? Or it can only be more conveniently verified after the new theme is activated.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/functions.php
+++ b/wp-content/themes/pub/wporg-learn-2024/functions.php
@@ -43,8 +43,8 @@ add_filter( 'search_template_hierarchy', __NAMESPACE__ . '\modify_search_templat
 add_filter( 'wporg_block_navigation_menus', __NAMESPACE__ . '\add_site_navigation_menus' );
 add_filter( 'wporg_block_site_breadcrumbs', __NAMESPACE__ . '\set_site_breadcrumbs' );
 add_filter( 'taxonomy_template_hierarchy', __NAMESPACE__ . '\modify_taxonomy_template_hierarchy' );
-
 add_filter( 'sensei_learning_mode_lesson_status_icon', __NAMESPACE__ . '\modify_lesson_status_icon_add_aria', 10, 2 );
+add_filter( 'jetpack_sitemap_post_types', __NAMESPACE__ . '\jetpack_sitemap_post_types' );
 
 remove_filter( 'template_include', array( 'Sensei_Templates', 'template_loader' ), 10, 1 );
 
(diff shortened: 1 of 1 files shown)


ryelle:
> Is there a better way to verify it? Or it can only be more conveniently verified after the new theme is activated.

[I did this recently for Documentation](https://github.com/WordPress/wporg-documentation-2022/commit/370522000a07070548392b496a5ebac8b0d0b464), and the only way I could tell to verify was to run it on my sandbox (which did affect production) — I added the code on my sandbox, and ran `wp --url=wordpress.org/documentation cron event run jp_sitemap_cron_hook` to generate the sitemap (my thinking was I could run it again without the change to reset the sitemap). Once I verified the sitemap had the correct info, I committed the change so it would be there for the future cron runs. I think you could probably just deploy this and skip that step, or follow my steps.

> I didn't index the lesson [here](https://github.com/WordPress/Learn/blob/trunk/wp-content/plugins/wporg-learn/inc/post-type.php#L252-L257) because I think this is the feature belonging to the theme, ie, requiring the lesson to be indexed.

I don't know much about the CPT architecture here, but I think it should live in the plugin, not the theme, especially if there is already a filter used there. Now the sitemap is managed in two places, it might be unexpected.

outdoor2kode:
I originally added it there, but it seems it mainly deals with the logic related to tutorial and lesson-plan post types there, and those two post types are planned to be deprecated in the future. Also since the lesson post type is specifically excluded by Sensei LMS, I thought this was a specific need for the 2024 theme and therefore added it to the theme. But you are right; having it in two places might be unexpected.

I will move it to the plugin and merge it first.

outdoor2kode:
Closes in favor of #2813

PR #2816, opened 2024-08-01
Filter searches by allowed post types if post_type is not set

### LINKED ISSUE #2815: Feedback - Tutorials, quizzes, and lesson plans should be excluded from global search results.
## Type of feedback

//dev

Go to learn.wordpress.org
When you do a global search, you will notice tutorials, quizzes and lesson plans.

adamwoodnz:
Fixes https://github.com/WordPress/Learn/issues/2815 by allowing only lessons and courses in global searches where `post_type` is not set. On archives this is set, so they should be unaffected.

### Testing

1. Using the global search from the home page search for an existing tutorial or lesson plan. It should not show in the results.
2. On the tutorial or lesson plan archive perform a search. The results should be unaffected.
3. On the lesson or course archive perform a search. The results should be unaffected.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/block-config.php
@@ -54,7 +54,8 @@ function get_content_type_options( $options ) {
 		'lesson' => __( 'Lesson', 'wporg-learn' ),
 	);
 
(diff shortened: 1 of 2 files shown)


outdoor2kode (approved):
Tested on sandbox, looks good to me 👍

PR #2817, opened 2024-08-02
Add script to fix the issue

outdoor2kode:
Sensei sets the `--sensei-wpadminbar-offset` for regular lessons, and it dynamically adjusts based on screen size. However, it doesn’t set it for freestanding lessons. Previously, we set a fixed value for `--sensei-wpadminbar-offset`, but it’s removed as it causes the header position to be off on smaller screens. 

image

[This](https://github.com/Automattic/sensei/blob/trunk/assets/course-theme/adminbar-layout.ts) is where sensei dynamically calculates the value, I didn't find out how that script can also apply to standalone lesson there, so it's migrated.


adamwoodnz (approved):
Works for me, let's ship this as a hotfix 👍 

Maybe as a follow up we could try either enqueueing the Sensei script, or reference the Sensei typescript file in our block build? Not a major anyway.

outdoor2kode:
Yeah, plan to look into that, but will merge it as a hotfix.

Issue #2577, opened 2024-06-27
Filtering search results by language is broken


adamwoodnz:
Steps to reproduce:

1. Navigate to an archive, eg. https://learn.wordpress.org/courses/?new-theme=1
2. Enter a search term that you can see in the title or excerpt of one of the cards, eg. 'friendly' exists in the excerpt of the Beginner WordPress User course
3. Submit the results and you should land at https://learn.wordpress.org/?s=friendly&post_type=course with results.
4. Filter the results by language. At present Italian is the only option.

Expected: No results as all the courses from the search are in english

Actual: The same result set as without the filter

| With search, no language | With search, and language |
|--------|--------|
| ![Image](https://github.com/WordPress/Learn/assets/1017872/2bb02f7e-6608-4a04-b4ba-bc4420c9d9b8) | ![Image](https://github.com/WordPress/Learn/assets/1017872/e7192e88-b872-46d2-8148-7b0d0c502aba) | 

Note that filtering all courses by language without a search does work: https://learn.wordpress.org/?language%5B%5D=it_IT&post_type=course

![Image](https://github.com/WordPress/Learn/assets/1017872/2cd92476-db74-41ec-93b6-8a3478937eee)

pkevan:
This is slightly complicated for a couple of reasons:

- the indexes for search are all classified as `en` language, I believe you can create a language specific index, but unsure how that is done
- language is stored in the site DB as a post_meta (language), but so you could potentially use this, but only if the meta is whitelisted in JP search so the request can be adjusted to include this

adamwoodnz:
I added `language` to the JP Search Post Meta whitelist, and triggered a partial sync. The sync validation now says that Post Meta is 100%.

This has still not resolved the issue however.

@kangzj do you have any insight on what more we need to do please?

| Post meta | Sync validation |
|--------|--------|
| ![Image](https://github.com/WordPress/Learn/assets/1017872/e0d7677e-1caa-4137-8842-5dd0c178d366) | ![Image](https://github.com/WordPress/Learn/assets/1017872/665115eb-9ddb-4401-8a3d-88ef79d023d1) |

kangzj:
We probably want to trigger a re-index - @trakos might know more.

trakos:
Hi, this is using the older Inline Search (sometimes also known as Jetpack Search Classic) instead of the Instant Search, correct?

[...]

PR #2830, opened 2024-08-07
Add a taxonomy for hiding lessons

### LINKED ISSUE #2735: Hidden Lessons appear in search results
This production Lesson is set to be hidden:

Image

We added a new post meta `_lesson_archive_excluded` to the Lesson post type, to enable to hiding specific Lessons (eg. quizzes) from archive and search views.

This works in local development and for the lessons archive in production, **but not with Jetpack search running in production**. I have tested this on the [staging site](https://learn.wordpress.org/test/) without Jetpack Search running and it works.

### Steps to reproduce

1. Log in to production
2. Create a Lesson and privately publish it. This should mean admins can see it but not regular users. Use a name you can easily search for later.
4. On the frontend navigate to the [Lessons archive](https://learn.wordpress.org/lessons/?new-theme=1) and locate your Lesson in the grid.
5. In the admin check the post meta box in the sidebar for 'Hidden Lesson -> Exclude this lesson from the archive'
Image
6. Refresh the frontend archive and the Lesson should be hidden
7. Search for your Lesson using the search box

#### Expected
The lesson remains hidden

#### Actual
The lesson is included in the search results

adamwoodnz:
Attempt to fix #2735 

Using post meta to hide lessons in search results doesn't work with Jetpack classic search, so this PR attempts to fix this by using a taxonomy, which are indexed. As suggested by @gibrown [here](https://github.com/WordPress/Learn/issues/2577#issuecomment-2248905487) I've used an [existing Jetpack taxonomy 'show'](https://github.com/Automattic/jetpack/blob/36b2d5232e2ec3a1ad14034ab673fddce3acab97/projects/packages/sync/src/modules/class-search.php#L1479) so that we don't need to add a new taxonomy.

I've left the existing post meta field for now while we migrate data to the new taxonomy, but a deprecation label has been added.

Screenshot 2024-08-08 at 2 47 43 PM

### Testing

This can only be tested in sandbox, and I think it will need to be shipped to be fully tested, due to the [Jetpack elastic search sync](https://jptools.wordpress.com/debug/?url=https://learn.wordpress.org/test/#async_get_sync_validator) needing to be run. I've tried syncing before shipping and it doesn't seem to work.


wp-content/plugins/wporg-learn/js/lesson-archive-excluded-meta/index.js
```diff
@@ -16,7 +16,7 @@ const LessonArchiveExcludedMeta = () => {
 	const [ lessonExcluded, setLessonFeatured ] = useState( postMetaData?._lesson_archive_excluded === EXCLUDED );
 
 	return (
-		<PluginDocumentSettingPanel title={ __( 'Hidden Lesson', 'wporg-learn' ) }>
+		<PluginDocumentSettingPanel title={ __( 'Hidden Lesson (deprecated)', 'wporg-learn' ) }>
```
adamwoodnz, line 19: Delete this setting panel block once data migration is complete

wp-content/themes/pub/wporg-learn-2024/inc/query.php
```diff
@@ -52,8 +52,9 @@ function add_excluded_to_lesson_archive_query( $query ) {
 	// Ensure this code runs only for the main query on lesson archive pages and search results.
 	if ( ! is_admin() && $query->is_main_query() && ( $query->is_archive( 'lesson' ) || $query->is_search() ) ) {
 		$meta_query = $query->get( 'meta_query', array() );
+		$tax_query = $query->get( 'tax_query', array() );
 
-		$exclude_lesson_query = array(
+		$exclude_lessons_by_post_meta = array(
```
adamwoodnz, line 57: Meta query can be removed once data migration is complete

adamwoodnz:
@gibrown @kangzj this has shipped now, but doesn't seem to have fixed this issue with hidden lessons appearing in search results. This production lesson 'Quiz time' has the 'hidden' term from the existing Jetpack 'show' taxonomy applied. It is hidden when viewing the archive, but [visible in search](https://learn.wordpress.org/?s=quiz&post_type=lesson):



I've used the [jptools](https://jptools.wordpress.com/debug/?url=https://learn.wordpress.org/#async_section_search) to check the post (id 257231) and I can see that it has the taxonomy applied.

I have tried to do an [incremental sync](https://jptools.wordpress.com/debug/?url=https://learn.wordpress.org/#async_get_incremental_sync_status), but discrepancies are still reported:

```
WP.com Cache has discrepancies as of last checksum on 2024-08-09 00:38:37


WP.com Cache Table Accuracy
posts: 100.00%
postmeta: 99.91%
comments: 100.00%
commentmeta: 100.00%
terms: 100.00%
term_relationships: 98.09%
term_taxonomy: 100.00%
users: 99.41%
usermeta: 99.42%
```

Should I try a full sync?

I have tried this in the test site too with a full sync and [it's not working there either](https://learn.wordpress.org/test/?s=hidden&post_type=lesson), but does work with Jetpack Search disabled:

| JP search off | JP search on |
|--------|--------|
| Screenshot 2024-08-09 at 12 29 32 PM | Screenshot 2024-08-09 at 12 29 52 PM | 

Any idea what's going wrong?

Thanks in advance for your help.

trakos:
I've replied in https://github.com/WordPress/Learn/issues/2735

PR #2833, opened 2024-08-08
Tools: update packages

adamwoodnz:
@wordpress/scripts 27.9.0 includes an RTL build so the custom webpack config can be removed


adamwoodnz:
Build passed, checked sandbox, all looks good.

Issue #2832, opened 2024-08-08
Remove old themes


adamwoodnz:
Now that the `wporg-learn-2024` theme has been activated, we can delete `wporg-learn` and `wporg-learn-2020`, and associated build scripts.

PR #2835, opened 2024-08-08
Remove deprecated 2020 Learn child theme and parent

adamwoodnz:
See #2832 

- Removes the `wporg-learn-2020` theme and the dependency on the `wporg` parent theme.
- Updates build and lint workflow scripts and configs.
- Removes some now unused code from the plugin.
- Updates the readme.

Note that the `wporg-learn` theme in SVN is not in this Git repo and needs to be deleted separately.

### Testing

1. Make a fresh clone of the repo and check out this branch.
2. Follow the readme and ensure the site starts up successfully with the 2024 theme activated.

To regression test the build you can check out the `build-test` branch (which was built off this branch), and run it locally, or sync the individual projects to a sandbox.


wp-content/plugins/wporg-learn/js/block-styles/style.scss
adamwoodnz, line 1: Old button styles

PR #2843, opened 2024-08-14
Support excluding posts based on taxonomy in search queries

### LINKED ISSUE #2735: Hidden Lessons appear in search results

trakos:
Attempt to fix https://github.com/WordPress/Learn/issues/2735

In https://github.com/WordPress/Learn/pull/2830 a taxonomy `show` was added. It also added a `NOT IN` condition to the search query, and posts with value `hidden` should not appear in search results.

It appears that Jetpack Search, which is used as a backend for searching lessons, does not support `NOT IN` queries for taxonomies out of the box. The code for converting tax_query is [here](https://github.com/Automattic/jetpack/blob/b1fe483220715ce11c87c6c0e62e005abd34e661/projects/packages/search/src/classic-search/class-classic-search.php#L661-L692). It returns early because `queried_terms` isn't set. If it would be set, it would actually be worse, as it would only include hidden results instead of excluding them.

In this PR, I propose to add two Jetpack Search filters to manually fix the query. One excludes `show` from `terms` in case `queried_terms` would be present, and the other converts `NOT IN` queries to `must_not` query in Jetpack Search.

I don't have access to .org sandbox, so I tested it by adding a hacky `print_r`:

```
if ( isset($this->query['s']) ) {
	$search = \Automattic\Jetpack\Search\Classic_Search::instance(47261184);
	$search->do_search( $this);
	echo '<pre>';
	print_r($search->get_search_result());
	die();
}
```

in `wp-includes/class.wp-query.php` on line 3161 (before applying filter `posts_pre_query`, which is what normally triggers Jetpack Search) in WP Learn Docker container. I've verified that the problematic post_id disappeared when visiting http://localhost:8888/lessons/?s=quiz&post_type=lesson.

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/inc/query.php
+++ b/wp-content/themes/pub/wporg-learn-2024/inc/query.php
@@ -9,6 +9,8 @@ add_action( 'pre_get_posts', __NAMESPACE__ . '\add_language_to_archive_queries' 
 add_action( 'pre_get_posts', __NAMESPACE__ . '\add_excluded_to_lesson_archive_query' );
 add_action( 'pre_get_posts', __NAMESPACE__ . '\filter_search_queries_by_post_type' );
 add_filter( 'request', __NAMESPACE__ . '\handle_all_level_query' );
+add_filter( 'jetpack_search_es_wp_query_args', __NAMESPACE__ . '\filter_jetpack_wp_search_query', 10, 2 );
+add_filter( 'jetpack_search_es_query_args', __NAMESPACE__ . '\filter_jetpack_es_search_query', 10, 2 );
 
 /**
  * Modify the query by adding meta query for language if set.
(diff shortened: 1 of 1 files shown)


trakos:
cc @adamwoodnz

adamwoodnz:
Thanks @trakos! Tested in sandbox and it works for the `Quiz time` lesson ✅ 

I added the term to another lesson, [Test your knowledge](https://learn.wordpress.org/lesson/test-your-knowledge/), but it is still showing in the [search results](https://learn.wordpress.org/?s=quiz&post_type=lesson).

Screenshot 2024-08-16 at 12 15 03 PM

I ran ['Schedule Checksum' and then 'Schedule Fix'](https://jptools.wordpress.com/debug/?url=https://learn.wordpress.org/#async_get_sync_validator), but that didn't help. 

Could you please help me understand the lifecycle for this data; how soon should we expect lessons to be hidden after applying the term in the admin? It looks the cron runs nightly, but I was hoping my actions would have had the same effect...

adamwoodnz:
Tested this again today and it does seem to be working, even if there is some delay. Merging, but it would still be good to understand the timing of data sync.

trakos:
> Could you please help me understand the lifecycle for this data; how soon should we expect lessons to be hidden after applying the term in the admin?

I think it should normally be updated in minutes. Indexing delay is set to 30 seconds, sync should take a couple minutes, and the search is cached for 5 minutes. That particular post was last re-indexed on `2024-08-18 22:56:36` UTC after it was modified on `2024-08-18 22:54:40`.

Did you update the terms simply by editing the post through wp-admin? Trying to figure out if the right Jetpack Sync actions were called.

adamwoodnz:
> Did you update the terms simply by editing the post through wp-admin? Trying to figure out if the right Jetpack Sync actions were called.

Yeah just wp-admin. That timing seems consistent with what I've observed. Thanks again for your help.

Issue #2496, opened 2024-06-03
Redirects for old content


adamwoodnz:
Setup redirects for old content (eg. tutorials) to more recent content. Follow a similar approach to what we do for the [block editor handbook in Dev Resources](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/import-block-editor.php#L46). Should be probably live at https://github.com/WordPress/Learn/tree/trunk/wp-content/plugins/wporg-learn/inc/redirects.php

kathrynwp:
Thanks for getting the ball rolling with this. Learn content creators are preparing a list of redirect URLs for old tutorials and lessons which are being retired. I'll link up the spreadsheet here when it's ready.

adamwoodnz:
Noting that for posts that we don't migrate, we have the option of adding noindex to exclude the content from external and internal search, but remain discoverable by direct link. We have another issue https://github.com/WordPress/Learn/issues/2585 for this.

peiraisotta:
Hello! Can you please also redirect the last three lessons on [this spreadsheet](https://docs.google.com/spreadsheets/d/1go7tfrhXXr3j3wWoHVJ06ixXwZMfBV35OsTLqBahEGk/edit?usp=sharing)? 
You'll find the new links in column L

(fyi @kathrynwp)

kathrynwp:
@adamwoodnz The Community team (thanks to Isotta) has confirmed that we can remove all their standalone lessons – which I will go ahead and do. Before removing the following three standalone lessons, however they would appreciate it if we could please redirect the following three lessons to corresponding WordPress.tv pages:

1) https://learn.wordpress.org/tutorial/block-editor-02-text-blocks/ 
- redirect to https://wordpress.tv/2021/06/03/shusei-toda-block-editor-02-text-blocks/

2) https://learn.wordpress.org/tutorial/block-editor-01-basics/ 
- redirect to https://wordpress.tv/2021/06/18/shusei-toda-naoko-takano-block-editor-01-basics/

3) https://learn.wordpress.org/tutorial/ja-login-password-reset/ 
- redirect to https://wordpress.tv/2021/02/16/login-password-reset/

Let me know if this is possible, thank you!

adamwoodnz:
✅ @peiraisotta @kathrynwp these 3 have now been saved as draft and redirected

kathrynwp:
@adamwoodnz Could you please redirect:

[...]

PR #2860, opened 2024-08-26
Redirect deprecated tutorials to wordpress.tv

adamwoodnz:
Adds redirects for several deprecated tutorials to external locations on WordPress.tv

Also updates the existing internal redirects to use 301, as these are permanent, and the default for `wp_safe_redirect` is 302.

See https://github.com/WordPress/Learn/issues/2496, specifically https://github.com/WordPress/Learn/issues/2496#issuecomment-2305200045

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/redirects.php
+++ b/wp-content/plugins/wporg-learn/inc/redirects.php
@@ -44,6 +44,10 @@ function wporg_learn_redirect_old_urls() {
 		'/social-learning'                => '/online-workshops',
 		'/workshop-presenter-application' => '/tutorial-presenter-application',
 		'/report-content-errors'          => '/report-content-feedback',
+		// External redirects.
+		'/tutorial/block-editor-01-basics/'      => 'https://wordpress.tv/2021/06/18/shusei-toda-naoko-takano-block-editor-01-basics/',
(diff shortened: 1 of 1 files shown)


wp-content/plugins/wporg-learn/inc/redirects.php
```diff
@@ -52,14 +56,17 @@ function wporg_learn_redirect_old_urls() {
...
 			if ( strlen( $request ) > strlen( $source ) ) {
 				$redirect .= substr( $request, strlen( $source ) );
 			}
 
-			wp_safe_redirect( $redirect );
-			die();
+			str_starts_with( $destination, 'https://' )
+				? wp_redirect( $redirect, $code, 'Learn WordPress' )
+				: wp_safe_redirect( $redirect, $code );
+			exit;
```
dd32: The alternative here is to add `wordpress.tv` to the 'allowed_redirect_hosts' filter, which is probably the route I'd go. That way it passes through `wp_safe_redirect()` as it's then a safe-destination.

adamwoodnz: Ah great, that's exactly the input I was after, cheers

dd32 (approved):

PR #2866, opened 2024-08-30
Make course lessons publicly previewable by default

### LINKED ISSUE #2807: Check whether it's possible to enable previews on all course lessons as default
Originally discussed in https://github.com/WordPress/Learn/issues/2446#issuecomment-2122057499
> During development of the learning pathway content, we discovered that Sensei has a featured to enable Lesson previews, which is what makes the individual lessons in the [Beginner WordPress User](https://learn.wordpress.org/course/beginner-wordpress-user/) course public.
>
> Would it be possible to investigate whether we're able to simply enable this on all course lessons as default, instead of needing to enable it on each course lesson?

- This is saved in [meta field](https://github.com/Automattic/sensei/blob/e3ef1943bc965070fd59419f734af458035bcd81/includes/class-sensei-lesson.php#L630). You should be able to set it to `true` on lesson creation. 
_Originally commented by @StevenDufresne_

>

- It's registered [here](https://github.com/Automattic/sensei/blob/e3ef1943bc965070fd59419f734af458035bcd81/includes/rest-api/class-sensei-rest-api-lessons-controller.php#L100), with no default value, and no way to set the default value to `true`, unless I'm missing something.  
- Maybe we could deregister and reregister with a default value.
_Originally commented by @adamwoodnz_

>

- Not sure if this is still relevant, but you can set the default value of a post meta field via the filter here: https://github.com/WordPress/wordpress-develop/blob/6.6/src/wp-includes/meta.php#L717  
_Originally commented by @pkevan_

adamwoodnz:
Closes #2807

Use the `wp_insert_post` hook to set public preview to be enabled by default on lessons which belong a course.

@jonathanbossenger we could enable public preview by default for all lessons, but your [comment](https://github.com/WordPress/Learn/issues/2446#issuecomment-2122057499) made me think we only want to do it for course lessons. Let me know if this isn't a requirement 🙏 

| Admin | Frontend |
|--------|--------|
| Screenshot 2024-08-30 at 5 42 05 PM | Screenshot 2024-08-30 at 5 43 45 PM | 

#### Course lesson
1. Create a course and add a lesson to the course outline
2. When the draft is created it should have `_lesson_preview` meta set. The 'preview' pill should appear in the editor on the lesson in the outline.
3. Edit the lesson and check that the Preview checkbox is ticked
4. Logout and view the lesson on the frontend. You should be able to view the content without being logged in.

#### Standalone lesson
1. Create a lesson outside of a course
2. Preview post meta should not be set

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -22,6 +22,7 @@ add_action( 'save_post_wporg_workshop', __NAMESPACE__ . '\save_workshop_meta_fie
 add_action( 'save_post_meeting', __NAMESPACE__ . '\save_meeting_metabox_fields' );
 add_action( 'admin_footer', __NAMESPACE__ . '\render_locales_list' );
 add_action( 'enqueue_block_editor_assets', __NAMESPACE__ . '\enqueue_editor_assets' );
+add_action( 'wp_insert_post', __NAMESPACE__ . '\set_default_lesson_preview', 10, 3 );
 
 /**
  * Register all post meta keys.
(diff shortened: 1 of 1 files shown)


jonathanbossenger:
> Let me know if this isn't a requirement 🙏

Hmm, that's a great question. 

I assume it would just be a case of removing these checks?

```
        // Check if the lesson belongs to a course.
	$course_id = get_post_meta( $post_ID, '_lesson_course', true );

	if ( empty( $course_id ) ) {
		return;
	}
```

At the moment, any stand-alone lesson we might create would be a lesson that was migrated from a stand-alone tutorial. So it would be published by default anyway. 

For any lesson that belongs to a course, we want to have publicly "previewable" while the course is in development, and then when the course is published, it will be "contained" in the course UI anyway.

So I think this is fine for now. If we decide to make this available for all lessons, we can always just remove those checks.

jonathanbossenger (approved):

PR #2875, opened 2024-09-05
fix the first query item doesn't display except

### LINKED ISSUE #2682: Bug: Some lessons don't display automatic excerpts
## Type of feedback

//dev

## Description

Some lessons are missing an automatic excerpt.

1. View Lessons archives: https://learn.wordpress.org/lessons/
2. Go through pagination
3. Note that some lessons have a blank excerpt area

## Screenshots or screen recording

Lessons___Learn_WordPress

## Expected

When there's no custom/manual excerpt on a lesson, an automatic excerpt should be displayed, pulling through the first 16 words from the lesson.

outdoor2kode:
Fixes #2682

It appears that the first post result of a query doesn't render an automatically generated excerpt due to an ordering issue or because `get_the_excerpt` is not being called. The automatic excerpt will only be generated after a get_the_excerpt call.

I couldn't find a relevant hook to solve this in the source code for the post-template block and post-excerpt block, so a `get_the_excpert` is added at the start.

**Screenshots**
Tested in the sandboxed environment, and everything looks fine. I also went through 20 pages, and they look good.

Screenshot 2024-09-06 at 05 05 57
Screenshot 2024-09-06 at 05 06 17

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
+++ b/wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
@@ -5,6 +5,8 @@
  * Inserter: no
  */
 
+get_the_excerpt();
+
 ?>
 
 <!-- wp:group {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained","justifyContent":"left","contentSize":"750px"}} -->
--- a/wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
+++ b/wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
@@ -5,6 +5,7 @@
  * Inserter: no
  */
 
+// Workaround for the first lesson in the grid not having an automatic excerpt generated.
 get_the_excerpt();
 
 ?>


wp-content/themes/pub/wporg-learn-2024/patterns/archive-lessons-content.php
```diff
@@ -5,6 +5,8 @@
  * Inserter: no
  */
 
+get_the_excerpt();
```
adamwoodnz, line 9:
```suggestion
// Workaround for the first lesson in the grid not having an automatic excerpt generated
get_the_excerpt();
```

adamwoodnz: Tested and this works for me, so I think it's good to merge as a workaround, with a comment. I think it would still be good to understand where this is failing though, and potentially raise a Gutenberg bug.

outdoor2kode: 2ebf0f6

Yeah, I spent some time looking into it, but couldn't find what part is causing the issue. There isn't any related discussion in Gutenberg either. Have already spent more time on this than expected, so just go with a workaround. I'll look into further the flow for this part in Gutenberg, and still can't figure it out, I'll open a ticket.

PR #2881, opened 2024-09-11
Fix quiz styles

### LINKED ISSUE #2686: Feedback - Quizzes with 'Awaiting Teacher Grade' has a 'Pending teacher grade' button that doesn't open
## Type of feedback
//content

## Description
In quizzes awaiting teacher grade like https://learn.wordpress.org/quiz/mentoring-community-organizers-3/ after taking the quiz, it will display a "Pending teacher grade" button that is clickable but doesn't open anywhere. 

Is  "Pending teacher grade" supposed to be a button or just a label? 

1. Go to https://learn.wordpress.org/quiz/mentoring-community-organizers-3/ or in lessons with quizzes that need to be graded by a teacher. 
2. Complete the quiz.
3. Click the "Pending teacher grade" button that appears. 

Screen Capture on 2024-07-16 at 18-53-08

Chrome 124, macOS 14.5

outdoor2kode:
Resolves #2686

This PR replaces the pending text to align with the design as well as adding required notices. 
Also updates quiz action styles and the success icon image.

## Screencast

https://github.com/user-attachments/assets/5cb567f2-121e-444e-a932-23ed975f135c

@kathrynwp, I implemented @fcoveram designs first as I think they make sense. Above is the screencast of the flow based on the design. The differences compared to the original are in the screenshots below.

| Before | After |
|-|-|
| image | Screenshot 2024-09-12 at 06 14 53 |
| image | Screenshot 2024-09-12 at 06 15 17 |

The flow is generally as shown. And after the teacher finishes grading, It will return to looking the same as the others:

Screenshot 2024-09-12 at 06 11 20

## TODO

1. Update the copy
image
image

2. Does the quiz grade on the lesson page also need to be removed? @fcoveram 
image


wp-content/themes/pub/wporg-learn-2024/patterns/sensei-quiz-notices.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @package wporg-learn-2024
+ */
+
+$lesson_id         = Sensei()->quiz->get_lesson_id();
+$course_id         = Sensei()->lesson->get_course_id( $lesson_id );
+$is_learning_mode  = Sensei_Course_Theme_Option::has_learning_mode_enabled( $course_id );
+$is_awaiting_grade = Sensei_Quiz::is_quiz_awaiting_grade_for_user( $lesson_id, get_current_user_id() );
+$is_pending_grade = $is_learning_mode && $is_awaiting_grade;
+$referrer = esc_url( $_SERVER['HTTP_REFERER'] );
```
adamwoodnz: Did you mean to use this on the next line?

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-columns.php
```diff
@@ -29,7 +29,11 @@
 
 		<?php if ( $is_completed ) : ?>
 			<!-- wp:wporg/notice {"style":{"spacing":{"margin":{"top":"var:preset|spacing|20"}}}} -->
-			<div class="wp-block-wporg-notice is-tip-notice" style="margin-top:var(--wp--preset--spacing--20)">
+			<div class="wp-block-wporg-notice is-tip-notice" 
+				style="
+					margin-top: <?php echo $module ? 'var(--wp--preset--spacing--20)' : 'var(--wp--preset--spacing--0)'; ?>;
```
adamwoodnz: If you need to unset this I'd use:
```suggestion
					margin-top: <?php echo $module ? 'var(--wp--preset--spacing--20)' : 'unset'; ?>;
```
Otherwise can you omit `margin-top` altogether?

wp-content/themes/pub/wporg-learn-2024/src/sensei-quiz/view.js
```diff
@@ -0,0 +1,43 @@
...
+	if ( noticeContent && ! grade ) {
+		const newParagraph = document.createElement( 'p' );
+		const noticeText = noticeContent.querySelector( '.sensei-course-theme-lesson-quiz-notice__text' );
+
+		noticeText.remove();
+		newParagraph.textContent =
+			'[TBD. Sentence conveying that user is waiting for the teacher to assign a grade]';
+		newParagraph.classList.add( 'sensei-course-theme-lesson-quiz-notice__description' );
+		noticeContent.insertAdjacentElement( 'afterend', newParagraph );
+	}
```
adamwoodnz: Have you considered doing this server side with the WP HTML API? It's rendered by the `wp-block-sensei-lms-course-theme-notices` block, and we have an example of filtering and modifying block output [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/local-navigation-bar/index.php#L14).

adamwoodnz: We might be able to modify the pending teacher grade button to become a link button too, rather than a button with click handler.

kathrynwp:
This looks good, thank you! 

The one thing I don't understand why a true/false question would need to be graded manually by a teacher. I thought that was only for open-ended questions? Or is it for any type of question, when the quiz is **not** set to be automatically graded?

fcoveram:
Everything looks great.

> 2. Does the quiz grade on the lesson page also need to be removed? @fcoveram

I will defer this question to @kathrynwp. If the info displayed is relevant to learners, then I will keep it.

kathrynwp:
> Does the quiz grade on the lesson page also need to be removed? 

> I will defer this question to @kathrynwp. If the info displayed is relevant to learners, then I will keep it.

Is the reason you're suggesting removing it because it will always be **0%** until the quiz is graded by the instructor? **If** that's the case, then I think it should be removed as it's very confusing to see 0% for a quiz that hasn't been graded yet.

If this is a "real" grade then I think we can keep it.

cc @westnz for input as well as I will be AFK over the next week and don't want to hold this up.

westnz:
At this stage, we must remind ourselves to select 'Auto-grade' as we don't have any assessments that require an instructor to grade them.  

> Is the reason you're suggesting removing it because it will always be 0% until the quiz is graded by the instructor? If that's the case, then I think it should be removed as it's very confusing to see 0% for a quiz that hasn't been graded yet.
> 
> If this is a "real" grade then I think we can keep it.

Agreed, if the real grade was 'auto-graded'.

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-quiz-notices.php
outdoor2kode: That's for testing, thanks 👍 
0437c22

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-columns.php
outdoor2kode: `margin-top` omitted. 
e49df94

wp-content/themes/pub/wporg-learn-2024/src/sensei-quiz/view.js
renintw: > Have you considered doing this server side with the WP HTML API? It's rendered by the wp-block-sensei-lms-course-theme-notices block, and we have an example of filtering and modifying block output [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/local-navigation-bar/index.php#L14).

Thanks, forget there's this function! 43705a7, 9e139b1

outdoor2kode: > Have you considered doing this server side with the WP HTML API? It's rendered by the wp-block-sensei-lms-course-theme-notices block, and we have an example of filtering and modifying block output [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/local-navigation-bar/index.php#L14).

Thanks, forgot that there's this function! 43705a7, 9e139b1

outdoor2kode: > We might be able to modify the pending teacher grade button to become a link button too, rather than a button with click handler.

It's trickier to replace the button here because there's no supported method to change the text or add a link. As a result a new button block was added directly. d79350d

outdoor2kode: [Add comments and clean up conditions being used](https://github.com/WordPress/Learn/pull/2881/commits/84f8cee1287a1310272f59355b7404f4301a0a8e)
[Remove sensei-quiz block and move the logic to block-hooks](https://github.com/WordPress/Learn/pull/2881/commits/f5062c851a62e30513802720bf75fbcae0aa4c1a)

outdoor2kode:
> The one thing I don't understand why a true/false question would need to be graded manually by a teacher. I thought that was only for open-ended questions? Or is it for any type of question, when the quiz is not set to be automatically graded?

That's right, as long as the switch here for a quiz isn't turned on, all questions will require the teacher to grade them. This part needs to be modified on the Sensei plugin side.
image

> Is the reason you're suggesting removing it because it will always be 0% until the quiz is graded by the instructor? If that's the case, then I think it should be removed as it's very confusing to see 0% for a quiz that hasn't been graded yet.
>
> If this is a "real" grade then I think we can keep it.

It's a real grade that will show up only after a quiz is completed and graded. The reason I asked was to align with the pending-graded notice. The 'Awaiting grade' text has been removed there. I'm currently following the feedback provided and keeping 'Your Grade' for now. If it needs to be removed, please let me know. Thanks. @westnz 

| Before | After |
|-|-|
| image | Screenshot 2024-09-12 at 06 15 17 |
|| image |

outdoor2kode:
@thetinyl, could you suggest a copy for the two messages below? Thanks!

image
image

**Here's the flow:**

https://github.com/user-attachments/assets/2d092b06-b539-4c85-a9bb-6fb27fc46749

westnz:
Thanks, Ren. Sounds good.

thetinyl:
I had a chat with @kathrynwp about this and my understanding for this flow is:

- The quizzes with open-ended questions will never be graded. 
- They're going to be phased out relatively soon. 

Given that there is no grading process (and there won't be one), I think it makes sense to frame the language here around these being "ungraded quizzes". I'll follow up later/tomorrow with an actual copy suggestion, but just wanted to note this alternate direction in case others have other thoughts.

thetinyl:
> @thetinyl, could you suggest a copy for the two messages below?

@renintw I think we're going to need to update the copy on the first screen as well (see screenshot below), since it still makes a reference to grading. 

image

Here's some suggested copy for each:

**Quiz submission message**

> Well done! You completed the quiz.

**Quiz completed message**
> **QUIZ COMPLETED**
> This quiz won’t be graded. If you found the questions easy, you likely have a good understanding of the lesson.

**View quiz message**
> This is an ungraded quiz. Use it to check your comfort level with what you’ve learned.

outdoor2kode:
Thanks @thetinyl, there are the updated screenshots.

Screenshot 2024-09-19 at 04 24 45
Screenshot 2024-09-19 at 04 26 25
Screenshot 2024-09-19 at 04 25 41

thetinyl:
Thanks @renintw. I just have one minor copy edit for the following message:

image

Let's revise to read:
> This quiz won’t be graded. If you found the questions easy, you likely understand the lesson well.

fcoveram:
Thanks @thetinyl, I updated the design file to reflect these copy changes

wp-content/themes/pub/wporg-learn-2024/inc/block-hooks.php
```diff
@@ -106,3 +107,92 @@ function update_course_outline_block_add_aria( $block_content, $block ) {
...
+function update_lesson_quiz_notice_text( $block_content ) {
+	if ( is_singular( 'lesson' ) && is_quiz_ungraded() ) {
+		$tag_processor = new \WP_HTML_Tag_Processor( $block_content );
+
+		// Hide the text "Awaiting grade" in the quiz notice.
+		if ( $tag_processor->next_tag( array(
+			'tag_name' => 'div',
+			'class_name' => 'sensei-course-theme-lesson-quiz-notice__text',
+		) ) ) {
+			$tag_processor->set_attribute( 'style', 'display:none;' );
```
adamwoodnz: I notice you've used `str_replace` below to add an element. Could you use it here to replace this element with an empty string and remove it from the DOM instead of hiding?

adamwoodnz (approved):
LGTM, one question but not blocking

wp-content/themes/pub/wporg-learn-2024/inc/block-hooks.php
outdoor2kode: Good call. 7831a3f

PR #2884, opened 2024-09-12
Adding deprecated course redirects

### LINKED ISSUE #2868: Course deprecation
Some courses have become redundant or outdated with the initial learning pathways now published. To streamline our offerings and ensure we provide the most current content to our learners, @westnz and @jonathanbossenger want to deprecate the following courses and direct users to the relevant learning pathway.

- https://learn.wordpress.org/course/simple-site-design-with-full-site-editing/
- https://learn.wordpress.org/course/part-2-personalized-site-design-with-full-site-editing-and-theme-blocks/
- https://learn.wordpress.org/course/part-3-advanced-site-design-with-full-site-editing-site-editor-templates-and-template-parts/
- https://learn.wordpress.org/course/developing-with-the-wordpress-rest-api/
- https://learn.wordpress.org/course/converting-a-shortcode-to-a-block/
- https://learn.wordpress.org/course/a-developers-guide-to-block-themes-part-1/
- https://learn.wordpress.org/course/a-developers-guide-to-block-themes-part-2/

jonathanbossenger:
Fixes #2868

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/redirects.php
+++ b/wp-content/plugins/wporg-learn/inc/redirects.php
@@ -49,7 +49,7 @@ function wporg_learn_redirect_old_urls() {
 		return;
 	}
 
-	$redirects = array(
+	$pages = array(
 		// Source => Destination, any characters after the source will be appended to the destination.
 		'/handbook'                              => 'https://make.wordpress.org/training/handbook/',
 		'/report-content-errors'                 => '/report-content-feedback',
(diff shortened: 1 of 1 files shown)


jonathanbossenger:
@adamwoodnz I'm not sure what you think of what I did here, but I thought it would make sense to group redirects by type.

wp-content/plugins/wporg-learn/inc/redirects.php
```diff
@@ -49,7 +49,7 @@ function wporg_learn_redirect_old_urls() {
 		return;
 	}
 
-	$redirects = array(
+	$pages = array(
 		// Source => Destination, any characters after the source will be appended to the destination.
```
adamwoodnz: Should probably move this comment out of this block as it applies to all arrays

wp-content/plugins/wporg-learn/inc/redirects.php
```diff
@@ -49,7 +49,7 @@ function wporg_learn_redirect_old_urls() {
 		return;
 	}
 
-	$redirects = array(
+	$pages = array(
```
adamwoodnz, line 55: Not all of the redirects in this group are pages, some are post types. If we're going to split out courses we should do the same for tutorials.

adamwoodnz:
> @adamwoodnz I'm not sure what you think of what I did here, but I thought it would make sense to group redirects by type.

I'm not opposed to doing this if you think there's value in organising them. Left some comments about going further with the approach if so.

wp-content/plugins/wporg-learn/inc/redirects.php
jonathanbossenger: I agree; that's why I wanted to check with you first. I'll update the PR today to separate these better.

jonathanbossenger:
Thanks @adamwoodnz PR updated based on your feedback.

adamwoodnz (approved):
LGTM :shipit:

Issue #2392, opened 2024-04-23
Lesson Plan data migration


adamwoodnz:
Migrate Lesson Plans to Lessons, with teacher's notes.

Add redirects from old URLs to new URLs (see https://github.com/WordPress/Learn/issues/2496)

kathrynwp:
Breaking this down a little more, here are some things I'm thinking we'll need to do:

### Admin/content work

- Map each Lesson Plan to the most appropriate lesson. For example, a Lesson Plan for teaching how to create a child theme could be mapped to a Lesson on how to create a child theme.
- Decide how to handle Lesson Plans that correspond best to a course, instead of a lesson.
- For Lesson Plans that _don't_ have a corresponding Lesson, decide how to proceed. Possible avenues include:
-- deprecating that Lesson Plan – whether that means unpublishing or redirecting it somehow
-- letting that Lesson Plan remain available somehow
- Decide what to do with Lesson Plans that are out of date
- Decide what we content want to be on the new Teacher Notes page, i.e. is it a simple porting of the Lesson Plan or do we want to make any changes? Can Teacher Notes include more than a lesson plan?

### Design work

- Design Teacher Notes button/CTA
- Design Teacher Notes page

### Dev work

- Add button/CTA on Lessons to a new Teacher Notes page
- Migrate Lesson Plans to Teacher Notes

adamwoodnz:
> * Migrate Lesson Plans to Teacher Notes

Feels like we could rename Lesson Plans to Teacher's Notes in the UI rather than migrating content to another post type. We have already done this for a couple of other post types on this site, for example Tutorials used to be called Workshops (that's why the slug for tutorials is `wporg_workshop`).

kathrynwp:
> Feels like we could rename Lesson Plans to Teacher's Notes in the UI rather than migrating content to another post type. We have already done this for a couple of other post types on this site, for example Tutorials used to be called Workshops (that's why the slug for tutorials is wporg_workshop).

This makes a lot of sense to me, and I even had the same thought while writing up the notes here!

Piyopiyo-Kitsune:
Just jumping in here for clarity--

The desired end state for “Teacher’s Notes” is a tab or popup sidebar section within a lesson that can host the details of a lesson plan for a facilitator. The information of the lesson plan would be stored in a custom field / block within the lesson editor.

This is in order to reduce the need for us to have to edit duplicative information across different post types, and instead manage all information in one lesson. This also will help us with streamlining translation work.

[...]

PR #2927, opened 2024-10-07
Add Facilitator Notes

### LINKED ISSUE #2869: A revamped design in Lesson flow
I revisited the Lesson flows related to the learning journey and [designed a new version](https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp/Learn?node-id=5492-15422&node-type=SECTION&m=dev) that simplifies the UI and addresses the requirements pointed out in the following tickets:

- #2809
- #2638
- #2864
- #2855

Below you'll see a description of the main changes applied. The call for feedback is open so we can assess whether the proposal meets the expectations and how doable is the outlined.

## General layout

To create a more immersive experience, I revisited the spacings and reduced the values in the header, the left sidebar, and the content area. Given that the WordPress global header and the Learn local nav are not present, we can use the areas more freely and not be constrained by the WordPress site context.

That being said, the header has less padding on the sides, and the exit course action was placed separately to create a lesson area where the breadcrumb and the new “Facilitator notes” action exist.

If you explore the mockups in dev mode, you will see that lesson content and quiz content are wrapped in a container with a max-width of 960px. I think this value keeps texts readable, but I would like your thoughts as this relates to other layouts implemented across the site.

## Main action

On every lesson page, there is a main action that allows users to complete the lesson or take the quiz to complete it. Since users can browse lessons regardless of their order, it seemed logical to place the previous and next actions at the end of the reading direction.

Regarding completing the lesson, I reused green to accentuate the lesson state and match the progress bar on the Course details page.

Mockups of lesson with two main actions

## Login and Take course

When users land on a lesson in preview, they must log in and then take the course to start the learning journey. In both cases, the message was placed below the content, replacing the browse and main actions to ease exploring the lesson.

The eye icon in the left sidebar was replaced by a locker icon, and the “Preview” label on the right of the lesson name was removed to convey the lesson state more directly. In addition, a “Preview mode” pill was added in the content area next to the module name.

Mockups of lesson with login and take lesson messages

## Lesson progress

Completing a lesson has two paths: taking the quiz or clicking on the “Complete lesson” button. For the former, the main blueberry was applied to convey the primary action and jump to the quiz page, and for the latter, the green color keeps the progress visual language.

The progress bar in the header is placed above the bottom border in gray instead of adding an extra component that pushes down the areas below.

Once a lesson is completed, the status icon on the left sidebar changes to a check icon and a completed message appears at the beginning of the page.

Mockups of lesson in progress

## Quiz

I cleaned up the spacing and text styles of this section, and applied the green in the "Complete quiz" button. Also, the correct and incorrect messages were updated with the color scheme used in the callout component implemented across the site.

Mockups of quiz without responses and quiz responded

## Facilitator notes

The facilitator notes are under a link in the header that opens/closes a new sidebar on the right side of the page. The spacing applied differs from that of the left sidebar to convey that the content is in a secondary hierarchy.

Mockups of lesson with facilitator notes closed and opened

---

## Feedback

These designs are not written in stone, so please share all your feedback.

[Here](https://www.figma.com/design/ZKaf6u8mIHkAanluWafXAp/Learn?node-id=5514-15445&node-type=SECTION&m=dev) are the components used in the section. There are also the messages with pending texts and others written by me that need revision from the experts in the project.

I hope this proposal meets you all expectations ✨

> [!NOTE]
> Pinging the folks involved in the tickets referred: @renintw @adamwoodnz @StevenDufresne @kathrynwp @Clorith @jonathanbossenger @westnz @ndiego

outdoor2kode:
Resolves #2869

This PR introduces the UI for Facilitator Notes, including updates for mobile view. Additionally, the Breadcrumbs section has been changed from "Home" to "Learn WordPress." 

**Screencast**

https://github.com/user-attachments/assets/06290332-96ea-4d72-8bfe-a1519778a57b

**Lesson Plan Migration**
From the discussion at https://github.com/WordPress/Learn/issues/2392, it appears that before migrating Lesson Plans to Facilitator Notes, it needs to map each Lesson Plan to the most appropriate lesson. Has there been any related discussion?


fcoveram:
The flow looks good to me. I'd love to hear @jonathanbossenger and @kathrynwp's thoughts. 

I would also like to see if we can test this view with actual use cases. [This post](https://learn.wordpress.org/lesson-plan/create-a-basic-child-theme-for-block-themes/) was shared before as an example of how long content could be, but it's vital to identify whether that page is an exception or not.

It would be frustrating to have a long modal, but also a poor user experience needing to open an external link to see short content that could perfectly fit into a modal.

adamwoodnz:
> **Lesson Plan Migration**
> From the discussion at #2392, it appears that before migrating Lesson Plans to Facilitator Notes, it needs to map each Lesson Plan to the most appropriate lesson. Has there been any related discussion?

No further discussion that I'm aware of. I'm imagining a new post meta field on a Lesson for selecting a Lesson Plan. The content from that Lesson Plan is rendered in the modal.

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-header.php
adamwoodnz, line 1: We'll need to apply these changes to `patterns/sensei-lesson-standalone.php` too

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -26,26 +53,153 @@ body.sensei {
...
+		&:hover {
+			text-decoration: none;
+		}
+
+		@media (max-width: 782px) {
+			display: none;
+		}
+	}
+
+	.wporg-learn-facilitator-notes-content {
```
adamwoodnz, line 125: I notice that the modal content overflow is hidden. Some of the Lesson Plan content is very long so this will need to be scrollable on desktop as well as mobile.

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-facilitator-notes.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Title: Sensei Lesson Facilitator Notes
+ * Slug: wporg-learn-2024/sensei-lesson-facilitator-notes
+ * Inserter: no
+ */
+
+?>
+
+<input type="checkbox" id="wporg-learn-facilitator-notes-toggle">
```
adamwoodnz, line 10: Need to consider the accessibility of this control and the modal content. At present this input isn't focusable with keyboard. We may want to move focus to the close button after activation. I'd consult some [modal best practice docs](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/).

On mobile it is definitely following a modal pattern, where focus should be trapped within the modal, and background page elements should not be interactive. On desktop it's arguable that the rest of the UI should remain interactive, as it's more of a panel.

@ryelle any thoughts on this?

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
```diff
@@ -26,26 +53,153 @@ body.sensei {
...
+				cursor: pointer;
+			}
+
+			@media (max-width: 782px) {
+				position: absolute;
+				top: var(--wp--preset--spacing--10);
+				left: 4%;
+				width: 92%;
+				height: 90vh;
+				border: 1px solid var(--sensei-course-progress-bar-color);
```
adamwoodnz, line 89: Seeing a grey border around the modal on mobile, I don't think this matches modals on other sites, eg. Meeting Calendar

| Facilitator notes | Facilitator notes inspected | Meeting Calendar |
|--------|--------|-|
| Screenshot 2024-10-17 at 11 36 15 AM | Screenshot 2024-10-17 at 11 36 24 AM | Screenshot 2024-10-17 at 11 34 56 AM |

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-facilitator-notes.php
adamwoodnz: Also I expected this pattern would use a button element, hooked up to the interactivity API to control the modal visibility, however I'm not sure whether this input based approach is any less accessible. Might need more aria attributes though as I think it's a less typical pattern.

adamwoodnz (changes requested):
Nice start! Visually the UI is working well. I've left some comments about accessibility concerns.

If we want to load the content from a linked Lesson Plan I guess we'll need a block with conditional rendering to handle the Lesson Plan being defined or erroring, etc.

wp-content/themes/pub/wporg-learn-2024/patterns/sensei-lesson-facilitator-notes.php
ryelle: > I expected this pattern would use a button element,

I think a button (with `aria-expanded` and `aria-controls` attributes) would be better. Right now this would read as checking a checkbox, with no way to know that it would reveal more content.

> On mobile it is definitely following a modal pattern, where focus should be trapped within the modal, and background page elements should not be interactive. On desktop it's arguable that the rest of the UI should remain interactive, as it's more of a panel.

If it's easier, you can make it a modal (with focus trapping) in both cases - and if someone clicks outside the panel to use the rest of the site, close the panel. Otherwise, on desktop, I would expect it to close when I tab through everything, and focus moves on to the next tab stop (from the screenshots, I assume that's "exit course"). The desktop global nav submenus work like this— if you open "Community" & start tabbing, the submenu closes when  "About" is focused.

jonathanbossenger:
> I would also like to see if we can test this view with actual use cases. [This post](https://learn.wordpress.org/lesson-plan/create-a-basic-child-theme-for-block-themes/) was shared before as an example of how long content could be, but it's vital to identify whether that page is an exception or not.
> 
> It would be frustrating to have a long modal, but also a poor user experience needing to open an external link to see short content that could perfectly fit into a modal.

I understand your concerns. I've selected a sampling of current lesson plans from the [lesson plan landing page](https://learn.wordpress.org/lesson-plans/)

- [Create a basic child theme](https://learn.wordpress.org/lesson-plan/create-a-basic-child-theme-for-block-themes/)
- [Upload a theme to the theme directory](https://learn.wordpress.org/lesson-plan/upload-a-theme-to-the-wordpress-repository/)
- [Intro to the Site Editor](https://learn.wordpress.org/lesson-plan/intro-to-site-editor-and-templates/)

Everything up to the "Example Lesson" section would need to be displayed in the lesson notes module, and as you can see in all three examples, even the shortest one (Upload a theme) is pretty detailed.

fcoveram:
Thanks @jonathanbossenger for adding more examples.

I think we should revisit this decision and, instead of showing the content in a modal, make the "Facilitator notes" link to open a new page.

Although the practice is not recommended ([G200](https://www.w3.org/TR/WCAG20-TECHS/G200.html)), I would suggest opening a new tab. Filling out forms or (quote) "prevent the loss of any form data that has already been entered" are use cases excluded from this requirement.

What do you think of this? Perhaps @ryelle knows more about it.

ryelle:
> I think we should revisit this decision and, instead of showing the content in a modal, make the "Facilitator notes" link to open a new page.
> 
> Although the practice is not recommended ([G200](https://www.w3.org/TR/WCAG20-TECHS/G200.html)), I would suggest opening a new tab. Filling out forms or (quote) "prevent the loss of any form data that has already been entered" are use cases excluded from this requirement.

Is there progress that a user can loose? Lessons are mostly articles, as I recall from past testing. Clicking a link and then using the back button wouldn't cause "lose of data" when reading an article. Would this show up on a quiz, or are there many interactive lessons?

If there are, and this would be a "lose of data" case, the G200 technique also recommends giving advance warning that this will be a new tab (typically this is done with the ↗️ icon, but we're already using that for external links). The indicator would need to be visible and screen-reader-accessible.

Otherwise, I think it would be fine to keep it as a new page and let users open it as they choose.

wp-content/themes/pub/wporg-learn-2024/src/style/_sensei.scss
outdoor2kode: Thanks. c125bcb

outdoor2kode:
> If we want to load the content from a linked Lesson Plan I guess we'll need a block with conditional rendering to handle the Lesson Plan being defined or erroring, etc.

I originally thought of keeping it simple and achieving it with CSS. If the content can't be retrieved, it could be handled within the pattern and display the corresponding notice/warning. That said, the requirements have changed now, I'll think about how to make adjustments for a better approach.

adamwoodnz:
> Is there progress that a user can loose? Lessons are mostly articles, as I recall from past testing. Clicking a link and then using the back button wouldn't cause "lose of data" when reading an article. Would this show up on a quiz, or are there many interactive lessons?

I don't believe there is any state to lose, and Lessons Plans don't relate to any quizzes afaik. I feel like a new tab/window seems like a good approach, with sufficient warning, so that the plan can be referenced alongside the lesson. That seems to be the intention.

kaitohm:
> I feel like a new tab/window seems like a good approach, with sufficient warning, so that the plan can be referenced alongside the lesson. That seems to be the intention.

This front-end implementation of a new tab would work. The team's request is more around the back-end implementation. quoting https://github.com/WordPress/Learn/issues/2392#issuecomment-2181675301

> The desired end state [is]... The information of the lesson plan would be stored in a custom field / block within the lesson editor.
> 
> This is in order to reduce the need for us to have to edit duplicative information across different post types, and instead manage all information in one lesson. This also will help us with streamlining translation work.

Would this still be possible?

adamwoodnz:
> > The desired end state [is]... The information of the lesson plan would be stored in a custom field / block within the lesson editor.
> > This is in order to reduce the need for us to have to edit duplicative information across different post types, and instead manage all information in one lesson. This also will help us with streamlining translation work.
> 
> Would this still be possible?

Thanks for reminding us of that requirement. While I'm not sure it will help streamline translation work (that might depend on which translation plugin we choose and how well it plays with custom fields), the ease of editing in one place is not to be overlooked.

I'm not sure how well a custom fields approach will handle content of this scale either. This seems to be the main challenge for both frontend and backend. Might need some testing.

A custom Facilitators Notes block within the post editor might be worth trying too.

Both these approaches will very likely require manual migration I expect.

kaitohm:
> Both these approaches will very likely require manual migration I expect.

The team will be fine with that 👍  I believe we expected that would be the case.

outdoor2kode:
Closed in favor of https://github.com/WordPress/Learn/pull/3052

PR #3049, opened 2024-12-02
Adds the Polylang Pro plugin to Learn.WordPress.org

### LINKED ISSUE #2284: Test Polylang
Front-end requirements:

* Language selector **does not show** as country flags. (Many countries speak multiple languages, so describing a language with a flag should be avoided.)
* Languages can be changed without needing to log in to one's WordPress.org account.
* Once a language is selected, it is applied globally.
* Content of the selected language is shown above non-translated content.
* English content is still shown where translations don't exist.

Back-end requirements:
* Translators do not have permission to create or publish new content.
* Translators have permission to both translate any content, and review content translated by others.
* A flow can be implemented where translations must be approved (reviewed) before published.
* Sensei content types can be translated in the same way as native post types.
* Taxonomies can be translated.
* Translated content is indexed and searchable by Jetpack Search.

jonathanbossenger:
See #2284


adamwoodnz:
I didn't get very far trying to test this. Activation didn't seem to run as expected and I now get an error `Notice: Undefined index: default_lang in /var/www/html/wp-content/plugins/polylang-pro/vendor/wpsyntex/polylang/include/links-directory.php on line 63`

Might need to wipe my DB and start fresh unless you know an easy fix @sebastienserre?

sebastienserre:
Hello,

Did you follow the wizard, which is automatically launched at activation?
You can follow it by going to the Languages > setup menu or creating some languages (at least one) & set it as the default language:

https://polylang.pro/doc/configure-the-languages/#default-language

adamwoodnz:
> Did you follow the wizard, which is automatically launched at activation? 

Hi,

The wizard launched but something went wrong, and I think I know why now.

I've managed to fix the default language issue by deleting all languages and adding en_US back.

Now when I visit the wizard I can see that some frontend assets are missing (CSS and JS), so I think this caused my path through the wizard to fail.

Screenshot 2024-12-03 at 8 04 44 PM

@jonathanbossenger could you please check that all the files have been committed?

jonathanbossenger:
Hi folks.

@adamwoodnz I was pretty sure everything was added, but when I checked the files you pointed out in your screenshot, PhpStorm warned me that I was viewing a file that was being ignored.

I manually compared what was committed to what was in the `polylang-pro.zip`, and I've added `.gitignore` exceptions for the `css/build` and `js/build` directories, as well as adding a `.gitkeep` file to the `languages` directory, so that directory gets committed as well. 

I'm not sure why adding `!/wp-content/plugins/polylang-pro` to the `.gitignore` didn't also include the `css/build` and the `js/build` directories? I can only think that the `/**/build` directive higher up in the `.gitignore` might cause this.

To make sure, I then also ran a diff comparing the two directories and confirmed they both contain the same content (except for the `languages/.gitkeep` file, which is only added to git)

jonathanbossenger:
Thinking about this further, the missing `languages` directory is probably why the initial activation failed, as it didn't have anywhere to store the language files.

adamwoodnz:
> I can only think that the `/**/build` directive higher up in the `.gitignore` might cause this.

Correct 🙂 

I've added the frontend assets to this PR, now that they're not ignored, and the plugin seems to run as expected locally now. I haven't tested any specific translation functionality, but I trust that is working as desired.

I'll get this merged and deployed so we can activate it on https://learn.wordpress.org/test/, to start with.

adamwoodnz (approved):
LGTM

kaitohm:
Thanks all! Let me know when this is committed, and I'll start testing on the test site.

PR #3053, opened 2024-12-04
Create wpml-config.xml

adamwoodnz:
Adds a WPML config file for Polylang Pro, see [comment](https://github.com/WordPress/Learn/issues/2284#issuecomment-2511099776).

See #2284

### DIFF
--- a/wp-content/themes/pub/wporg-learn-2024/wpml-config.xml
+++ b/wp-content/themes/pub/wporg-learn-2024/wpml-config.xml
@@ -1,5 +1,5 @@
 <wpml-config>
-    <custom-fields>
-        <custom-field action="ignore">_lesson_course</custom-field>
-    </custom-fields>
+	<custom-fields>
+		<custom-field action="ignore">_lesson_course</custom-field>
+	</custom-fields>
 </wpml-config>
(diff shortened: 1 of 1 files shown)


jonathanbossenger (approved):
LGTM

adamwoodnz:
How do I test this please @jonathanbossenger?

jonathanbossenger:
@adamwoodnz the way I've been testing it is to try and translate a lesson that's part of a module in a course.

[In this comment](https://github.com/WordPress/Learn/issues/2284#issuecomment-2359399610), you'll see that when I try to translate the lesson, the relationship between the lesson and its parent module is broken.

After applying the XML config file, that relationship is retained after translation, see example of this [in action here](https://github.com/WordPress/Learn/issues/2284#issuecomment-2459070351). 

So essentially, when the XML file is in place, it should be possible to translate a lesson that's part of a module and course, without the relationships between the module and course being lost.

adamwoodnz:
I've done some testing and I did observe some weirdness without the config, but I wasn't able to consistently reproduce your issue. Maybe the order of operations matters (I added the US lesson to a module, and then added a translation through Polylang UI), or maybe the config was cached when switching between prod and my sandbox. Anyway I will merge this and you can check the test site to make sure it's behaving.

PR #3125, opened 2025-03-31
Update taxonomy.php

pkevan:
Avoid fatal error for WP core function in namespace.

```
E_ERROR: Uncaught Error: Call to undefined function WPOrg_Learn\Taxonomy\_get_list_table() in [..]/plugins/wporg-learn/inc/taxonomy.php:685
```

cc @adamwoodnz


adamwoodnz:
Thanks Paul

Issue #3226, opened 2025-06-30
Feedback - Estimate time to complete a course has disappeared


peiraisotta:
Hello team! 

There are a few courses that don’t indicate the estimated completion time, and I need that information while drafting the activities for students who will enroll in a contribution internship with the WordPress Foundation.

Following @jonathanbossenger 's guidance in [this Slack thread](https://wordpress.slack.com/archives/C02RW657Q/p1751278235614829), I'm opening this issue to kindly request to review the content of the following courses, manually estimate the time to completion, and then enter that value.

- [Open source basics and WordPress](https://learn.wordpress.org/course/open-source-basics-and-wordpress/)
- [How decisions are made in the WordPress project](https://learn.wordpress.org/course/how-decisions-are-made-in-the-wordpress-project/)
- [Community meeting etiquette](https://learn.wordpress.org/course/community-meeting-etiquette/)
- [Writing in the WordPress voice](https://learn.wordpress.org/course/writing-in-the-wordpress-voice/)
- [Basic principles of conflict resolution](https://learn.wordpress.org/course/basic-principles-of-conflict-resolution/)

Thank you so much, please let me know if you need any other info or context from me.

devmuhib009:
I've added this on meeting agenda, hope we will get some feedback soon!

melbos:
@devmuhib009 @rjekic @peiraisotta  Approx. Estimates:

https://learn.wordpress.org/course/open-source-basics-and-wordpress/

- Quick Completion: About 2 hours.
- Solid Understanding with Extras: Around 3–4 hours.
- Hands-On Involvement: 5 hours or more.


https://learn.wordpress.org/course/how-decisions-are-made-in-the-wordpress-project/

- Minimum: 1.5 hours
- With extra reading: 2.5 hours
- Adding practice / contribution: 4–5+ hours

devmuhib009:
Hi @melbos 
Thanks for sharing your opinion. Actually, we need real data, you can share feedback for 1 course, but try to complete the course and share actual time you needed to complete the course.

melbos:
@devmuhib009 I have completed them and it took me minumum of invested time I proposed

devmuhib009:
Ah! You completed so many courses, I just checked your profile. Thanks @melbos for sharing the feedback! ❤️

rjekic:
First of all, thank you @melbos for actively participating in this issue. The data you provided is valuable and much appreciated.

[...]

PR #3240, opened 2025-07-11
Added course duration per request from #3226

rjekic:
I've added a course duration field and made it part of the sidebar meta fields block. I am aware that there is a Learning Duration block already created (I used part of the code from that block), but if we add that block to the single course template, it will be displayed even if the duration meta field in the course settings sidebar is not populated. 

@jonathanbossenger @devmuhib009


jonathanbossenger:
>  if we add that block to the single course template, it will be displayed even if the duration meta field in the course settings sidebar is not populated.

I hope you don't mind my asking, but I'm interested in understanding the use case for having a fallback. Why would we not just use the Duration in hours field?

rjekic:
@jonathanbossenger 

In my previous comment, I was comparing the difference between using the Learning Duration block (which we are currently using) and the code I added in this PR. Both of them are using data from the Duration in hours field; however, if the field is not populated, the block will still be visible on the frontend, and the code I added in the PR will not be visible in that specific case.

Successfulsebunya (approved):
Nicely done. Thank you @rjekic

theaminulai (approved):
This pull request is ready to be merged and the code is good to go.

PR #3592, opened 2026-08-12
Activity kits: server-side download tracking

Piyopiyo-Kitsune:
## Problem

`get_clicks()` in Jetpack Stats only tracks **outbound** (external-domain) clicks. Since the activity kit ZIP files are hosted on `learn.wordpress.org` (same domain), they are never captured — download counts always showed 0.

## Solution

Replace the Jetpack click approach with a thin REST redirect endpoint that increments a post meta counter before forwarding the browser to the actual file.

```
GET /wp-json/activity-kits/v1/download/{slug}
→ 302 Location: https://learn.wordpress.org/.../activity-kit.zip
```

This works regardless of file host domain, requires no JS, and persists across page loads.

### `inc/activity-kit-rest.php`
- **New route** `GET activity-kits/v1/download/{slug}` (public, no auth required)
  - Resolves the kit by slug, fetches the ZIP attachment URL
  - Increments `_activity_download_count` post meta
  - Returns `WP_REST_Response( null, 302, [ 'Location' => $zip_url ] )`
- **Fix view stats** — switch from `get_top_posts()` to `get_total_post_views()` so recently published kits with low traffic aren't omitted from results (same fix as #3591, included here since this branch is off trunk)
- **Remove** `get_jetpack_download_clicks()` — now dead code
- **Stats endpoint** reads downloads from `_activity_download_count` post meta directly, removing the Jetpack clicks dependency entirely

### `inc/post-meta.php`
- Register `_activity_download_count` meta (integer, default 0, `show_in_rest => false` — admin-read-only via REST stats endpoint)

### `patterns/single-activity-kit-content.php`
- Both download buttons now point at the `/download/{slug}` endpoint
- Removed the `window._stq` JS tracking block — it was tracking clicks to internal REST URLs (meaningless), and server-side counting replaces it entirely

## Testing

1. Load any published activity kit page
2. Click **Download kit** — you should be redirected to the ZIP file
3. Visit the [Stats page](https://learn.wordpress.org/wp-admin/edit.php?post_type=activity_kit&page=activity-kit-stats) — the download count for that kit should increment

> 🤖 Written with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
+++ b/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
@@ -71,11 +71,6 @@ function register_routes() {
 /**
  * Get the tracked download URL for an activity kit.
  *
- * Templates must link kit ZIPs through this endpoint (rather than the raw
- * attachment URL) so the download is counted before the redirect. Keeping the
- * route path in one place next to its registration means a route change cannot
- * leave a template linking to a 404.
- *
  * @param int $kit_id Post ID of the activity kit.
  * @return string     URL of the counting download endpoint.
  */
@@ -151,15 +146,9 @@ function handle_download( $request ) {
 
 	if ( ! $is_bot ) {
 		/*
-		 * Downloads are stored in one meta row per kit per UTC day so the stats
-		 * endpoint can sum them over the same day span as the Jetpack view
-		 * ranges (see get_download_counts()). Seed today's bucket ($unique=true
-		 * is a no-op once it exists), then increment with a direct UPDATE —
-		 * update_post_meta()'s $prev_value CAS drops its WHERE clause when the
-		 * previous value is 0, so two concurrent downloads would both write 1.
-		 * If two first-downloads of the day race the seed into duplicate rows,
-		 * every UPDATE increments both and reads take MAX per bucket, so no
-		 * download is lost.
+		 * Seed today's bucket, then increment atomically — update_post_meta()'s
+		 * CAS is racy when the previous value is 0, and a seed race is harmless
+		 * because reads take MAX per bucket.
 		 */
 		$meta_key = '_activity_downloads_' . gmdate( 'Ymd' );
 		add_post_meta( $kit_post->ID, $meta_key, 0, true );
@@ -267,12 +256,8 @@ function handle_stats( $request ) {
 /**
  * Get the day span a stats range covers.
  *
- * Single source of truth for both metrics: get_jetpack_post_views() derives
- * its API windows from this and get_download_counts() sums bucket rows over
- * it, so the download rate always divides two figures covering the identical
- * period. 'all' caps at 180 days — each extra 30-day window is another
- * sequential API call, and 6 is a reasonable ceiling for an admin-only
- * dashboard with a small post count.
+ * Shared by get_jetpack_post_views() and get_download_counts() so views and
+ * downloads always cover the same period.
  *
  * @param string $range One of '7d', '30d', '90d', 'all'.
  * @return int          Number of days the range covers.
--- a/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
+++ b/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
@@ -144,10 +144,22 @@ function handle_download( $request ) {
 			)
 		);
 		if ( 0 === $updated ) {
-			// No row yet — insert with an initial count of 1.
-			// $wpdb->query() returns false on DB error and 0 when no rows matched;
-			// strict comparison avoids falling into this branch on a real error.
-			add_post_meta( $kit_post->ID, '_activity_download_count', 1, true );
+			// No row yet (e.g. a kit published before the pre-seeding hook was
+			// added). Insert with an initial count of 1. If a concurrent
+			// first-download wins the INSERT race, add_post_meta() returns false
+			// ($unique=true blocks the duplicate); re-run the UPDATE so this
+			// download is still counted.
+			$inserted = add_post_meta( $kit_post->ID, '_activity_download_count', 1, true );
+			if ( ! $inserted ) {
+				// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Atomic fallback after concurrent INSERT collision; cache invalidated immediately below.
+				$wpdb->query(
+					$wpdb->prepare(
+						"UPDATE {$wpdb->postmeta} SET meta_value = meta_value + 1 WHERE post_id = %d AND meta_key = %s",
+						$kit_post->ID,
+						'_activity_download_count'
+					)
+				);
+			}
 		}
 		wp_cache_delete( $kit_post->ID, 'post_meta' );
 	}
--- a/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
+++ b/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
@@ -69,6 +69,21 @@ function register_routes() {
 }
 
 /**
+ * Get the tracked download URL for an activity kit.
+ *
+ * Templates must link kit ZIPs through this endpoint (rather than the raw
+ * attachment URL) so the download is counted before the redirect. Keeping the
+ * route path in one place next to its registration means a route change cannot
+ * leave a template linking to a 404.
+ *
+ * @param int $kit_id Post ID of the activity kit.
+ * @return string     URL of the counting download endpoint.
+ */
+function get_download_url( $kit_id ) {
+	return rest_url( 'activity-kits/v1/download/' . absint( $kit_id ) );
+}
+
+/**
  * Shorten Jetpack's stats API cache so the activity kit dashboard tracks
  * WordPress.com's near-real-time counts more closely. Registered only for the
  * duration of the activity kit stats REST request (see handle_stats()), so it
@@ -109,7 +124,11 @@ function handle_download( $request ) {
 	 * download, not a failed one.
 	 */
 	$user_agent = sanitize_text_field( wp_unslash( $_SERVER['HTTP_USER_AGENT'] ?? '' ) );
-	$is_bot     = empty( $user_agent ) || preg_match( '/bot|crawl|slurp|spider|mediapartners|facebookexternalhit|linkedinbot|twitterbot|whatsapp|slack|discord|prefetch/i', $user_agent );
+	// Browsers signal speculative loads via headers, not the User-Agent: Sec-Purpose (standard), Purpose (WebKit), X-moz (older Firefox).
+	$purpose = sanitize_text_field( wp_unslash( $_SERVER['HTTP_SEC_PURPOSE'] ?? $_SERVER['HTTP_PURPOSE'] ?? $_SERVER['HTTP_X_MOZ'] ?? '' ) );
+	$is_bot  = empty( $user_agent )
+		|| preg_match( '/bot|crawl|slurp|spider|mediapartners|facebookexternalhit|linkedinbot|twitterbot|whatsapp|slack|discord|prefetch/i', $user_agent )
+		|| preg_match( '/prefetch|prerender|preview/i', $purpose );
 
 	$kit_id   = absint( $request->get_param( 'id' ) );
 	$kit_post = get_post( $kit_id );
@@ -246,6 +265,32 @@ function handle_stats( $request ) {
 }
 
 /**
+ * Get the day span a stats range covers.
+ *
+ * Single source of truth for both metrics: get_jetpack_post_views() derives
+ * its API windows from this and get_download_counts() sums bucket rows over
+ * it, so the download rate always divides two figures covering the identical
+ * period. 'all' caps at 180 days — each extra 30-day window is another
+ * sequential API call, and 6 is a reasonable ceiling for an admin-only
+ * dashboard with a small post count.
+ *
+ * @param string $range One of '7d', '30d', '90d', 'all'.
+ * @return int          Number of days the range covers.
+ */
+function get_range_days( $range ) {
+	switch ( $range ) {
+		case '7d':
+			return 7;
+		case '30d':
+			return 30;
+		case '90d':
+			return 90;
+		default:
+			return 180;
+	}
+}
+
+/**
  * Get per-post view counts from Jetpack Stats for a given time range.
  *
  * Uses get_total_post_views() to query specific post IDs directly, rather than
@@ -273,74 +318,17 @@ function get_jetpack_post_views( $range, array $kit_ids ) {
 	$stats = new \Automattic\Jetpack\Stats\WPCOM_Stats();
 
 	/*
-	 * Map the UI range to one or more 30-day windows. Each window is defined by
-	 * how many days back its end-date is offset from today. 'all' uses 6 windows
-	 * (≈ 6 months) rather than 3, giving a meaningful distinction from '90d'.
-	 * Extending further would multiply sequential API calls proportionally; 6 is
-	 * a reasonable ceiling for an admin-only dashboard with a small post count.
+	 * Break the range's day span into windows of at most 30 days — the WPCOM
+	 * /stats/views/posts API caps `num` at 30 per call. Each window is defined
+	 * by how many days back its end-date is offset from today.
 	 */
-	switch ( $range ) {
-		case '7d':
-			$windows = array(
-				array(
-					'num'    => 7,
-					'offset' => 0,
-				),
-			);
-			break;
-		case '30d':
-			$windows = array(
-				array(
-					'num'    => 30,
-					'offset' => 0,
-				),
-			);
-			break;
-		case '90d':
-			$windows = array(
-				array(
-					'num'    => 30,
-					'offset' => 0,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 30,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 60,
-				),
-			);
-			break;
-		case 'all':
-		default:
-			$windows = array(
-				array(
-					'num'    => 30,
-					'offset' => 0,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 30,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 60,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 90,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 120,
-				),
-				array(
-					'num'    => 30,
-					'offset' => 150,
-				),
-			);
-			break;
+	$days    = get_range_days( $range );
+	$windows = array();
+	for ( $offset = 0; $offset < $days; $offset += 30 ) {
+		$windows[] = array(
+			'num'    => min( 30, $days - $offset ),
+			'offset' => $offset,
+		);
 	}
 
 	// Pre-compute window end-dates once so every chunk uses the same calendar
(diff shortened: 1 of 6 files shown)


Piyopiyo-Kitsune:
## Context on the tracking approach

This PR uses a REST redirect endpoint for download tracking, which is architecturally different from what dd32 flagged as non-viable in #3496 — but similar enough that I want to flag it proactively rather than wait for a review surprise.

**What was rejected in #3496**: A `POST /activity-kits/v1/track` endpoint called client-side as a fire-and-forget beacon — an *extra* REST request layered on top of the real action. Dion's fix was to use Jetpack Stats instead, which #3496 shipped with.

**Why Jetpack still doesn't capture downloads**: Dion noted in #3496 that *"The dedicated File Downloads report is structurally unavailable for this site"* because WordPress.org serves uploads from `/files/` rather than `/wp-content/uploads/`, so Jetpack's automatic click tracker skips them entirely. The `_stq.push()` beacon approach that shipped in #3496 has the same limitation — `get_clicks()` in `WPCOM_Stats` only returns outbound (external-domain) clicks, so same-domain file downloads never register. That's why the download count has shown 0 since launch.

**How this PR differs**: The REST endpoint here *is* the download mechanism — both download buttons now point at `/wp-json/activity-kits/v1/download/{slug}`, which increments a post meta counter and 302-redirects to the file. There's no extra tracking request; the user can't reach the file without going through the endpoint. The counter is a single integer per kit (not a row-per-event table), so the storage footprint is minimal.

**The open question**: This still bootstraps WordPress + REST + a `update_post_meta` write on every download. That's lighter than the original rejected approach, but heavier than a pure Jetpack solution. @dd32 — given that Jetpack's click tracker can't cover `/files/` paths, is a REST redirect endpoint like this acceptable on WordPress.org infrastructure, or would you prefer a different mechanism (e.g., a plain PHP handler via rewrite rule that bypasses the REST bootstrap)?

Happy to rework the approach based on feedback before this gets further into review.

> 🤖 Written with [Claude Code](https://claude.ai/claude-code)

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -66,6 +83,52 @@ function stats_cache_expiration( $expiration ) {
...
+
+	$zip_url = wp_get_attachment_url( $zip_id );
+
+	if ( ! $zip_url ) {
+		return new \WP_Error( 'activity_kit_zip_url', __( 'Could not resolve the download URL.', 'wporg-learn' ), array( 'status' => 500 ) );
+	}
+
+	// Increment the download counter stored in post meta.
+	$current_count = (int) get_post_meta( $kit_post->ID, '_activity_download_count', true );
+	update_post_meta( $kit_post->ID, '_activity_download_count', $current_count + 1 );
```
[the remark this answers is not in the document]
Piyopiyo-Kitsune: Fixed — switched to a compare-and-swap retry loop using `update_post_meta()`'s `$prev_value` parameter. The loop re-reads the current value and retries up to 5 times if a concurrent write landed between the read and the write.

wp-content/themes/pub/wporg-learn-2024/patterns/single-activity-kit-content.php
```diff
@@ -7,10 +7,12 @@
...
-$kit_id   = get_the_ID();
-$duration = get_post_meta( $kit_id, '_activity_duration', true );
-$zip_id   = (int) get_post_meta( $kit_id, '_activity_zip_id', true );
-$zip_url  = $zip_id ? wp_get_attachment_url( $zip_id ) : '';
+$kit_id       = get_the_ID();
+$duration     = get_post_meta( $kit_id, '_activity_duration', true );
+$zip_id       = (int) get_post_meta( $kit_id, '_activity_zip_id', true );
+$zip_url      = $zip_id ? wp_get_attachment_url( $zip_id ) : '';
+// Use the server-side download endpoint so every click increments the post meta counter.
+$download_url = $zip_id ? rest_url( 'activity-kits/v1/download/' . get_post_field( 'post_name', $kit_id ) ) : '';
```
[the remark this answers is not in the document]
Piyopiyo-Kitsune: Removed. `$zip_url` was the only remaining reference; since both buttons now point at the REST redirect endpoint, it was never used in output.

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -1046,4 +1046,17 @@ function register_activity_kit_meta() {
...
+		'activity_kit',
+		'_activity_download_count',
+		array(
+			'description'       => __( 'Number of times this activity kit ZIP has been downloaded via the tracked download endpoint.', 'wporg-learn' ),
+			'type'              => 'integer',
+			'single'            => true,
+			'default'           => 0,
+			'sanitize_callback' => 'absint',
+			'show_in_rest'      => false,
+		)
```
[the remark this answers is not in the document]
Piyopiyo-Kitsune: Added `'auth_callback' => $auth_callback` — consistent with the other activity_kit meta keys in this function.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -66,6 +83,52 @@ function stats_cache_expiration( $expiration ) {
...
+
+	return new \WP_REST_Response( null, 302, array( 'Location' => $zip_url ) );
```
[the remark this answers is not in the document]
Piyopiyo-Kitsune: Fixed — wrapped the attachment URL in `esc_url_raw()` before setting the Location header.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -142,7 +184,7 @@ function handle_stats( $request ) {
 				$data['views'] = $views_map[ $kit_post->ID ] ?? 0;
 			}
 			if ( 'both' === $metric || 'downloads' === $metric ) {
-				$data['downloads'] = $downloads_map[ $kit_post->ID ] ?? 0;
+				$data['downloads'] = (int) get_post_meta( $kit_post->ID, '_activity_download_count', true );
```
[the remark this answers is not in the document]
Piyopiyo-Kitsune: Fixed — the `$jetpack_unavailable` gate now only affects views. Downloads are read from `_activity_download_count` post meta in all cases, so they continue to show real counts even when Jetpack is deactivated.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -44,6 +44,23 @@ function register_routes() {
 			),
 		)
 	);
+
+	register_rest_route(
+		'activity-kits/v1',
+		'/download/(?P<slug>[a-z0-9-]+)',
```
obenland: This route regex is narrower than what WordPress allows in `post_name`, so some kits get a dead Download button (REST 404 JSON instead of the ZIP) — a regression vs. the old direct link.

`sanitize_title_with_dashes()` keeps underscores, and non-Latin titles become percent-encoded slugs (`utf8_uri_encode()`) — relevant here since activity kits are explicitly multilingual and the bulk importer sets `post_title` with no slug control. A kit slugged `trivia_night` or `%e3%81%82…` renders an `href` that `WP_REST_Server` never matches, and the user gets `rest_no_route`. The `sanitize_callback => 'sanitize_title'` can't help — sanitization runs only *after* a route matches.

Fix options: widen to `(?P<slug>[^/]+)` and `rawurlencode()` the slug when building the URL, or route on post ID instead.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -66,6 +83,58 @@ function stats_cache_expiration( $expiration ) {
...
+ * Handle GET /activity-kits/v1/download/{slug}
+ *
+ * Increments the kit's download counter (stored in post meta) and redirects
+ * the browser to the actual ZIP file URL. Using a server-side redirect lets
+ * us track same-domain downloads that Jetpack's outbound-click tracker misses.
+ *
+ * @param \WP_REST_Request $request The REST request.
+ * @return \WP_REST_Response|\WP_Error 302 redirect on success, WP_Error on failure.
+ */
+function handle_download( $request ) {
```
obenland, line 102: This is an unauthenticated GET that performs a state write, with no bot/prefetch/dedup guard — and GET-with-side-effects is exactly what entitles intermediaries to prefetch it. Crawlers, Slack/Discord link unfurlers, email scanners, and browser prefetch/speculation rules all issue plain GETs on the button `href`, and each one increments the counter, so counts can inflate arbitrarily.

Also, the 302 goes out to anonymous visitors with zero cache headers (WP core only sends nocache headers when `is_user_logged_in()`). A 302 isn't heuristically cacheable per RFC 9110, but sending `nocache_headers()` before returning the redirect would make that explicit rather than dependent on edge config. Basic bot filtering/dedup is worth considering too.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -66,6 +83,58 @@ function stats_cache_expiration( $expiration ) {
...
+		return new \WP_Error( 'activity_kit_zip_url', __( 'Could not resolve the download URL.', 'wporg-learn' ), array( 'status' => 500 ) );
+	}
+
+	// Increment the download counter stored in post meta.
+	// Use a compare-and-swap retry loop (update_post_meta's $prev_value arg) to avoid
+	// lost increments when two requests arrive simultaneously.
+	$retries = 0;
+	do {
+		$current_count = (int) get_post_meta( $kit_post->ID, '_activity_download_count', true );
+		$updated       = update_post_meta( $kit_post->ID, '_activity_download_count', $current_count + 1, $current_count );
```
obenland: This compare-and-swap loop unfortunately doesn't prevent lost increments in either scenario it targets:

1. **`$prev_value = 0` degrades to a blind write.** `update_metadata()` only adds the `meta_value` WHERE clause `if ( ! empty( $prev_value ) )`, so the 0→1 increment is unconditional — two concurrent first downloads both write `1`.
2. **For counts > 0, a genuine CAS failure can't recover.** When `$wpdb->update` matches zero rows it returns before `wp_cache_delete`, so the loser's `get_post_meta()` re-reads the same stale cached value on every retry — all 5 iterations issue the identical failing UPDATE and the increment is silently dropped.

A single atomic query avoids both:

```php
$updated = $wpdb->query( $wpdb->prepare(
	"UPDATE {$wpdb->postmeta} SET meta_value = meta_value + 1 WHERE post_id = %d AND meta_key = %s",
	$kit_post->ID,
	'_activity_download_count'
) );
if ( ! $updated ) {
	add_post_meta( $kit_post->ID, '_activity_download_count', 1, true );
}
wp_cache_delete( $kit_post->ID, 'post_meta' );
```

(The comment above the loop would need updating to match, since the concurrency claim it documents doesn't hold.)

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -133,17 +181,17 @@ function handle_stats( $request ) {
...
+				$data['views']               = 0;
+			} else {
 				$data['views'] = $views_map[ $kit_post->ID ] ?? 0;
 			}
-			if ( 'both' === $metric || 'downloads' === $metric ) {
-				$data['downloads'] = $downloads_map[ $kit_post->ID ] ?? 0;
-			}
+		}
+		if ( 'both' === $metric || 'downloads' === $metric ) {
+			$data['downloads'] = (int) get_post_meta( $kit_post->ID, '_activity_download_count', true );
```
obenland, line 242: Downloads are now an **all-time** counter, but the stats UI still range-scopes views and computes `downloads / views` "Download Rate" percentages per row, in the summary, and in the CSV export (`index.js` lines ~97–99, 119–121, 295–297, 472–475), stamping the chart with the range label. With "Last 7 days" selected, a kit with 500 lifetime downloads and 20 views this week renders **2500.0%**. Only the default "all" range is internally consistent.

Either drop the range buttons, or label downloads as all-time and suppress the rate when `range !== 'all'`.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -173,99 +226,28 @@ function get_jetpack_post_views( $range ) {
 		$num    = intval( str_replace( 'd', '', $range ) );
 	}
 
-	$result = $stats->get_top_posts(
+	$result = $stats->get_total_post_views(
```
obenland: `get_total_post_views()` is being passed the old `get_top_posts()` range mapping, but the underlying `/stats/views/posts` endpoint accepts only `post_ids`/`num`/`date`/`offset` (checked against Jetpack 13.3.1 as pinned in composer.lock, and the WPCOM endpoint docs): `period` is ignored and `num` is capped at 30 days. So "90 days" silently returns ~30 days of views, and "All time" (`period=month, num=36`) also returns ~30 days — wrong by orders of magnitude, with no error. `post_ids` is additionally capped at 100 IDs, so once the library exceeds 100 kits the overflow IDs silently report 0 views.

Suggest mapping ranges within the `num <= 30` constraint (or summing multiple windowed calls) and chunking `post_ids`.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -66,6 +83,58 @@ function stats_cache_expiration( $expiration ) {
...
+		return new \WP_Error( 'activity_kit_no_zip', __( 'No downloadable file attached to this activity kit.', 'wporg-learn' ), array( 'status' => 404 ) );
+	}
+
+	$zip_url = wp_get_attachment_url( $zip_id );
+
+	if ( ! $zip_url ) {
+		return new \WP_Error( 'activity_kit_zip_url', __( 'Could not resolve the download URL.', 'wporg-learn' ), array( 'status' => 500 ) );
+	}
+
+	// Increment the download counter stored in post meta.
```
obenland: Style nit: three stacked `//` lines — per the project comment conventions this should be a single `/* */` block with ` * ` continuations. (Its content also needs a rewrite per the CAS comment below, since the claim it documents doesn't hold.)

obenland (commented):
Ran a review on this PR — six inline comments below, plus two findings in files this PR doesn't touch:

**1. Library grid/search cards still bypass the counter entirely.** `wp-content/themes/pub/wporg-learn-2024/inc/block-hooks.php` (~line 274) and `wp-content/plugins/wporg-learn/views/block-activity-kit-card.php` (lines 89–93) still link straight at `wp_get_attachment_url()` with the now-orphaned `data-post-id`/`data-track-download` attributes. This PR deletes the only consumer of those attributes (the `_stq` click script) and removes the Jetpack clicks aggregation, so every download from the Activity Library grid or search results is counted nowhere — the dashboard silently under-reports. Suggest one `get_download_url( $kit_id )` helper in the plugin used by all four link sites, and dropping the dead attributes.

**2. The stats page still hard-gates on Jetpack, making download counts unreachable exactly when they'd matter.** `wp-content/plugins/wporg-learn/js/activity-kit-stats/index.js` (~line 402): `render()` early-returns on the `jetpackAvailable` flag before `fetchStats()` is ever called, with a message saying download counts *require* a Jetpack connection — which this PR makes false. The recorded `_activity_download_count` values are never displayed when Jetpack is disconnected. Suggest always fetching and showing views as "—" when the response carries `jetpack_unavailable` (which is currently dead payload no JS reads).

Also worth noting: historical Jetpack-clicks download data is lost with no backfill — the new counter starts at zero for every kit, so the dashboard will show a cliff.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
Piyopiyo-Kitsune: Fixed — route now uses `(?P<id>\d+)` (post ID) rather than a slug pattern. The template builds the URL with `$kit_id` directly, which is always a plain integer and has no encoding ambiguity.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
Piyopiyo-Kitsune: Added two guards: (1) a User-Agent check that 404s known crawlers, unfurlers (Slack, Discord, WhatsApp, Twitter/LinkedIn bots), and empty UAs — they redirect to the file without touching the counter; (2) explicit `Cache-Control: no-store` + `Pragma: no-cache` headers on the 302 response so no intermediary can treat it as cacheable.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
Piyopiyo-Kitsune: Replaced entirely. The new code issues a single `UPDATE … SET meta_value = meta_value + 1` with `add_post_meta` as a fallback for the first-download case, then calls `wp_cache_delete`. The `UPDATE` is atomic at the DB level with no read; the comment above it now documents both failure modes of the old CAS approach (zero-skips-WHERE and stale-cache-on-retry) so the reasoning is clear.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
Piyopiyo-Kitsune: Fixed in all three spots (summary, table rows, CSV export): the Download Rate is now suppressed to `—` / `N/A` when `activeRange !== 'all'`, and the rate summary box is hidden in those cases. The Downloads column header and summary box label also update to read `Downloads (all time)` when a range-scoped view window is active, so admins can see the mismatch at a glance rather than misread the rate.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
Piyopiyo-Kitsune: Fixed. `period` is no longer passed. `num` is capped per window at 30 (7d → 1 call of 7, 30d → 1 call of 30, 90d/all → 3 calls of 30 offset by 0/30/60 days, results summed). `post_ids` is chunked in groups of 100 and the inner loop issues one call per chunk per window. The `rangeLabel()` in the JS now shows `'All time (max 90 days)'` to be honest about the ceiling.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
Piyopiyo-Kitsune: Rewritten as a single `/* */` block. Also took the opportunity to document the two specific failure modes of the old CAS approach that made the comment's original claim incorrect.

wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
```diff
@@ -66,6 +88,81 @@ function stats_cache_expiration( $expiration ) {
...
+				"UPDATE {$wpdb->postmeta} SET meta_value = meta_value + 1 WHERE post_id = %d AND meta_key = %s",
+				$kit_post->ID,
+				'_activity_download_count'
+			)
+		);
+		if ( 0 === $updated ) {
+			// No row yet — insert with an initial count of 1.
+			// $wpdb->query() returns false on DB error and 0 when no rows matched;
+			// strict comparison avoids falling into this branch on a real error.
+			add_post_meta( $kit_post->ID, '_activity_download_count', 1, true );
```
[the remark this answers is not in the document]
Piyopiyo-Kitsune: Fixed in 53962d1. The INSERT fallback now captures `add_post_meta()`'s return value; when it returns `false` (the concurrent request already inserted the row), the atomic UPDATE is re-run to ensure this download is still counted.

For kits published after the `publish_activity_kit` hook lands, the row is pre-seeded at publish time, so this fallback path is never reached.

> 🤖 Written with [Claude Code](https://claude.com/claude-code)

wp-content/plugins/wporg-learn/inc/post-meta.php
```diff
@@ -1078,4 +1047,36 @@ function register_activity_kit_meta() {
 			'auth_callback'     => $auth_callback,
 		)
 	);
+
+	register_post_meta(
+		'activity_kit',
+		'_activity_download_count',
+		array(
+			'description'       => __( 'Number of times this activity kit ZIP has been downloaded via the tracked download endpoint.', 'wporg-learn' ),
```
obenland:
```suggestion
			'description'       => 'Number of times this activity kit ZIP has been downloaded via the tracked download endpoint.',
```
I don't think this is user-facing

Piyopiyo-Kitsune: Applied in 12284c9 — removed the `__()` wrapper.

> 🤖 Written with [Claude Code](https://claude.com/claude-code)

obenland:
@Piyopiyo-Kitsune Could you give this another look? I made a few changes. Still doing what you want it to do?

Piyopiyo-Kitsune:
> @Piyopiyo-Kitsune Could you give this another look? I made a few changes. Still doing what you want it to do?

Yes @obenland looks like it's still doing what we want it to. Happy to have it merged if it looks good to you. :-)

PR #3658, opened 2026-09-02
Render the Learn patterns only for the post types they are written for

bor0:
## What

`single-tutorial-embed` and `sidebar-lesson-plan` read post meta off the current post and print it. Both assume they are only ever rendered from their own template, but a registered pattern can be rendered outside the template it was written for — `Inserter: no` controls inserter visibility, not registration.

That assumption also reaches the sanitizers. `video_url`, `slides_view_url` and `slides_download_url` are registered with `register_post_meta()`, which scopes a `sanitize_callback` to a single object subtype, so the `esc_url_raw` callbacks do not run for post types the plugin does not register the key against.

## Changes

- Guard both patterns on the post type they are designed around, so the post they read meta from is always one whose meta is registered.
- Treat the tutorial video value as a URL before handing it to `autoembed()`, whose return value is deliberately printed unescaped so provider markup survives.
- Use `esc_url()` rather than `esc_attr()` for the slides `href` values, so the scheme is checked against `wp_allowed_protocols()`.
- Also register the three URL keys without a subtype, giving `sanitize_meta()` a callback to fall back on for other post types.
- Store the admin form's video URL with the same escaper the meta key is registered with.

## How to test

**1. Tutorial embed still works (main regression check).** Open a Tutorial with a WordPress.tv video URL in `video_url`. The video should render exactly as it does on `trunk` — compare the two side by side. Check one tutorial whose URL is a supported provider and one whose URL is not.

**2. Empty value.** On a Tutorial, set `video_url` to an empty string. The page should render with no embed figure at all. On `trunk` it prints an empty `<figure>`; this is an intentional change.

**3. Non-URL value.** Set `video_url` on a Tutorial to a plain string such as `not a url`. The tutorial page should still render, and the value should appear as inert text rather than as markup. View source to confirm nothing was interpreted as HTML.

**4. Lesson plan slides still work (regression).** Open a Lesson Plan with `slides_view_url` and `slides_download_url` set to normal `https://` URLs. Both buttons should render and link correctly, unchanged from `trunk`.

**5. Scheme handling.** Set `slides_view_url` to `javascript:void(0)`. The rendered `href` should come out empty rather than carrying the value through.

**6. Patterns are inert outside their own post type.** Using the post editor's code editor, add these to the content of an ordinary post or page:

```
<!-- wp:pattern {"slug":"wporg-learn-2024/single-tutorial-embed"} /-->
<!-- wp:pattern {"slug":"wporg-learn-2024/sidebar-lesson-plan"} /-->
```

View the post. Neither should produce any output. On `trunk` both render against whatever post they are placed in.

**7. Templates unaffected.** Confirm the single Tutorial and single Lesson Plan templates still render their patterns normally — those are the legitimate call sites and must be unchanged.

**8. Admin form.** Edit a Tutorial through the Workshop Details metabox, save a normal video URL, and confirm it round-trips correctly.

## Notes for reviewers

- The pattern guards are hardcoded to `wporg_workshop` and `lesson-plan`. Each pattern has exactly one call site today (`single-wporg_workshop.html` and `single-lesson-plan.html`), so nothing else should regress — but if either pattern is ever reused from another template, the guard needs updating alongside it.
- `single-activity-kit-content` and `page-course-complete-content` read meta the same way. Their keys are all underscore-prefixed and their output is already escaped, so they are deliberately left alone.

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/post-meta.php
+++ b/wp-content/plugins/wporg-learn/inc/post-meta.php
@@ -291,6 +291,26 @@ function register_common_meta() {
 		);
 	}
 
+	/*
+	 * `register_post_meta()` scopes a `sanitize_callback` to one object subtype, so the callbacks
+	 * below only run for the post types they are registered against. These keys hold URLs that
+	 * the theme prints, so also register them without a subtype: `sanitize_meta()` then has a
+	 * callback to fall back on for post types with no registration of their own.
+	 */
+	$url_meta_keys = array( 'video_url', 'slides_view_url', 'slides_download_url' );
+	foreach ( $url_meta_keys as $url_meta_key ) {
+		register_meta(
+			'post',
+			$url_meta_key,
+			array(
+				'type'              => 'string',
+				'single'            => true,
+				'sanitize_callback' => 'esc_url_raw',
+				'show_in_rest'      => false,
+			)
+		);
+	}
+
 	// Video URL field.
 	$post_types = array( 'wporg_workshop', 'lesson' );
 	foreach ( $post_types as $post_type ) {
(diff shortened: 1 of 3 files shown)


mcliwanow (approved):
Tested on wp-env. Both patterns render nothing on a regular post with their `wp:pattern` blocks, junk in `video_url` comes out as plain text, non-http `slides_view_url` gives an empty href, normal wordpress.tv and https values look the same as on trunk. `update_post_meta()` on a plain post now runs `esc_url_raw` too.

Didn't test the metabox save via the browser, that change is a straight swap. LGTM.

PR #3660, opened 2026-09-03
Revert "Add WPFTP feedback export: Tools admin page + WP-CLI command (#3651)"

maciejpilarski:
Reverts #3651 (merge commit 14c2a19) so the corrected version can land as one clean change. A follow-up PR re-adds the feature with the fixes applied; it will be linked here once open.

## Why revert rather than patch on top

A full review of the merged code (checked against the Jetpack Forms source, the shared `Export_CSV` mu-plugin utility and WordPress core) found problems that change what the export contains and that make the CLI unsafe to run as merged:

- **Date range drops both boundary days.** `date_query` is built from bare `Y-m-d` strings without `inclusive`, so `WP_Date_Query` renders `post_date > '<after> 23:59:59'` and `post_date < '<before> 00:00:00'`. A one-day range exports nothing.
- **Free-text answers are rewritten.** `Export_CSV::add_row()` runs `sanitize_text_field()` on every cell (paragraph breaks collapsed, `<` encoded, `%xx` stripped) and its `esc_csv()` inserts an apostrophe before any `-`, `+`, `@` or `=` that follows a space. Jetpack's own export writes values verbatim with only a leading-character guard.
- **CLI arguments widen the export instead of failing.** An unrecognised `--status` (e.g. `all`) makes `WP_Query` drop the status clause and export spam, trash and temporary partial submissions; an unparseable `--after` becomes `1970-01-01` and exports everything; a non-numeric `--post` casts to `0`, which means all forms.
- **The CLI writes into the current directory by default.** Run from the site root, `wpftp-feedback_<date>.csv`, a guessable name holding usernames, display names, user IDs and IP addresses, lands in the web root.
- Smaller: an all-test-response result becomes a `wp_die()` instead of the "no responses" notice; spam/trash rows are merged in with no status column; same-day exports overwrite each other; the ID column is matched against the English literal `ID` while Jetpack builds the key from `__( 'ID', 'jetpack-forms' )`; the general-purpose `export` capability gates a de-anonymising export.

Since nothing has used the export yet, reverting keeps trunk in a known-good state while the fixed version goes through review.

## Test plan

- [ ] `wp-content/plugins/wporg-learn/inc/` no longer contains `feedback-logged-in-user-export.php` or `class-feedback-export-cli.php`, and `wporg-learn.php` no longer requires them.
- [ ] Tools no longer lists "WPFTP Feedback".


maciejpilarski:
Follow-up that re-lands the feature with the review fixes applied: #3661 (stacked on this PR, so merge this one first).

PR #3662, opened 2026-09-03
Activity Kit Stats: unique downloaders, last downloaded/viewed, rate-limited counting (supersedes #3654)

maciejpilarski:
Supersedes #3654, which it re-lands without the custom table. Original work by @Piyopiyo-Kitsune; the columns, tooltips and download-rate change are hers, the storage and the fixes below are the difference.

## What the dashboard gains (from #3654)

- **Unique Daily Downloaders** per kit, and **Download Rate** redefined as unique downloaders divided by views, so repeat downloads can no longer push the rate past 100%.
- **Last Downloaded** and **Last Viewed** columns, and "Last Updated" renamed to "Kit Last Updated".
- Hover tooltips on every metric header saying exactly what the number is.

## Why no table

#3654 added a `wp_activity_kit_downloads` table beside the daily postmeta buckets that already count downloads, with `dbDelta()` on `init` and a note that WordPress.org may need the table provisioned by hand. Uniques do not need a table. This PR counts them in a second daily postmeta bucket, `_activity_unique_downloads_YYYYMMDD`, read by the same `get_download_counts()` with a key prefix, and dedupes visitors with a transient that expires at the next UTC midnight. Transients live in the persistent object cache on WordPress.org, so there is no schema, no provisioning step and no version option.

## What was wrong in #3654 and how this differs

- **Last Viewed parsed the wrong shape.** WordPress.com's `/stats/post/{id}` returns `weeks` as a list of weeks, each a list of `{ day, count }` objects. The parser iterated it as a date-keyed map, so it returned a list index and every kit would have shown "Invalid Date"; sorting the column would then throw. `extract_last_viewed_date()` now handles the live list shape first, keeps the two map shapes as fallbacks, and only ever returns a real `Y-m-d` or `null`. Checked with a harness against the fixture shape (all eight cases pass).
- **Schema check on every request.** Gone with the table.
- **User-Agent in the visitor hash.** The hash was `hash( salt | ip | user_agent | kit )`, so one client could mint unlimited "unique" downloaders by varying the header. The visitor key is now an HMAC of IP and kit ID with a daily salt derived from `wp_salt()`; nothing about the visitor is stored, and the same person hashes to an unrelated value tomorrow. Known trade-off, stated in the tooltip: a classroom behind one NAT address counts as one downloader per day.
- **The download endpoint had no rate limit at all** (a finding on the merged code that this PR touches). The same transient now carries a per-visitor counter: past five counted downloads per kit per day the file is still served but the counters stop moving, which bounds what a scripted loop can do to the numbers.
- **`get_last_viewed_dates()` used `wp_cache_set()`**, which is request-scoped without a persistent cache, and kept calling Jetpack once per kit after the first failure. It now uses a six-hour transient and stops at the first `WP_Error`, since one failure means the connection is down for every kit and each further call costs a full timeout.
- **"All time" was 180 days.** Views are capped by the WordPress.com API at 30-day windows, and downloads are shown over the same span so the rate compares like with like, so the range is now labelled "Last 6 months" on the button, the chart subtitle and the CSV filename, with a tooltip explaining why.

## Dashboard JS

- A failed fetch used to leave the previous range's summary numbers and chart on screen under the newly highlighted range button, and Export CSV exported that stale data. The summary strip and chart are now cleared on error.
- Range and kit clicks were fire-and-forget; a slow earlier request could overwrite a later one. Each `render()` now takes a ticket and drops responses that are no longer current.
- The CSV export appended nothing to the document and revoked the blob URL synchronously after `click()`, which fails in Firefox and Safari. The anchor is attached first and revoked after a delay; the filename now includes the range.
- The sort comparator, which spelled every column out twice (and had to be edited in both halves for each new column), is one `sortValue()` switch.

## Testing

- `php -l` and `phpcs` against this repo's `phpcs.xml.dist` are clean on both PHP files; `node --check` passes on the JS.
- The `weeks` parser is covered by a standalone harness that loads the real file with WordPress stubbed and asserts the live list shape, both legacy map shapes and four malformed inputs; not committed, since the plugin has no PHPUnit suite.
- I have no Learn wp-env with a Jetpack connection here, so please @Piyopiyo-Kitsune run the dashboard end to end before merging:
  - [ ] Download a kit twice from one browser: Downloads +2, Unique Daily Downloaders +1, Last Downloaded is today.
  - [ ] Download the same kit a sixth time the same day from the same address: neither counter moves, the file still downloads.
  - [ ] Last Viewed shows a real date for a kit with recent views and a dash for a kit with none; sorting that column does not throw.
  - [ ] Deactivate Jetpack, then reactivate: no fatal, and the stats REST call makes at most one Jetpack request when the connection is broken (check the request log).
  - [ ] Switch ranges quickly: the highlighted button always matches the numbers shown.
  - [ ] Export CSV in Firefox and Safari as well as Chrome.
  - [ ] Hover each column header and the "Last 6 months" button for the tooltips.

### DIFF
--- a/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
+++ b/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
@@ -496,14 +496,22 @@ function get_download_counts( $range, array $kit_ids, $prefix = '_activity_downl
 
 	$id_placeholders = implode( ',', array_fill( 0, count( $kit_ids ), '%d' ) );
 
-	// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- The meta API has no ranged multi-key read.
+	/*
+	 * phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery
+	 * phpcs:disable WordPress.DB.DirectDatabaseQuery.NoCaching
+	 * phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+	 * phpcs:disable WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber
+	 *
+	 * The meta API has no ranged multi-key read; $id_placeholders is a list of
+	 * %d tokens built dynamically from count( $kit_ids ) — phpcs cannot count them.
+	 */
 	$rows = $wpdb->get_results(
 		$wpdb->prepare(
-			// phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared -- $id_placeholders is a list of %d placeholders built above from count().
 			"SELECT post_id, meta_key, MAX( CAST( meta_value AS UNSIGNED ) ) AS downloads FROM {$wpdb->postmeta} WHERE post_id IN ( {$id_placeholders} ) AND meta_key BETWEEN %s AND %s GROUP BY post_id, meta_key",
 			array_merge( $kit_ids, array( $first_key, $last_key ) )
 		)
 	);
+	// phpcs:enable WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber
 
 	$map = array();
 	foreach ( $rows as $row ) {
--- a/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
+++ b/wp-content/plugins/wporg-learn/inc/activity-kit-rest.php
@@ -10,6 +10,12 @@ namespace WPOrg_Learn\Activity_Kit_REST;
 defined( 'WPINC' ) || die();
 
 /**
+ * Most downloads counted per visitor per kit per UTC day. Further downloads are still
+ * served, just not counted; see count_download().
+ */
+const DOWNLOAD_COUNT_CAP_PER_VISITOR = 5;
+
+/**
  * Actions and filters.
  */
 add_action( 'rest_api_init', __NAMESPACE__ . '\register_routes' );
@@ -101,18 +107,15 @@ function stats_cache_expiration( $expiration ) {
 /**
  * Handle GET /activity-kits/v1/download/{id}
  *
- * Increments the kit's download counter (stored in one post meta row per UTC
- * day) for requests that look like a person, and redirects the browser to the
- * actual ZIP file URL.
- * Using a server-side redirect lets us track same-domain downloads that
+ * For requests that look like a person, counts the download (see
+ * count_download()) and then redirects the browser to the actual ZIP file
+ * URL. Using a server-side redirect lets us track same-domain downloads that
  * Jetpack's outbound-click tracker misses.
  *
  * @param \WP_REST_Request $request The REST request.
  * @return \WP_REST_Response|\WP_Error 302 redirect on success, WP_Error on failure.
  */
 function handle_download( $request ) {
-	global $wpdb;
-
 	/*
 	 * Skip counting unfurlers, scanners and prefetches. Still redirect: the
 	 * heuristic has false positives, and those should cost an uncounted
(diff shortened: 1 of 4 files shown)


Piyopiyo-Kitsune:
## QA Results

Tested on the `feature/activity-kit-stats-downloader-metrics-v2` branch (commit `048d92b`) using wp-env + Jurassic Tube at `https://activitylibrary.jurassic.tube`.

### Checklist from PR description

- [x] **Download a kit twice from one browser: Downloads +2, Unique Daily Downloaders +1, Last Downloaded = today.**
  Confirmed via the REST API and post meta. Two curl requests from the same IP incremented `_activity_downloads_YYYYMMDD` by 2 and `_activity_unique_downloads_YYYYMMDD` by 1 (stayed at 1 on the second). `_activity_last_downloaded` was set to today's date.

- [x] **Download the same kit a sixth time the same day from the same address: neither counter moves, the file still downloads.**
  After 5 counted downloads, the 6th returned HTTP 302 to the ZIP (file served) but both the daily download bucket and unique bucket stayed unchanged.

- [x] **Last Viewed shows a real date for a kit with recent views and a dash for a kit with none; sorting that column does not throw.**
  All kits show `—` (no Jetpack connection in local env, as expected). Clicking the Last Viewed column header twice produced no JS errors and the table remained stable.

- [x] **Deactivate Jetpack, then reactivate: no fatal, and the stats REST call makes at most one Jetpack request when the connection is broken.**
  Deactivating Jetpack rendered a clean "Jetpack Stats is not connected" message with no fatal. After reactivation, reviewing the code confirms `get_last_viewed_dates()` breaks on the first `WP_Error` — so at most one Jetpack API call fires per stats page load when the connection is down.

- [x] **Switch ranges quickly: the highlighted button always matches the numbers shown.**
  Rapid-clicked all four range buttons in sequence. The final selection (Last 90 days) held `is-active` and the chart subtitle matched. No stale data bleed-through.

- [x] **Export CSV in Firefox and Safari as well as Chrome.**
  CSV blob was generated correctly (text/csv, all new columns present). Code inspection confirms the anchor is appended to `document.body` before `.click()` and the blob URL is revoked after a delay — the Firefox/Safari fix is in place. Filename includes the active range (e.g. `activity-kit-stats-90d.csv`). Could only run the in-app browser locally; recommend a manual click in Firefox/Safari before merge.

- [x] **Hover each column header and the "Last 6 months" button for the tooltips.**
  All column headers (Views, Downloads, Unique Daily Downloaders, Download Rate, Last Downloaded, Last Viewed, Kit Last Updated) carry `title` tooltip text. The "Last 6 months" button tooltip explains the 30-day window cap. All present and accurate.

### Additional fixes pushed

- **`profiles.php` phpcs** (pre-existing, touched by this PR): added missing `@package` tag, strict comparison `!=` → `!==`, and five missing `@param` tags on `add_course_completed_activity()` — committed as `048d92b`.

### What wasn't covered locally

- **Last Viewed with real data** requires a Jetpack-connected site. The parser and fallback logic are unit-tested in the standalone harness described in the PR; the live shape is handled first.
- **Firefox/Safari CSV download** — code path is correct, but a manual click in those browsers before merge is still worth doing.

> 🤖 Written with [Claude Code](https://claude.com/claude-code)

Piyopiyo-Kitsune:
All six Copilot threads addressed in `241b06e`:

**`activity-kit-rest.php`**

- **Line 204 — Rate-limit race** — Replaced the non-atomic `get_transient`/`set_transient` read-modify-write with an atomic `INSERT … ON DUPLICATE KEY UPDATE` on `wp_options`. MySQL's affected-row semantics (1 = new row, 2 = updated, 0 = at cap / no-op) let us atomically detect the first download and prevent concurrent requests from both slipping through the cap or both inflating the unique bucket.

- **Line 287 — Extra Jetpack request on failure** — `get_jetpack_post_views()` now returns `null` on `WP_Error` (distinct from an empty array for "no views"). `handle_stats()` skips `get_last_viewed_dates()` when the result is `null`, keeping the total Jetpack request count at one when the connection is broken.

- **Line 643 — List-of-lists week shape unhandled** — `extract_last_viewed_date()` now normalises `$week` to its `'days'` sub-array when that key exists, otherwise iterates `$week` directly, so both the live list-of-lists shape and the legacy object shape are covered.

- **Line 664 — Invalid calendar dates accepted** — `later_viewed_day()` now adds a `date_create_from_format()` round-trip check after the regex, so values like `2024-02-31` are rejected before they can be returned or displayed.

**`index.js`**

- **Line 432 — `totalKits` not cleared on error** — Added `totalKits` to the elements reset by `clearDisplayedData()`.

- **Line 437 — Pagination controls left stale on error** — `clearDisplayedData()` now also hides `chartSliderWrap` (removes `is-visible`) and resets `chartSliderLabel` and `chartSlider` so the old `1–N of M` label and active slider don't persist after a failed fetch.

> 🤖 Written with [Claude Code](https://claude.com/claude-code)

PR #3671, opened 2026-09-15
Plugin: Use wp_safe_redirect() for the My Courses login redirect

mcliwanow:
Same two changes as WordPress/wordpress.org#903, applied to the Learn copies of these files.

## Why

`restrict_my_courses_page_access()` in `inc/sensei.php` takes `$_GET['redirect_to']`, wraps it in `wp_login_url()` and passes the result to `wp_redirect()`. The wrap means the target always resolves to this host, and `wp-login.php` puts the `redirect_to` query arg through its own `wp_safe_redirect()` before using it after login (the one `wp_redirect()` left in core's login flow only fires for values core computed itself). So this is hardening rather than a live open redirect: it holds the redirect on-host if a filter on `login_url` ever points somewhere else, and it stops the plugin from having a lone call that hands request data to the unrestricted function.

`wporg_learn_redirect_meetings()` in `inc/redirects.php` is the case that has to keep `wp_redirect()`. Meeting posts exist to link out to Meetup and similar, and the target is already checked with `wp_http_validate_url()` on the line above. Nothing said that was deliberate, so it read as an oversight to anyone scanning for the same pattern.

## What changed

- `inc/sensei.php`: `wp_redirect()` to `wp_safe_redirect()` for the My Courses login redirect.
- `inc/redirects.php`: `phpcs:ignore WordPress.Security.SafeRedirect.wp_redirect_wp_redirect` on the meeting redirect, with the reason inline. No behaviour change.

## Testing

Local `wp-env`, requesting `/my-courses/` logged out on trunk and on this branch in the same session, reading the `Location` header. (`WP_DEBUG_DISPLAY` had to be turned off first: the textdomain notices print before headers in this env, so the redirect never fires and every request comes back 200.)

| `redirect_to` | trunk | branch |
|---|---|---|
| absent | `wp-login.php?redirect_to=http%3A%2F%2Flocalhost%3A8910%2Fmy-courses` | same |
| `https://evil.example.com/x` | `wp-login.php?redirect_to=https%3A%2F%2Fevil.example.com%2Fx` | same |
| `//evil.example.com/x` | `wp-login.php?redirect_to=%2F%2Fevil.example.com%2Fx` | same |
| `/courses/` | `wp-login.php?redirect_to=%2Fcourses%2F` | same |

Identical in all four, which is the expected result given `wp_login_url()` is doing the wrapping. The change is there for the filtered case, and the off-host value is still rejected downstream by `wp-login.php`.

`phpcs` clean on both changed files.


mcliwanow:
fyi @obenland

